PR #25, opened 2016-02-29
Remove useless/unavailable info from AppsIndex

drew-gross:
Instead, display the server version. We can think of other stuff to add later, hopefully.


parseplatformorg:
🎉 This change has been released in version [4.1.1-alpha.4](https://github.com/parse-community/parse-dashboard/releases/tag/4.1.1-alpha.4)

PR #58, opened 2016-03-05
Adds commander CLI, improve remote deployments

### LINKED ISSUE #54: Improve the dashboard experience when deploying to Heroku or other remote machines
Currently, deploying the dashboard to Heroku or similar services is awkward because we go to great lengths to ensure people don't accidentally leak their master key. This is useful, but can also make deploying difficult due to early SSL termination, untrusted proxies, and so on.

This would be a good first PR for someone new to the dashboard, but probably not so much to someone who is new to web security.

flovilmart:
- Adds support for --config, --port --allowInsecureHTTP
- Adds support for PARSE_DASHBOARD_ALLOW_INSECURE_HTTP
- Adds npm start script for easy deployments
- Adds documentation for running behind HTTPS load balancers

fixes #54

### DIFF
--- a/Parse-Dashboard/index.js
+++ b/Parse-Dashboard/index.js
@@ -5,6 +5,22 @@
  * This source code is licensed under the license found in the LICENSE file in
  * the root directory of this source tree.
  */
+// Command line tool for npm start
+
+var DEFAULT_DASHBOARD_CONFIG = __dirname + '/parse-dashboard-config.json';
+
+var program = require("commander");
+program.option('--port [port]', "the port to run parse-dashboard");
+program.option('--config [config]', "the path to the configuration file");
+program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set that flag when parse server is behind an HTTPS load balancer/proxy');
+
+program.parse(process.argv);
+
+// collect the variables
+var configFile = program.config || DEFAULT_DASHBOARD_CONFIG;
+var port = program.port || process.env.PORT;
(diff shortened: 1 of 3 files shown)


peterdotjs:
nice! could we not reuse the `npm run dashboard` command rather than use `npm start`?

flovilmart:
not really as we 'demonize' with & and watch for changes. that breaks the passed command line arguments :(

peterdotjs:
:( lgtm. We're still referencing `npm run dashboard` elsewhere like in the blog post but we should consolidate the scripts in future. I'm not sure how useful having watch for config changes will be. And it'll be one less script for devs to think about.

flovilmart:
I've updated the README to reference it for 'deployments'.
IMHO, webpack is nice, but is a pain when it comes to copy files.
Not sure what the production.config.js is doing now or if it's supposedly functional

drew-gross:
`npm run dashboard` still works. Even if we recommend other ways of running, we should keep that one around since it's handy for hacking on the dashboard.

@flovilmart `production.config.js` builds minified bundles and enables a custom Babel plugin that removes proptypes. It cuts the size of the bundle from 3.2mb to 1.1mb, which is nice.

flovilmart:
yeah but it don't build a production version in dist/ with the minified that I would use for `npm start` for example :)

I could add this to web pack so when we run `npm run build` it creates a dist folder ?

drew-gross:
Yeah feel free to modify it. I think `bundles` is the name we should stick with though, no real reason to change it.

flovilmart:
Good! so `bundles/parse-dashboard/...`

PR #63, opened 2016-03-05
Docker

timchunght:
Added runnable Dockerfile and updated README with instructions on how to use parse-dashboard with Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,4 +3,4 @@ ADD package.json /src/package.json
 RUN cd /src && npm install
 ADD . /src
 WORKDIR /src
-CMD ["npm", "run", "dashboard"]
+ENTRYPOINT ["npm", "start", "--"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM node:4.3.2
+ADD package.json /src/package.json
+RUN cd /src && npm install
+ADD . /src
+WORKDIR /src
+CMD ["npm", "run", "dashboard"]
--- a/README.md
+++ b/README.md
@@ -92,16 +92,15 @@ Run the image with your ``config.json`` mounted as a volume
 docker run -d -p 8080:4040 -v host/path/to/config.json:/src/Parse-Dashboard/parse-dashboard-config.json parse-dashboard 
 ```
 
-The container will boot up using the default command ``npm run dashboard``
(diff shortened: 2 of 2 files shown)


drew-gross:
Is it possible to remove the `--allowInsecureHTTP=1` by default? It's incredibly important to disallow insecure connections by default because accessing the dashboard on the public internet via HTTP will leak all your master keys.

timchunght:
@drew-gross The --allowInsecure is not enabled by default. "npm run dashboard" is the default command for the image.

timchunght:
@drew-gross I just updated the README with explanation that the custom command without HTTPS is only used for test/development purposes using Docker.

flovilmart:
@timchunght the npm run dashboard is also watching the files locally which is fine for development, but not recommended for production.

timchunght:
@flovilmart I just updated the Dockerfile to use an ENTRYPOINT of the production command and it will allow custom configurations to be passed in.

timchunght:
@drew-gross @flovilmart I also think it could be better if we have the option to build the webpack in a separate command from starting the server. This way, we can have faster start time. Please advise on what you think is the best way to achieve that. From reading the *.build.js files, it appears that the webpack build script doesn't have explicit dependency on the ``parse-dashboard-config``, which is the only external variable. Thanks.

flovilmart:
@timchunght I'm actually working onto just that :)

timchunght:
@flovilmart That's brilliant! I am happy that we are on the same page. Haha. I really like the Parse project. I personally really like Go and built a compatible Go version [This](https://github.com/timchunght/goparse). It is just the binary file to let some of friends test it out without setting up Go environment. It is still very early and I will open source it when it is more ready. 

I am wondering why didn't you guys open source the Go version of the API.

flovilmart:
Easier to deploy with node.js, more integrated with cloud code, no subprocesses to manage I believe :)

timchunght:
@flovilmart I see. That's a shame that Go version is open source. Haha. I think Go is easier to deploy (could be biased). Regardless, these (server and dashboard) are great projects and it has been fun to work on them. @drew-gross I hope my docker contribution is useful. Thanks.

drew-gross:
Yeah, the go version is massively multi-tenant and includes tons of multi-tenant specific code. It also relies on like 6 different databases, takes 2-3 days to get it running, and would be unlikely to fit in the smallest tier on any IaaS provider. I wouldn't have been a good experience. Thanks for the Docker contribution!

PR #76, opened 2016-03-06
Adds distribution with built package

flovilmart:
- Adds gulp build system to build the parse-dashboard dist
- Adds instruction for safe production deployments
- Adds ability to build npm package
- npm package only contains dist/ and optimized bundles
- 5 steps configuration for production deployments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,16 @@ If you are not familiar with Docker, ``--port 8080`` with be passed in as argume
 
 ## Deploying in production
 
+For production deployments, it's recommended to use the npm package
+
+1. Create a folder for your project
+2. run `$ npm init`
+3. Create your dashboard.json in the root ot your project
+4. run `$ npm install --save parse-dashboard`
+5. add a start script in your package.json  `"start": "parse-dashboard --config ./dahsboard.json"` 
+6. run `$ npm start`
+
+
 If you're deploying to a provider like Heroku, or Google App Engine, the SSL endpoint is terminated early and handled by the provider and you may encounter this error `Parse Dashboard can only be remotely accessed via HTTPS`. 
 
 :warning: :warning: Before going further, make sure your server **cannot** be reachable via **HTTP**. See the provider documentation for force HTTPS connections to your deployment.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -104,6 +104,16 @@ If you are not familiar with Docker, ``--port 8080`` with be passed in as argume
 
 ## Deploying in production
 
+For production deployments, it's recommended to use the npm pacakge
```
peterdotjs: package (sp)

package.json
```diff
@@ -1,5 +1,15 @@
 {
-  "private": true,
+  "name": "parse-dashboard",
+  "description": "The parse dashboard",
+  "version": "1.0.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ParsePlatform/parse-dashboard"
+  },
+  "license": "BSD-3-Clause",
```
peterdotjs: We'll need to reference the `license` file. 

`"license: "LicenseRef-LICENSE"` 

Ref: https://github.com/npm/npm/issues/8795

flovilmart: ok!

flovilmart: Put the same license as is parse-server

peterdotjs:
other than the two comments, lgtm!

flovilmart:
@peterdotjs version 1.0.0 is right? @drew-gross has the access to npmjs to publish the package, I can't publish it :(

peterdotjs:
Only @drew-gross I think has publish perms. Drew could you give us permissions?

Hmm regarding version # we could go the same route as `parse-server` and start with 2.0.0? Thoughts @gfosco ?

flovilmart:
bumped to 2.0.0, at publish time, @drew-gross could lower leave it

drew-gross:
@flovilmart what benefit do we get from gulp in this case? It doesn't seem like it's doing much right now, and I don't really want to add an entire other build step unless it solves a really big problem.

flovilmart:
@drew-gross webpack is really bad at moving files around, and it would likely require another package, like webpack-copy-files, a new webpack config, based on production. I don't mind removing gulp, but I tried with webpack and never got a good reliable solution. I'm open to remove gulp.

drew-gross:
@flovilmart why would webpack need to move files around? Seems like you should be able to just change the output directory.

flovilmart:
We need to copy the dashboard files into dist, that is not committed, and only created upon publish.

PR #110, opened 2016-03-09
Use browserHistory from React Router 2 instead of createHashHistory

kevinold:
Patch for PIG, per console warning

screen shot 2016-03-09 at 10 38 24 am


peterdotjs:
awesome! lgtm.

PR #111, opened 2016-03-09
Add an 'Add a Class' entry to the edit menu

bgw:
The data browser 'Edit' menu has related options like 'Add a Row' or 'Delete this class', so 'Add a class' is a natural addition.

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -561,7 +561,8 @@ export default class Browser extends DashboardView {
             onPointerClick={this.handlePointerClick.bind(this)}
             setRelation={this.setRelation.bind(this)}
             onAddColumn={this.showAddColumn.bind(this)}
-            onAddRow={this.addRow.bind(this)} />
+            onAddRow={this.addRow.bind(this)}
+            onAddClass={this.showCreateClass.bind(this)} />
         );
       }
     }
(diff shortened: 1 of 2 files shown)


peterdotjs:
nice! one nit - could we add it at the bottom of the menu with a separator?
        
```
<Separator />
<MenuItem text='Add a class' onClick={onAddClass} />
```

My reasoning is that other actions are specific to the class itself. And this is for creating a new class.

drew-gross:
I actually like it where it is - all the "add" operations are grouped and all the "delete" operations are grouped.

adampietrasiak:
There is a lot of space there, My skills don't allow me to make PR, but from my UX experience I see it like it's begging to make some icons there instead of dropdown to reduce task from 2 clicks to 1 click and to make you instantly see what you can do.

drew-gross:
@adampietrasiak agreed - I would love to see it responsively move items from the toolbar to the dropdown as the browser window shrinks. That is a bigger effort though :) A medium amount of effort would be to have all the items always in the dropdown, and have only the toolbar be responsive.

peterdotjs:
@drew-gross it is nice that they are grouped together but odd that it is isn't specific to this class like the others. I don't think it's really worth arguing about. I'll go ahead and merge.

@adampietrasiak we had some plans to move the actions out of the dropdown towards the left side of the header which would make for better accessibility but we never got to it.

bgw:
I'd be interested in taking a stab at implementing some sort of responsive toolbar, if people think it's worth the engineering effort.

PR #126, opened 2016-03-10
App name ellipsis

bgw:
I know the only thing discussed before was ellipsis in the sidebar, but ellipsis on the app index page were broken too. So I fixed both.

The commit fixing the app index page depends on flexbox. Hopefully this isn't a problem, since I expect that we only care about targeting modern browsers, but I'm open to discussing alternate solutions if we can't use flexbox.

# Before

selection_023
selection_026

# After

selection_024
selection_025
selection_027

### DIFF
--- a/src/components/Sidebar/Sidebar.scss
+++ b/src/components/Sidebar/Sidebar.scss
@@ -112,6 +112,7 @@
 }
 
 .currentApp, .menuRow {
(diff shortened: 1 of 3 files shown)


peterdotjs:
@bgw nice! We're only supporting the latest and greatest so flex box totally works.

PR #152, opened 2016-03-16
Save ordering between different use

darkkiss:
Save ordering between different use

### DIFF
--- a/src/lib/ColumnPreferences.js
+++ b/src/lib/ColumnPreferences.js
@@ -51,6 +51,7 @@ export function getColumnSort(sortBy, appId, className) {
     updated = true;
   }
   if (updated && sortBy) {
+    cachedSort = sortBy;  
     updatePreferences(sortBy, appId, objectName);
   }
   return cachedSort;
--- a/src/lib/ColumnPreferences.js
+++ b/src/lib/ColumnPreferences.js
@@ -7,6 +7,8 @@
  */
 const VERSION = 'v1'; // In case we ever need to invalidate these
 const DEFAULT_WIDTH = 150;
+const COLUMN_SORT = '_columnSort'; // Used for storing classes sort field
+const COLUMN_DEFAULT_SORT = '-createdAt'; // Default column sorting
 let cache = {};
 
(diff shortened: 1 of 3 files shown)


durunvo:
Is this in progress or done?

src/lib/ColumnPreferences.js
```diff
@@ -43,6 +43,20 @@ export function getPreferences(appId, className) {
   }
 }
 
+export function getColumnSort(sortBy, appId, className) {
+  let objectName = className + ':' + 'columnSort';
```
durunvo: const objectName = `${className}:columnSort`
would be nice

darkkiss: Hello, I don't understand what do you mean by that ? Constant as a variable?

peterdotjs:
@darkkiss awesome job for contributing! Have you tested this out? Currently for preferences we are storing the column width and order in an array. You'll need to add an additional `sort` field to the localStorage preference and ensure we are reading it.

If you take a quick look at the local storage you'll understand what I mean. Let me know if I can make anything more clear.

src/lib/ColumnPreferences.js
```diff
@@ -43,6 +43,20 @@ export function getPreferences(appId, className) {
...
+  let cachedSort = getPreferences(appId, objectName) || [ { name: '-createdAt' } ];
+  let updated = false;
+  if(cachedSort != sortBy) {
```
peterdotjs: nit: !==

src/lib/ColumnPreferences.js
peterdotjs: I think `columnSort` as a `const` placed at the top would be nice. Consistent with our existing style: https://github.com/ParsePlatform/parse-dashboard/blob/master/src/lib/ColumnPreferences.js#L79

peterdotjs:
Just a couple nits but otherwise looks and works great.

mmarshak:
I am using Chrome and I have dashboard version 1.0.22, and the order of the columns is not being saved between different launches of the browser.
Do I need to do something to ask dashboard to save the order?

PR #154, opened 2016-03-17
Improve getting started experience for Parse Dashboard

drew-gross:


Parse-Dashboard/index.js
```diff
@@ -6,31 +6,25 @@
...
-
-var program = require("commander");
-program.option('--port [port]', "the port to run parse-dashboard");
-program.option('--config [config]', "the path to the configuration file");
-program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set that flag when parse server is behind an HTTPS load balancer/proxy');
+const program = require('commander');
+program.option('--appId [appId]', 'the app Id of the app you would like to manage');
+program.option('--masterKey [masterKey]', 'the master key of the app you would like to manage');
+program.option('--serverURL [serverURL]', 'the server url of the app you would like to manage');
+program.option('--appName [appName]', 'the master key of the app you would like to manage');
```
peterdotjs: update description

peterdotjs:
nice! lgtm. Only feedback is in the readme we make the cli information standout more. I feel like someone new will get lost in with the rest of the info

Parse-Dashboard/index.js
drew-gross: Good catch! Thanks.

Parse-Dashboard/index.js
```diff
@@ -41,26 +35,66 @@ packageJson('parse-dashboard').then(latestPackage => {
...
-app.get('/parse-dashboard-config.json', function(req, res) {
-  jsonFile(configFile)
-  .then(config => {
-    config.data.apps.forEach((app) => {
-      if (!app.appName) {
-        return res.send({ success: false, error: 'An application is misconfigured, appName is required' });
-      }
-    });
-    var response = {
+const port = program.port || process.env.PORT || 4040;
```
flovilmart: this is risky

Parse-Dashboard/index.js
```diff
@@ -41,26 +35,66 @@ packageJson('parse-dashboard').then(latestPackage => {
...
+          appId: program.appId,
+          serverURL: program.serverURL,
+          masterKey: program.masterKey,
+          appName: program.appName,
+        }
+      ]
+    }
+  }
+} else {
+  configFile = program.config || path.join(__dirname, 'parse-dashboard-config.json');
```
flovilmart: maybe add a warning that we'll deprecate the file at one point in the future?

drew-gross: Are we going to deprecate that file? That doesn't really seem necessary. At any rate, we can do that in a future version.

flovilmart: :+1:

Parse-Dashboard/index.js
```diff
@@ -6,31 +6,25 @@
...
-var program = require("commander");
-program.option('--port [port]', "the port to run parse-dashboard");
-program.option('--config [config]', "the path to the configuration file");
-program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set that flag when parse server is behind an HTTPS load balancer/proxy');
+const program = require('commander');
+program.option('--appId [appId]', 'the app Id of the app you would like to manage.');
+program.option('--masterKey [masterKey]', 'the master key of the app you would like to manage.');
+program.option('--serverURL [serverURL]', 'the server url of the app you would like to manage.');
+program.option('--appName [appName]', 'the name of the app you would like to manage. Optional.');
+program.option('--config [config]', 'the path to the configuration file');
```
flovilmart: (as per slack) maybe allow the config as an argument instead of --config to match parse-server implementation

Parse-Dashboard/index.js
peterdotjs: Users with multiple apps can still used it. I also don't see a reason to deprecate.

flovilmart: __dirname with npm would mean `node_modules/parse-dashboard/Parse Dashboard`

Parse-Dashboard/index.js
drew-gross: The danger with non-named arguments is that it's easy to forget which ones exist and in what order, especially as the arguments list grows. I prefer named arguments for everything. This also would be easy to add later, but difficult to to remove later.

flovilmart: Ok, then I'll add it also for parse-server a --config so we have something consistent

PR #158, opened 2016-03-17
Fix loading of default config

drew-gross:


peterdotjs:
trying out the patch now

Parse-Dashboard/index.js
```diff
@@ -38,34 +38,44 @@ packageJson('parse-dashboard').then(latestPackage => {
...
+      }
+    };
+  } else if (!program.serverURL && !program.masterKey && !program.appName) {
+    configFile = path.join(__dirname, 'parse-dashboard-config.json');
   }
 } else {
-  configFile = program.config || path.join(__dirname, 'parse-dashboard-config.json');
+  configFile = program.config;
   if (program.appId || program.serverURL || program.masterKey || program.appName) {
     console.log('You must provide either a config file or app ID, Master Key, and server URL, not both.');
```
peterdotjs: You must provide either a config file or required CLI options (app ID, Master Key, server URL); not both.

Think this would be more clear.

peterdotjs:
patch looks good for legacy scripts, nit on error content.

peterdotjs:
CLI looks good.

PR #181, opened 2016-03-23
Fixed bug with determining new features in latest version.

andresgarza:
New features from latest package weren't being properly parsed. This caused "Upgrade" banner never to show.


drew-gross:
Thanks!

PR #195, opened 2016-03-29
Add support for relation's count in toolbar details

gavrix:
Before objects count is cached with app schema and only shows for 'raw' classes presentations. This change adds support for objects count for relation presentations. The count itself is fetched every time Browser component gets new relation.


drew-gross:
Nice! Thanks for the contribution.

PR #179, opened 2016-03-22
Change secure check in production environment & Run with Heroku 

gimdongwoo:
req.secure is undefined in production evironment.

It works on heroku with https.

Thank you.

PR #196, opened 2016-03-29
Heroku-friendly, Production deployable, Dashboard.

mnoble:
This PR is two-fold.

First, this makes the project buildable via a production deploy. For that, build dependencies need to be moved to the `dependencies` instead of `devDependencies` so that they're available on the server, post-deploy. I also added a couple scripts to build assets and put them into the correct place. This happens `post-install`, but only when you explicitly set the `POST_BUILD_INSTALL` env var. This should prevent it from running when installing deps locally on your machine (or when running locally).

Second, this makes Dashboard Heroku-friendly. Because of the way routing works on Heroku, all requests appear to be from localhost and non-HTTPS. This is just because TLS termination happens way before the app. We set the usual `X-FORWARDED-FOR` and `X-FORWARDED-PROTO` headers, so this uses those when available.


mnoble:
Actually, hold off on this for a moment. My bash-fu is not strong and this is compiling the app locally without the env var :stuck_out_tongue:

drew-gross:
Awesome! Looks like this overlaps with #179 a lot, and has one of the same issue, which is that this would make the dashboard insecure when you aren't deploying behind any load balancers or anything like that. The `--allowInsecureHTTP` flag was intended to be how you tell the dashboard that it's running behind a trusted proxy, could check that flag before trusting the forwarded-for and forwarded-proto headers? Also, did you test this on windows? Obviously deploying to Heroku doesn't need to work on Windows, but we do need to make sure this doesn't break the `localhost` deployment on Windows. If you haven't, I can test that later tonight.

mnoble:
@drew-gross Heh, yea, I saw that PR last week right before I was going to push this one and held off. I figured there's more to this one that fixes some of the build issues of #179, so why not push it up.

re: `--allowInsecureHTTP`, yea, makes sense. I'll add a guard around the headers checking that flag.

re: Windows: No, I don't even have a Windows VM setup, tbh. I'll try to get around to that later today.

gimdongwoo:
Good! Love It! :+1:

gimdongwoo:
@mnoble But, It didn't works on Heroku.
If config vars [NPM_CONFIG_PRODUCTION=true] then, having browser error.
If config vars [NPM_CONFIG_PRODUCTION=false] then, having build error related to babel-polyfill.
Please, help me. I love Heroku.

mnoble:
@gimdongwoo You'd need to `heroku config:set POST_INSTALL_BUILD=1` for this to work. That's the env var the `postinstall` script looks for when determining to build assets.

gimdongwoo:
@mnoble It works perfectly! :+1:  Thank you!!

jgaull:
I'm also using it. Thanks!

mnoble:
I'm closing this in favor of splitting this work up into a few PRs. Mainly #216 which sets up the dependencies and scripts to build on deploy.

The `allowInsecureHTTP` necessity on Heroku is going to need to be handled through separate means (which I have in the works), since that's a little platform-specific for a general change, imo.

PR #202, opened 2016-03-29
Dropdown height

drew-gross:
This will hopefully prevent people from being confused about there being more items available in the dropdown if you scroll


rogerhu:
accepted

drew-gross:
Thanks Roger!

PR #209, opened 2016-03-31
Adds support for using Parse Dashboard as Express middleware

Gameleon12:
Added support for using Parse Dashboard as a Express middleware.
Based on Parse Server middleware.

Project structure changes:

-Wrapped Parse Dashboard code in a ParseDashboard class in
Parse-Dashboard\index.js
-Moved CLI/NPM invoked code to Parse-Dashboard\cli\index.js

### DIFF
--- a/Parse-Dashboard/index.js
+++ b/Parse-Dashboard/index.js
@@ -5,114 +5,47 @@
  * This source code is licensed under the license found in the LICENSE file in
  * the root directory of this source tree.
  */
-// Command line tool for npm start
-"use strict"
+"use strict";
 const packageJson = require('package-json');
 const basicAuth = require('basic-auth');
 const path = require('path');
 const jsonFile = require('json-file-plus');
 const express = require('express');
 
-const program = require('commander');
-program.option('--appId [appId]', 'the app Id of the app you would like to manage.');
-program.option('--masterKey [masterKey]', 'the master key of the app you would like to manage.');
-program.option('--serverURL [serverURL]', 'the server url of the app you would like to manage.');
-program.option('--appName [appName]', 'the name of the app you would like to manage. Optional.');
-program.option('--config [config]', 'the path to the configuration file');
-program.option('--port [port]', 'the port to run parse-dashboard');
-program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set this flag when you are running the dashboard behind an HTTPS load balancer or proxy with early SSL termination.');
+class ParseDashboard {
 
-program.parse(process.argv);
-
-const currentVersionFeatures = require('../package.json').parseDashboardFeatures;
-
-var newFeaturesInLatestVersion = [];
-packageJson('parse-dashboard', 'latest').then(latestPackage => {
-  if (latestPackage.parseDashboardFeatures instanceof Array) {
-    newFeaturesInLatestVersion = latestPackage.parseDashboardFeatures.filter(feature => {
-      return currentVersionFeatures.indexOf(feature) === -1;
-    });
-  }
-});
-
-const port = program.port || process.env.PORT || 4040;
-const allowInsecureHTTP = program.allowInsecureHTTP || process.env.PARSE_DASHBOARD_ALLOW_INSECURE_HTTP;
-
-let explicitConfigFileProvided = !!program.config;
-let configFile = null;
-let configFromCLI = null;
-let configServerURL = program.serverURL || process.env.PARSE_DASHBOARD_SERVER_URL;
-let configMasterKey = program.masterKey || process.env.PARSE_DASHBOARD_MASTER_KEY;
-let configAppId = program.appId || process.env.PARSE_DASHBOARD_APP_ID;
-let configAppName = program.appName || process.env.PARSE_DASHBOARD_APP_NAME;
-let configUserId = program.userId || process.env.PARSE_DASHBOARD_USER_ID;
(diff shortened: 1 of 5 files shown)


Gameleon12:
Just now discovered a bug left in the code. Since all lookup urls for assets in the html and react.js scripts are defined as relative to `/`, the express app will mount but will not load if it is mounted on any other endpoint than `/` (since lookup for asset and script files will fail).

mamaso:
Extremely interested in this enhancement as well.  

If it helps, I mounted the parse dashboard off the /parse-dashboard/ route in this commit: https://github.com/mamaso/ParseDashboardExtension/commit/8821cbf869bc723c2ec28f3a269d9fe3f999cfb0 

Relevant changes were setting <base href="/parse-dashboard/" /> in index.html and changing most /apps/ and /bundles/ root-relative urls in src and webpack to fully relative urls

Gameleon12:
@mamaso Thanks for referencing that commit. It is indeed a very helpful pointer in the right direction. 

The problem with it's current form is that the relative uri's stack up on top of each other while navigating through Parse Dashboard (leading to urls such as `/apps/MyApps/apps/MyApps/browser/apps/MyApps/browser/_Installation/apps/MyApps/browser/apps/MyApps/browser/MyCustomClass`). 

It seems the way to go is to prefix all the uri's in the react.js files with where the Express application has mounted parse-dashboard relative to the root. I am not very proficient with React so any tips, tricks, commits etc. to help with this would be very welcome.

flovilmart:
I'm not sure you need all those changes.

You just need to implement in your middleware a way to:

1. serve parse-dashboard-config.json
2. redirect all requests to index.html

So if there is 1 thing to do, it's maybe just extract the handler for `GET /parse-dashboard-config.json`

mamaso:
@flovilmart Redirecting all requests to index.html doesn't quite solve the issue.  Because all the react urls are root relative and will strip the mount path, how can you detect whether they are actually intended for the root or for whatever path the parse dashboard is mounted on?

@Gameleon12 Interesting, not the biggest expert on react either.  Did you set `<base href="/parse-dashboard/" />`?  See this for more info, useBasename in react seems like the solution they've created for this: https://github.com/reactjs/react-router/issues/2014

flovilmart:
@mamaso  So that's even better, that allows you to do 
```
app.use('/parse-dashboard/*, (req, res) => {
   res.sendFile(__dirname + '/index.html');
});
```

Parse-Dashboard/index.js
```diff
@@ -5,114 +5,48 @@
...
 
-const program = require('commander');
-program.option('--appId [appId]', 'the app Id of the app you would like to manage.');
-program.option('--masterKey [masterKey]', 'the master key of the app you would like to manage.');
-program.option('--serverURL [serverURL]', 'the server url of the app you would like to manage.');
-program.option('--appName [appName]', 'the name of the app you would like to manage. Optional.');
-program.option('--config [config]', 'the path to the configuration file');
-program.option('--port [port]', 'the port to run parse-dashboard');
-program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set this flag when you are running the dashboard behind an HTTPS load balancer or proxy with early SSL termination.');
+class ParseDashboard {
```
flovilmart: this won't work as we don't pass it through babel AFAIK

mamaso:
@flovilmart Just played around a bit and I think you're right, looks like when I did this I went a wee bit overboard on the relative urls.  Here's what I think are necessary changes:

- `<base href="my-express-mount-path"/>`, or modify history.js to specify the basename for react
- serve parse-dashboard-config.json on the mount path
- serve bundle files correctly (this is where the relative urls are needed because actual server requests are made and react basename history doesn't apply)
  - https://github.com/mamaso/ParseDashboardExtension/commit/8821cbf869bc723c2ec28f3a269d9fe3f999cfb0#diff-1d8c20cda1305902f57d145e65b946f5L6
  - https://github.com/mamaso/ParseDashboardExtension/commit/8821cbf869bc723c2ec28f3a269d9fe3f999cfb0#diff-d307b40242a72a84b7296cc30ae87dd6L21
  - https://github.com/mamaso/ParseDashboardExtension/commit/8821cbf869bc723c2ec28f3a269d9fe3f999cfb0#diff-82ac41430b83ec334292b6a1f90128ecL17

Parse-Dashboard/index.js
Gameleon12: The code doesn't pass through babel, but Node.js supports some features of ECMAScript 6 without the need for babel compilation. (See: https://nodejs.org/en/docs/es6/)

flovilmart:
I'm working on something that just does that

Gameleon12:
@flovilmart

>  I'm not sure you need all those changes.
> 
> You just need to implement in your middleware a way to:
> 
>     serve parse-dashboard-config.json
>     redirect all requests to index.html
> 
> So if there is 1 thing to do, it's maybe just extract the handler for GET /parse-dashboard-config.json
> 

While splitting off the general dashboard and cli related code isn't a necessity to add middleware support, it does make the code a bit cleaner to maintain and use. In addition, initialization of Parse Dashboard as middleware in this way is consistent with how Parse Server is initialized as middleware.

flovilmart:
see #214.

Now I need to find a way to properly set <base href="/mount-path"/>, either by temptlates or just a pure regex

drew-gross:
Superseded by #214.

PR #214, opened 2016-04-01
:construction: Splits express app from express server

flovilmart:


flovilmart:
Ahah :) if you mount the app somewhere else but / it won't work...

drew-gross:
Yeah I figured the `construction` thing meant that this was some supporting work that will help implement other stuff later. As long as it's not breaking, it's fine to merge.

flovilmart:
Supporting a relative mount path will be problematic as the bundle expects the config on /, as well as other resources

mnoble:
fwiw, this broke Dashboard for me. `allowInsecureHTTP` isn't defined (or accessible) from `app.js` since it's a flag parsed via commander in `index.js`, but used at https://github.com/ParsePlatform/parse-dashboard/blob/master/Parse-Dashboard/app.js#L43.

flovilmart:
oops! let's fix that

PR #215, opened 2016-04-01
Added 'route' config option for mounting dashboard

mamaso:
Adding a 'route' property to the parse-dashboard-config will now run the dashboard on that route.

@flovilmart If you have a chance to play around with this I'd appreciate it.  You'll see most functionality works, but in developer tools there are requests which seem to be failing due to some root relative '/apps/' urls.  I'd love to find a better solution than changing all the root relative '/apps/' to 'apps/'.


flovilmart:
I have a less intrusive solution in the making

flovilmart:
see #217, I just came back home and pushed it

PR #216, opened 2016-04-01
Build on Deploy

mnoble:
Moves all build deps to `dependencies` (opposed to `devDependencies`) so that the project can be built when deploying to production (like on Heroku, in this case). You enable this by setting the `POST_INSTALL_BUILD`.

Also fixes an issue introduced after splitting up the backend/static express apps where `allowInsecureHTTP` was being referenced while never being defined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,28 +29,25 @@
     "LICENSE"
   ],
   "dependencies": {
-    "basic-auth": "^1.0.3",
-    "commander": "^2.9.0",
-    "express": "^4.13.4",
-    "json-file-plus": "^3.2.0",
-    "package-json": "^2.3.1"
-  },
-  "devDependencies": {
     "babel-core": "~5.8.12",
     "babel-loader": "~5.3.0",
     "babel-plugin-remove-proptypes": "~1.0.0",
     "babel-polyfill": "^6.7.2",
     "babel-runtime": "~5.8.25",
     "css-loader": "~0.18.0",
+    "babel-runtime": "~5.8.25",
+    "basic-auth": "^1.0.3",
+    "css-loader": "~0.18.0",
+    "commander": "^2.9.0",
+    "express": "^4.13.4",
     "file-loader": "^0.8.5",
-    "history": "~1.9.1",
-    "http-server": "~0.8.5",
     "immutable": "~3.7.5",
     "immutable-devtools": "~0.0.4",
-    "jest-cli": "^0.7.1",
+    "json-file-plus": "^3.2.0",
     "js-beautify": "~1.5.0",
     "marked": "^0.3.5",
     "node-sass": "~3.4.2",
+    "package-json": "^2.3.1",
     "parse": "1.6.14",
     "prismjs": "~1.2.0",
     "react": "^0.14.0",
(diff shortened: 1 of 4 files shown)


flovilmart:
@mnoble I've fixed it with #217

package.json
```diff
@@ -62,17 +59,24 @@
     "sass-loader": "~3.1.2",
     "style-loader": "~0.12.3",
     "svg-prep": "~1.0.0",
-    "transform-jest-deps": "^2.1.0",
     "webpack": "~1.12.0"
   },
+  "devDependencies": {
+    "history": "~1.9.1",
```
flovilmart: history should not be dev...

package.json
```diff
@@ -62,17 +59,24 @@
...
+    "http-server": "~0.8.5",
```
flovilmart: nor that.

flovilmart:
While this looks really good, that makes all those dependencies get installed while you use NPM, which is something we obviously want to avoid as we distribute the package with solely the right folders.

package.json
```diff
@@ -62,17 +59,24 @@
...
   "scripts": {
     "dev": "node ./Parse-Dashboard/index.js & webpack --config webpack/build.config.js --devtool eval-source-map --progress --watch",
     "dashboard": "node ./Parse-Dashboard/index.js & webpack --config webpack/build.config.js --progress --watch",
+    "dist": "npm run build && cp -r production/bundles Parse-Dashboard/public/bundles",
     "pig": "http-server ./PIG -p 4041 -s & webpack --config webpack/PIG.config.js --progress --watch",
     "build": "NODE_ENV=production webpack --config webpack/production.config.js && webpack --config webpack/PIG.config.js",
     "test": "NODE_PATH=./node_modules jest",
     "generate": "node scripts/generate.js",
     "prepublish": "webpack --config webpack/publish.config.js --progress",
+    "postinstall": "if [ -n \"$POST_INSTALL_BUILD\" ]; then npm run dist; fi",
```
flovilmart: you could use: "heroku-postbuild" (as described [here](https://devcenter.heroku.com/articles/nodejs-support#customizing-the-build-process))

mnoble:
@flovilmart Yea, some of those deps might just be copy/pasta errors.

So is the general approach to not specify a `NODE_ENV`, but explicitly specify it in `start` script or something? That way build has access to all deps+devDeps, while the application runs in `production` mode? Not having the env set in production seems like an anti-pattern to me, but I'm also not necessarily a Node developer, so ¯\_(ツ)_/¯

flovilmart:
I was mistaken about NODE_ENV, this is actually `NPM_CONFIG_PRODUCTION=false` ,NODE_ENV will still be production in that case.

to sum up:
1. for heroku, set `heroku config:set NPM_CONFIG_PRODUCTION=false`, this will fetch the dev dependencies
1. in package.json add: `"heroku-postbuild": "npm run dist"` in `"scripts"`

And that's it'!

mnoble:
Ahh, gotcha. Heh, I actually didn't even realize Hunter added the `heroku-postbuild` stuff to the buildpack. Thanks! :+1:

flovilmart:
testing right now but that seem fine

the post build may not even be needed as prepublish will be run after npm install by npm itself and bundles the package by itself

```
remote:        NPM_CONFIG_LOGLEVEL=error
remote:        NPM_CONFIG_PRODUCTION=false
remote:        NODE_ENV=production
remote:        NODE_MODULES_CACHE=true
```

mnoble:
Yea, the problem was that it compiled everything, but didn't move/symlink it to `Parse-Dashboard/public/bundles`, which is why I added the `dist` script to do so. Also, committing compiled production assets to the repo seems less than ideal.

flovilmart:
yes prepublish runs, now I have a trouble with heroku, will keep you posted

flovilmart:
Ok, so it's all great, but we need to remove --progress in prepublish so it don't break on heroku.

flovilmart:
@mnoble I've remove the --progress that was making heroku build fail. So now the only thing to do I believe is to set the heroku env and you're good to go!

PR #217, opened 2016-04-01
:zap: adds ability to run the dashboard from any mount point

flovilmart:
- fixes allowInsecureHTTP


Parse-Dashboard/app.js
```diff
@@ -83,7 +83,28 @@ module.exports = function(config) {
 
   // For every other request, go to index.html. Let client-side handle the rest.
   app.get('/*', function(req, res) {
-    res.sendFile(__dirname + '/index.html');
+    let url = req.url;
```
flovilmart: infer the mount path

Parse-Dashboard/app.js
```diff
@@ -83,7 +83,28 @@ module.exports = function(config) {
...
+    var mountPathLength = req.originalUrl.length - req.url.length;
+    var mountPath = req.originalUrl.slice(0, mountPathLength);
+    if (!mountPath.endsWith('/')) {
+      mountPath += '/';
+    }
+    res.send(`<!DOCTYPE html>
+      <head>
+        <base href="${mountPath}"/>
+        <script>
+          PARSE_DASHBOARD_CONFIG_URI = "${mountPath}parse-dashboard-config.json";
```
flovilmart: instead of passing the mount, just pass the configURI, that's cleaner

src/components/Icon/Icon.react.js
```diff
@@ -18,7 +18,7 @@ let Icon = ({ name, fill, width, height }) => {
   }
   return (
     <svg {...props} >
-      <use xlinkHref={`/bundles/sprites.svg#${name}`} />
+      <use xlinkHref={`bundles/sprites.svg#${name}`} />
```
flovilmart: no / as it should be relative to base

Parse-Dashboard/app.js
```diff
@@ -83,7 +83,28 @@ module.exports = function(config) {
...
+    let url = req.url;
+    let originalUrl = req.originalUrl;
+    var mountPathLength = req.originalUrl.length - req.url.length;
+    var mountPath = req.originalUrl.slice(0, mountPathLength);
+    if (!mountPath.endsWith('/')) {
+      mountPath += '/';
+    }
+    res.send(`<!DOCTYPE html>
+      <head>
+        <base href="${mountPath}"/>
```
flovilmart: needed to properly route the request

src/dashboard/Dashboard.js
```diff
@@ -117,7 +116,7 @@ class Dashboard extends React.Component {
   }
 
   componentDidMount() {
-    get('/parse-dashboard-config.json').then(({ apps, newFeaturesInLatestVersion = [] }) => {
+    get(this.props.configURI).then(({ apps, newFeaturesInLatestVersion = [] }) => {
```
flovilmart: get the config URI

src/dashboard/index.js
```diff
@@ -17,4 +17,5 @@ import 'babel-polyfill';
 require('stylesheets/fonts.scss');
 installDevTools(Immutable);
 
-ReactDOM.render(<Dashboard/>, document.getElementById('browser_mount'));
+var configURI = window.PARSE_DASHBOARD_CONFIG_URI || '/parse-dashboard-config.json';
+ReactDOM.render(<Dashboard configURI={configURI}/>, document.getElementById('browser_mount'));
```
flovilmart: pass it as a prop, that kinda makes sense as one would be able to mount multiple dash with multiple auth

mamaso:
Good stuff, definitely less intrusive. I like how you inferred the mount path.

 A few issues that you might want to test:
- does it load the phone images on the push tab?
- are there any failing requests in the network tab of developer tools?

flovilmart:
You're free to test!

drew-gross:
Turns out the images in the push page don't work. Otherwise this looks sweet though.

flovilmart:
Yep, also the internal API, need to inject dashboardURL into the apps.

drew-gross:
How would you feel about just injecting the config into the page directly? Then an extra fetch would not be necessary. Also, if you are moving the index.html into the source, can you also delete the actual index.html file?

flovilmart:
Yeah I was thinking about that too.
We can't really remove the endpoint as we use it for authentication purposes and that would expose the keys, but we could rename it to /config.json

drew-gross:
Ah, that is unfortuanate :(

flovilmart:
no I'm getting through it!

drew-gross:
Just noticed the favicon is missing too. Obviously that is lower-pri :p

flovilmart:
Just remains the internal API calls in lib/ParseApp.js

flovilmart:
@drew-gross that should be it!

src/lib/AJAX.js
```diff
@@ -9,8 +9,22 @@ import * as CSRFManager from 'lib/CSRFManager';
...
+export function setBasePath(newBasePath) {
+  basePath = newBasePath || '';
+  if (basePath.endsWith('/')) {
+    basePath = basePath.slice(0, basePath.length-1);
+  }
+}
+
 // abortable flag used to pass xhr reference so user can abort accordingly
 export function request(method, url, body, abortable = false, withCredentials = true, useRequestedWith = true) {
+  if (!url.startsWith('http://')
```
flovilmart: open to suggestions about that, that's the less invasive I've found

drew-gross: Yeah this is fine. The only things that aren't going to go through the Parse sdk are the config loading, and potentially this app creation stuff that is being discussed.

fabiogomezdiaz:
@flovilmart great job with this PR! With this, will we now be allowed to deploy the dashboard to Heroku? If so, how could I go about that? Do I need to set any specific environment/config variables?

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.2](https://github.com/parse-community/parse-dashboard/releases/tag/4.0.0-alpha.2)

parseplatformorg:
🎉 This change has been released in version [4.0.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/4.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [4.0.0](https://github.com/parse-community/parse-dashboard/releases/tag/4.0.0)

PR #220, opened 2016-04-02
allows config geopoints

dcdspace:
Fixes https://github.com/ParsePlatform/parse-dashboard/issues/193


Parse-Dashboard/parse-dashboard-config.json
```diff
@@ -2,4 +2,4 @@
   "apps": [
 
   ]
-}
+}
```
flovilmart: can you revert that change?

dcdspace: done

flovilmart: thx! I'll let @drew-gross review and merge as this is his part

PR #224, opened 2016-04-04
allow to set a custom host

### LINKED ISSUE #222: Option to run dashboard on localhost only

- [✔] You're running version >=2.1.4 of Parse Server.

- [✔] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Ubuntu 15.10
parse-dashboard v1.0.7 installed with `npm install -g parse-dashboard`

#### Steps to reproduce
`parse-dashboard --config myconfig.json`

#### Logs/Trace
```
The dashboard is now available at http://localhost:4040/
```
```
$ sudo netstat -plaunt | grep 4040
tcp6       0      0 :::4040                 :::*                    LISTEN      32108/node
```
------------------------

Hello,

Is it possible to have the dashboard listen on localhost only?
So far, the app listens on `::` (or IPv4 `0.0.0.0`) by default, which makes the message `available at http://localhost:4040/` confusing.

A possibility would be to have an `--host` option, and/or a `HOST` env var (similarly to the option for the port) and pass it directly to express's `listen()` function along with the port.

My goal is to run the dashboard behind a HTTPS nginx proxy, and by listening on localhost only I can better restrict the access.

hpello:
Fixes  #222

### DIFF
--- a/Parse-Dashboard/index.js
+++ b/Parse-Dashboard/index.js
@@ -18,11 +18,13 @@ program.option('--masterKey [masterKey]', 'the master key of the app you would l
 program.option('--serverURL [serverURL]', 'the server url of the app you would like to manage.');
 program.option('--appName [appName]', 'the name of the app you would like to manage. Optional.');
 program.option('--config [config]', 'the path to the configuration file');
+program.option('--host [host]', 'the host to run parse-dashboard');
 program.option('--port [port]', 'the port to run parse-dashboard');
 program.option('--allowInsecureHTTP [allowInsecureHTTP]', 'set this flag when you are running the dashboard behind an HTTPS load balancer or proxy with early SSL termination.');
 
 program.parse(process.argv);
 
+const host = program.host || process.env.HOST || '0.0.0.0';
 const port = program.port || process.env.PORT || 4040;
 const allowInsecureHTTP = program.allowInsecureHTTP || process.env.PARSE_DASHBOARD_ALLOW_INSECURE_HTTP;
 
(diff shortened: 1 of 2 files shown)


flovilmart:
I think host is misleading for the beginners as this refers to the bound IP address.

Also, no need to update the address in the README. Locallhost will still be working fine. Not sure 0.0.0.0 will as this represents nothing much for the local machine

hpello:
Thanks for your answer.

Sure. Any suggestion? What about `--address` and `ADDRESS` then?

I see. I had changed the address in the README (as well as in the console.log message) in order not to confuse users, who could think their dashboard is not accessible from the outside (`localhost`) when it actually is (`0.0.0.0`).
But I'll leave `localhost` if you prefer.

flovilmart:
It's just that when people will see: Dashboard running on 'http://0.0.0.0/dashboard' they will paste that on their web browser and won't be able to access it

hpello:
Well, with the current version 1.0.7, it works well for me on Firefox and Chrome :
screen shot 2016-04-05 at 15 48 21

flovilmart:
ok then

hpello:
:+1: 

And for the `adress`/`host` flag and env var?
I liked the `host` one, as it will work with any host, such as `localhost`, `192.168.1.16` or `mywebsite.com`, but I'm fine with what you guys would like.

flovilmart:
we can stick with host, @peterdotjs what do you think?

drew-gross:
I also prefer `localhost` but I'm just going to merge this and then update the docs.

Issue #235, opened 2016-04-06
Dashboard cannot use a mount path that is a prefix of '/parse-dashboard-config.json'


mamaso:
https://github.com/ParsePlatform/parse-dashboard/pull/217/files#diff-8191c30dbcedec21adc7973ea995cdf5R25

This line will match the routes /parse, /parse-dashboard, /parse-dashboard-config, etc and fail to add the mount path to that request.

A bit more context:

1. run dashboard at localhost:4040/parse
2. dashboard makes GET /parse-dashboard-config.json request
3. above line matches /parse mount path to /parse in /parse-dashboard-config.json
4. request is made to localhost:4040/parse-dashboard-config.json instead of the correct localhost:4040/parse/parse-dashboard-config.json

Potential fix:

Instead of 
```js
!url.startsWith(basePath)) {
```

use

```js
!url.startsWith(basePath + '/')
```

mamaso:
Potential fix:

Instead of 
```js
!url.startsWith(basePath)) {
```

use

```js
!url.startsWith(basePath + '/')
```

hpello:
:+1:  I got the same problem.

In other words, the current version forbids one to set a mount path that is a prefix of '/parse-dashboard-config.json'
e.g. '/parse', or '/parse-dashboard'

PR #237, opened 2016-04-06
allow to set a custom mount path

hpello:
Hello,

While we are at it, could we also add an option for the mount path ? This may be useful when running the dashboard behind an NGINX proxy, for instance.

These changes should enable one to set a mount path for the dashboard, e.g.

```
$ parse-dashboard --config myconfig.json --mountPath /dashboard

The dashboard is now available at http://0.0.0.0:4040/dashboard
```

As stated in #235, only mount paths that are not prefixes of `"parse-dashboard-config.json"` seem to work so far...


funkenstrahlen:
when will this feature be released?

drew-gross:
Just waiting for a Parse Server release that includes Logs support, then I will do a dashboard release.

PR #240, opened 2016-04-07
Add basic support for logs viewing

drew-gross:
This is fairly incomplete but probably valuable regardless - at least you can see the logs. We can upgrade level based filtering and so-on later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "Parse Config",
     "API Console",
     "Class Level Permissions Editor",
-    "Send Push Notifications"
+    "Send Push Notifications",
+    "Logs Viewer"
   ],
   "description": "The Parse Dashboard",
   "keywords": [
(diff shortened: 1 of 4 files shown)


tanmays:
@drew-gross Any guide on how to set this up or should it be working without any config? I'm on `parse-server` v2.2.7 and latest `parse-dashboard` master and I'm still unable to see any logs. 

Is there any specific syntax to make logs show up in dashboard. I'm using the traditional `console.log("Message");` syntax.

drew-gross:
The logs that show up are the ones output by the `LoggingAdapter` in Parse Server. Set `VERBOSE=1` to log every request and response. We don't have good docs yet for how to set up logging from your cloud code, unfortunately. The default logger is `winston` so check out the npm package. If you get it working in a way you like, a PR with some docs would be much appreciated :)

notiles:
Hello, Hi also try to get log from my cloud code in dashboard but it's not working for now, so if there is a best practice to follow, let me know. Tx

PR #248, opened 2016-04-11
Tweak Dockerfile (Fixes #136)

### LINKED ISSUE #136: SyntaxError: expected expression, got '<'
I am running parse-dashboard with Docker, and I have connected to the url http://172.17.0.3:4040, where the service is running.

I am not able to see anything, and the only console error in Firefox says: _SyntaxError: expected expression, got '<'_, when the navigator tries to fetch the file http://172.17.0.3:4040/bundles/dashboard.bundle.js

When I open the file dashboard.bundle.js I see:
```
<!DOCTYPE html>
<html>
  <title>Parse Dashboard</title>
  <body>
    <div id="browser_mount"></div>
    <script src="/bundles/dashboard.bundle.js"></script>
  </body>
</html>

```
I don't understand why is this, because the file should be in javascript, I guess I am doing something wrong but i don't know what is it.

Other question: is it possible to run without docker? I have not used docker until today and something like node index.js would be easier to test

kkc:
1. using npm run dashboard to generate bundles/dashboard.bundle.js
2. shrink docker image size

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
-FROM node:4.3.2
-ADD package.json /src/package.json
-RUN cd /src && npm install
-ADD . /src
+FROM node:4.4.2
 WORKDIR /src
-ENTRYPOINT ["npm", "start", "--"]
+ADD . /src
+RUN cd /src \
+ && npm install \
+ && npm run build \
+ && npm cache clear \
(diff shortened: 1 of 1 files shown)


drew-gross:
I'm not a docker user but this seems like it should work, and thats better than what we have now which I know won't work...

Issue #230, opened 2016-04-05
Feature Requests


TylerBrock:
Hey Drew,

I was wondering if there were any plans to implement:
- [x] a refresh button in the data browser (useful when running scripts from the command line)
- [x] a way to back out of viewing a relation (go back to the referring view)
- [ ] the analytics portion of the dashboard (on top of mongodb)

I'd be happy to help with either but I just wanted to feel out how warm the project is to taking something like that on (or if something was already in the works)

drew-gross:
We are definitely welcome to new features! A refresh button for the data browser sounds great - and even better if works for every page where it makes sense. So does backing out of relation views, I think that one could be done fairly easily using the router/history stuff we already use.

Analytics hasn't really a big focus of Open Source Parse because we feel that there are a lot of other analytics products out there that do a better job but if you are interested in helping we would welcome PRs.

ido43210:
Regarding the analytics, the main benefit for using parse analytics is the automation part in the client side (using the sdk). So we won't need to add more code.
And if the analytics part stay out of focus will it be a good idea to have a way for hooking the analytics code in the client SDK so we will only need to implement it in a single class.

meoz:
+1 Refresh button.
Need to refresh the whole page to see updates.

TylerBrock:
I actually didn't mean the app analytics but more like how many requests is the app doing, how much disk are we using up, etc...

drew-gross:
Ah yes, we have been calling that type of thing "ops-ish analytics" internally. Yeah, this is also something we think would be very valuable. The `serverInfo` endpoint in Parse Server was actually designed to be extended to "ops-ish analytics" so adding disk/memory/cpu usage there would be a good first step, with a corresponding place to view it in the dashboard. How to handle multiple Parse Servers behind a load balancer is an open question, but one we don't necessarily need to answer right away. One solution might be to give the dashboard configuration a list of servers and it bypasses the load balancer to get the data, another might be to have one (or many) Parse Servers in "monitoring mode" that collect this info. @TylerBrock what do you think? Would you be willing to get started in adding support for some basic stuff in the `serverInfo` endpoint of Parse Server?

TylerBrock:
Serving up some code to check off numero uno. Stay tuned for the others.

MobileVet:
This is something that Backendless does really well, showing what calls are made, how often and how long each call takes.  This can really help root out time outs etc.

jdtzmn:
[...]

PR #261, opened 2016-04-14
Add refresh button to data browser

TylerBrock:
This isn't quite ready but I'm soliciting input if you wouldn't mind taking a look.


drew-gross:
The code looks good, can you try to match the look and feel of the existing buttons in the data browser though? We have a refresh icon you can use with the `Icon` component.

TylerBrock:
Note: this is related to requests in #230

drew-gross:
Nice, now we're looking good :)

PR #268, opened 2016-04-16
In some environment, req.sequre is always false even https

gimdongwoo:
If req.secure is false then do 2nd check logic.
Because, in some environment, req.sequre is always false even https.

Thank you!!


flovilmart:
Use the allowInsecureHTTP option to accomplish that as a malicous request can be forged with x-forwarded-proto = https can be forged.

PR #274, opened 2016-04-18
Allows showing app icons (or placeholder)

### LINKED ISSUE #269: Display app icons

- [x] You're running version >=2.1.4 of Parse Server.

Currently, the dashboard doesn't show yet the app icons. Ideally, it would be very helpful to have the icons to better visualize each app, specially if you have a lot of them.

I think that we could add a `iconImage` or `icon` parameter to each app in the dashboard-config.json file, and use a placeholder if no icon (can we use the one that is currently shown in the Parse.com dashboard?). Any special recommendations for making a PR on this?

natanrolnik:
Thanks @drew-gross for pointing me the direction. 
See the Readme diff for more info.
This should close #269 

dashboardwithicons

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,23 @@ You can also manage apps that on Parse.com from the same dashboard. In your conf
 }
 ```
 
+Parse Dashboard also supports adding an optional icon for each app, so you can identify them easier in the list. To do so, you *must* use the configuration file, define an `iconsFolder` in it, and define the `iconName` parameter for each app (including the extension). The path of the `iconsFolder` is relative to the configuration file. To visualize what it means, in the following example `icons` is a directory located under the same directory as the configuration file:
+
+```
+{
+  "apps": [
+    {
+      "serverURL": "http://localhost:1337/parse",
+      "appId": "myAppId",
+      "masterKey": "myMasterKey",
+      "appName": "My Parse Server App",
+      "iconName": "MyAppIcon.png",
+    }
+  ],
+  "iconsFolder": "icons"
+}
+```
+
 ![Parse Dashboard](.github/dash-shot.png)
 
 # Advanced Usage
--- a/src/dashboard/Apps/AppsIndex.react.js
+++ b/src/dashboard/Apps/AppsIndex.react.js
@@ -76,7 +76,9 @@ let AppCard = ({
     </div>;
 
   return <li onClick={canBrowse}>
-    {icon ? <a className={styles.icon}><img src={icon} /></a> : null}
+    <a className={styles.icon}>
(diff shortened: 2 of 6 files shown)


src/dashboard/Apps/AppsIndex.react.js
```diff
@@ -156,7 +158,7 @@ export default class AppsIndex extends React.Component {
         <ul className={styles.apps}>
           {apps.map(app =>
             app.name.toLowerCase().indexOf(search) > -1 ?
-              <AppCard key={app.slug} app={app} /> :
+              <AppCard key={app.slug} app={app} icon={this.props.iconsFolder && app.icon ? this.props.iconsFolder + '/' + app.icon : null}/> :
```
drew-gross: I think we probably want to not have the client know about the `iconsFolder`. I'd suggest you use a consistent path like `/appicons/${appId}` and then in `app.js` add a `appicons` route and serve up the image based on the `appId` in the parameter. Then you don't need to worry about the icons folder being publicly accessible.

README.md
```diff
@@ -60,6 +60,23 @@ You can also manage apps that on Parse.com from the same dashboard. In your conf
 }
 ```
 
+Parse Dashboard also supports adding an optional icon for each app, so you can identify them easier in the list. To do so, you will need to define an `iconsFolder` in the config file, and define the `iconName` parameter for each app (including the extension). This folder needs to be located in the root of your project. For example, if you are building from this repo's source, the folder should be inside `Parse-Dashboard`; if you are using npm, it should be in your root directory. If no icon is set for a specific app, the dashboard will show a placeholder.
+
```
drew-gross: I think the docs should probably talk about 2 use cases: installing globally with `npm install -g parse-dashboard` and installing as part of a project and mounting as an express app. Installing from source is now only recommended if you are working on the dashboard so I don't think we need to call out that use case in the docs.

drew-gross:
Code looks good now, I think (although I'm torn on whether to use `__dirname` or not), just need a little documentation update now.

natanrolnik:
@drew-gross I've made the changes you suggested. The requirement is that a config file must be used (not enough passing the `appId` and `masterKey` from command line), and as we talked, the `iconsFolder` is relative to the path of the config file. I've tested locally and got the desired result.

drew-gross:
Sweet! This is looking really good now :) once tests pass I will merge.

natanrolnik:
1.0.10 coming 😁 ?

drew-gross:
I'd like to see if I can fix #276 first. Need some more info though. If I don't hear anything for a couple days I will do 1.0.10 regardless.

natanrolnik:
👍 cool
Thanks for the review - I learned a lot with it!

natanrolnik:
@drew-gross it looks like #276 is closed, right?

drew-gross:
Yep. Will do a release shortly.

Issue #250, opened 2016-04-12
Running parse-dashboard at Heroku?


onosol:
Is there an example of running parse-dashboard at Heroku like the parse-server-example GitHub project?

grassland-curing-cfa:
A working example would be really helpful!

mattgoldspink:
I just set this up - it's very straight forward. I'm assuming you already have a Heroku account. Run the following:

$ mkdir my-parse-dashboard
$ cd my-parse-dashboard
$ git init
$ npm init
$ npm install --save parse-dashboard

For npm init just enter anything that makes sense. Now you'll need to create 2 files:

1) parse-dashboard-config.json - this should follow the pattern listed here: https://github.com/ParsePlatform/parse-dashboard
2) Procfile - The contents of this should be one line:

web: ./node_modules/.bin/parse-dashboard --config ./parse-dashboard-config.json --allowInsecureHTTP

Now you need to commit all this to git:

$ git add package.json Procfile parse-dashboard-config.json
$ git commit -m "My First Commit"

Now install the Heroku toolbelt if you haven't already: https://toolbelt.heroku.com/
Then run:

$ heroku login
$ heroku create
$ git push heroku master
$ heroku ps:scale web=1
$ heroku open

That should be it

onosol:
mattgoldspink, This is great, thanks! The procfile was the missing piece for me. I will try this out today.

drew-gross:
Thanks for the help @mattgoldspink! Do you think you might be able to add some formatting and open a PR adding those instructions to the readme? They look really useful.

onosol:
@mattgoldspink, Just tested and this worked perfectly! Thanks again.

natanrolnik:
Thank you, @mattgoldspink . That helped me!

gateway:
Hmm, question, is ir procfile and what folder does this go into?

cherukumilli:
# parse-dashboard-example
If anyone is interested then please check out and give me feedback about my **parse-dashboard-example** project. It is similar to the **parse-server-example**.

Please check it out at: `https://github.com/cherukumilli/parse-dashboard-example`

gateway:
@cherukumilli thanks for doing this, its much appreciated! Ill give it a test because most of my stuff will probably be in AWS.

natanrolnik:
@gateway as it's not intended to run for production to users, you could easily deploy to heroku in their free dyno

Somkid2555:
[...]

PR #282, opened 2016-04-20
Add link to how to run on Heroku

mattgoldspink:
Also added a new section to the Readme so that other platform setups can be contributed either as Wiki pages or blog posts

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,12 @@ docker run -d -p 80:8080 -v host/path/to/config.json:/src/Parse-Dashboard/parse-
 
 If you are not familiar with Docker, ``--port 8080`` will be passed in as argument to the entrypoint to form the full command ``npm start -- --port 8080``. The application will start at port 8080 inside the container and port ``8080`` will be mounted to port ``80`` on your host machine.
 
+## Running Parse Dashboard elsewhere
+
+It's relatively straightforward to setup Parse Dashboard on any Docker or Node.js hosting provider:
+
+- [Parse Dashboard on Heroku in 3 steps](https://www.codementor.io/nodejs/tutorial/deploy-parse-dashboard-on-heroku)
+
 ## Contributing
 
 We really want Parse to be yours, to see it grow and thrive in the open source community. Please see the [Contributing to Parse Dashboard guide](CONTRIBUTING.md).


natanrolnik:
@mattgoldspink do you think you could squash the commits in one single commit?

noder199:
Is there a way to stop somebody from just visiting the link and messing with your database?

drew-gross:
Since your guide uses the --allowInsecureHTTP option, can you include in your guide how to ensure that Heroku does not allow HTTP? Using HTTPS only is incredibly important for the dashboard.

mattgoldspink:
@natanrolnik Sure - let me get that fixed.

@drew-gross So I added that flag because when on Heroku the proxy forwards requests over http and hence you end up with.

image

I'll file a seperate ticket for this, but we should perhaps either add another flag which enables http if the "X-Forwarded-Proto" is https, or extend the existing flag to allow http requests if the "X-Forwarded-Proto" header is set.

@noder199 Can you explain a little bit more about what you mean? The example I gave sets up the user/pass in the config ensuring users have to log in. Are you talking about something else?

flovilmart:
X-Forwarded-Proto is not safe, any HTTP client could set it and force the server to serve on HTTP thinking it's safe...

mattgoldspink:
@flovilmart Agreed - let me open a separate ticket to discuss this on.

drew-gross:
Using the --allowInsecureHTTP flag is find if you have SSL termination somewhere earlier in your request handling, and that layer refuses non HTTPS. If you can explain in your guide why it's safe to use in this case, I'd be happy to merge :)

gateway:
I added my way to get the dashboard running with Heroku .. its up and running fine.  #250

mattgoldspink:
@drew-gross Added an "Important Note" section to the article explaining that in general --allowInsecureHTTP flag shouldn't be set, but in this case it is valid. Also included a link to the open ticket about improving for proxies. Let me know if it needs tweaking.

cherukumilli:
# parse-dashboard-example
If anyone is interested then please check out and give me feedback about my **parse-dashboard-example** project. It is similar to the **parse-server-example**.

Please check it out at: `https://github.com/cherukumilli/parse-dashboard-example`

drew-gross:
Closing due to lack of activity

Issue #290, opened 2016-04-22
Clarify README - Add more sections and jump links


andrewcbancroft:
When I first began to work with Parse Dashboard, I had trouble figuring out which sections of the README applied to local development, and which applied to a production scenario.  Obviously some areas of the README apply to both.  I've been analyzing the README to try and figure out how to clarify it a bit, and I wondered if something like the following proposal might be a well-received contribution...

I'd be willing to help clarify the README by adding more sections to the tune of the outline specified below.  I'd also propose (and would be willing to add) "jump-to" links at the top of the README so that newcomers can easily find what they're most interested in and jump right to that section.  

Here's the outline I'm thinking about...

-----

# Local Installation
...
## Configuring Parse Dashboard
...
### Managing Multiple Apps
...

### Other Configuration Options
...
#### App Icon Configuration

Then, instead of "Advanced Usage", the next section could talk about deployment:

# Deploying Parse Dashboard
...
## Preparing for Deployment
"ensure URLs are not `localhost`..."

## Security Considerations
... SSL ...
### Configuring Basic Authentication
...
### Separating App Access Based on User Identity
(I'm honestly not sure what this does.  Is it saying that user1 has access to the apps in the array below it? ... Maybe the example could be expanded a little so it's a little more clear what scenario this applies to)

## Run with Docker
...
# Contributing
...

-----

I think having more sections helps break up the page enough to easily spot what you're looking for when scrolling down the page, too.  Thoughts on this?

drew-gross:
Sounds good to me. Separating apps access based on user means exactly what you think, each user can only see the apps in their array. More docs would always be nice :)

pivanov:
+1

PR #294, opened 2016-04-22
Update README.md with additional headers and TOC

andrewcbancroft:
Added additional headers and reorganized README so that it's easier to follow from a local installation / deployment scenario perspective.
Added a table of contents with "jump-to" links for easier navigation of the document.

Closes Issue #290.


drew-gross:
Looks good!

PR #308, opened 2016-04-30
Allow trust proxy if allowInsecureHTTP

yongjhih:
`app.enable('trust proxy');` when you are running the dashboard behind an HTTPS load balancer or proxy.

### DIFF
--- a/Parse-Dashboard/index.js
+++ b/Parse-Dashboard/index.js
@@ -105,6 +105,7 @@ p.then(config => {
 
   const app = express();
 
+  if (allowInsecureHTTP) app.enable('trust proxy');
   app.use(mountPath, parseDashboard(config.data, allowInsecureHTTP));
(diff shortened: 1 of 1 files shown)


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @drew-gross, @Vortec4800 and @gimdongwoo to be potential reviewers.

drew-gross:
We actually have plans to make the `allowInsecureHTTP` option actually just set `trust proxy`, instead of an approach like this. Either way, we definitely don't want to `trust proxy` by default, as we don't always know if there is a trusted proxy.

yongjhih:
Ok, I see. Thanks for your explaining.

facebook-github-bot-1:
@yongjhih updated the pull request.

drew-gross:
Awesome! Thanks :)

PR #390, opened 2016-06-01
Release 1.0.13 and fix bug when push alert is an object

drew-gross:


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @JeremyPlease, @flovilmart and @durunvo to be potential reviewers.

src/dashboard/Push/PushIndex.react.js
```diff
@@ -145,7 +145,7 @@ let getPushName = (pushData) => {
       payload = JSON.parse(payload);
     } catch(e) { }
     if(payload){
-      return payload.alert ? payload.alert : payload;
+      return payload.alert ? payload.alert.toString() : payload;
```
JeremyPlease: @drew-gross If `payload.alert` is an object then the push name will be  "[object Object]". Can we do the following instead to handle objects and strings:

```
return payload.alert ? JSON.stringify(payload.alert) : payload;
```

JeremyPlease:
One comment/suggestion. Otherwise, looks good!

drew-gross:
Yeah thats definitely a better idea. Thanks!

facebook-github-bot-8:
@drew-gross updated the pull request.

PR #410, opened 2016-06-10
Fix past pushes page

drew-gross:
When updatedAt is a date object and pushTime is null, this crashes. Workaround by checking `typeof time`, since we don't have local time push in Parse Server anyway.


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @JeremyPlease and @durunvo to be potential reviewers.

Issue #124, opened 2016-03-10
Delete all rows in collection


flessard:
Where is the delete all rows for a collection?

- [X] You're running version >=2.1.4 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

DazChong:
+1 I use mongo to drop collection.

AmbroiseCollon:
+1 would be useful for testing

miracle7:
+1

hl4hck:
+1

adampietrasiak:
+1

JustinDMoore:
+1

adampietrasiak:
Maybe at least some "Select all on screen" option on data browser?

Some websites have also such option that if I press mouse button on some checkbox and then (without releasing it) I move mouse to other checkboxes, they get selected too. That would save me a lot of clicking.

quaz3l:
Can someone explain why this checkbox in the header of collections is disabled?
![](https://i.imgur.com/1xXq87s.png)
I assume it would be for performance reasons? But If you inspect element, and enable it, it works and selects every element in the collection. It doesn't actually delete everything because it is probably a special case in code, but I could fix that to work as if every row was selected manually.

We could also just limited it to remove only 50 rows at a time to prevent overloading the server. Would this work?

drew-gross:
We disabled it because it was confusing some people, and a few people accidentally deleted their entire collection. We were eventually going to change it so it selected only the  visible rows, but we never got around to that.

andrew8712:
>We disabled it because it was confusing some people, and a few people accidentally deleted their entire collection.

Well, I believe that disabled *select all* checkbox abuses people even more.

If few people accidentally deleted their entire collection, that's their problem. Do we have to find a workarounds because they are not smart enough to avoid pressing `Delete` button when all rows are selected?

JustinDMoore:
Agree with @andrew8712 . Aren't they presented with a confirmation to delete, forcing them to type in the class name to confirm?

grassland-curing-cfa:
Deployed the latest Parse-server 2.2.15 and Parse-dashboard 1.0.14.
When I ran "Delete all rows" from the Edit menu list, it gave me a "404 Not Found" error when it tried to run "DELETE clear", which was identified from my Firebug console.
[...]

PR #413, opened 2016-06-11
Delete all rows in class

Marco129:
Related to #124 
Merge only if ParsePlatform/parse-server#2032 accepted!

### DIFF
--- a/src/lib/ParseApp.js
+++ b/src/lib/ParseApp.js
@@ -357,8 +357,13 @@ export default class ParseApp {
   }
 
   clearCollection(className) {
-    let path = `/apps/${this.slug}/collections/${className}/clear`;
-    return AJAX.del(path);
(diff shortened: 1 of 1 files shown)


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @drew-gross, @JeremyPlease and @gavrix to be potential reviewers.

src/lib/ParseApp.js
```diff
@@ -357,8 +357,13 @@ export default class ParseApp {
   }
 
   clearCollection(className) {
-    let path = `/apps/${this.slug}/collections/${className}/clear`;
-    return AJAX.del(path);
+    if (this.serverInfo.parseServerVersion == 'Parse.com') {
+      let path = `/apps/${this.slug}/collections/${className}/clear`;
```
drew-gross: We don't actually need to keep the version that goes to Parse.com around, it doesn't work anyway.

Issue #415, opened 2016-06-11
Dashboard app list sort order


185driver:
When the dashboard loads and displays the app list, the sort order of the apps seems to vary each time. It would be nice if the list had a consistent alpha default sort order.

I'm managing several apps using the same parse server and on a single dashboard, if that's a factor.

Thanks for considering.

drew-gross:
FIxed y=by #417

185driver:
Thank you, sir.

PR #417, opened 2016-06-12
Sorting applications by name in the dashboard, #415

bohemima:
Always sorts before returning apps() - should probably live somewhere else but I'm unable to find a suitable location for it. Solves #415

### DIFF
--- a/src/lib/AppsManager.js
+++ b/src/lib/AppsManager.js
@@ -18,6 +18,9 @@ const AppsManager = {
   },
 
   apps() {
+    appsStore.sort(function(app1, app2) {
+      return app1.name.localeCompare(app2.name);
+    });
     return appsStore;
   },
 


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @drew-gross to be a potential reviewer.

drew-gross:
LGTM. I might have put this in the place where the apps are rendered, but this works too.

bohemima:
Looking a bit more at this and there is actually some sorting-code in https://github.com/ParsePlatform/parse-dashboard/blob/master/src/dashboard/Apps/AppsIndex.react.js#L137. However, createdAt seems to be null everywhere, so that doesnt have any effect.

I'll have a look at refining my code and possibly adding other sorting options. (User count, Installation count or appName?)

drew-gross:
Yeah createdAt only existed on dashboard.parse.com, in this Parse Dashboard we don't know when apps were created.

PR #422, opened 2016-06-15
Update endpoint url

Marco129:
Update endpoint url for deleting all rows in dashboard
Maybe we can merge later, i.e. just before next release of parse server, in case there are changes again


facebook-github-bot-1:
By analyzing the blame information on this pull request, we identified @drew-gross, @JeremyPlease and @gavrix to be potential reviewers.

drew-gross:
I think this is fine. I doubt there will be more changes, and if there are we can add them later pretty easily.

PR #436, opened 2016-06-26
ConfigStore now uses the master key to fetch the Parse config

Gameleon12:
The ConfigStore now uses the master key to fetch the Parse config. This is to prevent an issue
where the fetch command would not complete because of a `403 Unauthorized` error. This issue led to a bug on the dashboard's config page causing previously saved config entries to not load properly.

### DIFF
--- a/src/lib/stores/ConfigStore.js
+++ b/src/lib/stores/ConfigStore.js
@@ -22,8 +22,13 @@ function ConfigStore(state, action) {
   action.app.setParseKeys();
   switch (action.type) {
     case ActionTypes.FETCH:
-      return Parse.Config.get().then(({ attributes }) => {
-        return Map({ lastFetch: new Date(), params: Map(attributes) });
+      return Parse._request(
+        'GET',
+        'config',
+        {},
+        { useMasterKey: true }
+      ).then(({ params }) => {
+        return Map({ lastFetch: new Date(), params: Map(params) });
       });
     case ActionTypes.SET:
       return Parse._request(


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @drew-gross to be a potential reviewer.

drew-gross:
Do you know what caused the unauthorized error? Getting config should never require the master key.

Gameleon12:
Alright. Altough even while the master key isn't required, a normal key is required, I assume? Because when taking a closer look it seems not a single key is being sent with the config GET request using version 1.0.14 of the Dashboard.

Example request headers of config GET request on Parse dashboard page:

```
Host: OMITTED
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:47.0) Gecko/20100101 
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-GB,en;q=0.5
Accept-Encoding: gzip, deflate, br
Content-Type: text/plain
Referer: OMITTED
Content-Length: 162
Connection: keep-alive
```

And the body that goes with the request:

{"_method":"GET","_ApplicationId":"OMITTTED","_ClientVersion":"js1.6.14","_InstallationId":"OMITTED"}

I will look further into this and maybe find a better workaround for this than just using the master key.

Gameleon12:
It looks like the issue was unrelated to any of the parse-dashboard implementation: It seems there was an issue with the encoding of my parse dashboard config file which caused the javascript and rest key to not be read properly. I have fixed the encoding and now the keys are properly sent with the request.

This pull request can therefore be discarded.

PR #446, opened 2016-07-05
[Feature] Improve DataBrowser relation viewer

Beingbook:
relation viewer
* has url statement
* can go back
* can add a row
* can bring existing row to attach
* fixed filter

### DIFF
--- a/src/dashboard/Push/PushIndex.react.js
+++ b/src/dashboard/Push/PushIndex.react.js
@@ -140,15 +140,17 @@ let getPushName = (pushData) => {
       <strong>{title}</strong>
     );
   } else {
-    let payload = pushData[PushConstants.PAYLOAD_FIELD];
+    let payload = pushData[PushConstants.PAYLOAD_FIELD] || '';
     try {
       payload = JSON.parse(payload);
     } catch(e) { }
-    if (payload) {
+    if (typeof payload === 'object') {
       if (typeof payload.alert === 'string') {
         return payload.alert;
       }
-      return payload.alert ? JSON.stringify(payload.alert) : payload;
+      return payload.alert ? JSON.stringify(payload.alert) : JSON.stringify(payload);
+    } else {
+      return '';
     }
   }
 }


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @drew-gross, @TylerBrock and @gavrix to be potential reviewers.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -59,26 +61,45 @@ export default class Browser extends DashboardView {
...
+      let filters = new List();
       if (this.props.location.query && this.props.location.query.filters) {
-        let filters = new List();
         let queryFilters = JSON.parse(this.props.location.query.filters);
         queryFilters.forEach((filter) => {
           filters = filters.push(new Map(filter));
         });
+      }
+
+      const { className, entityId, relationName } = this.props.params;
```
flovilmart: Is that a re-declaration here?

Beingbook: I was just blinded for shortcut. Remove above?

flovilmart: If above is not used yes :)

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -59,26 +61,45 @@ export default class Browser extends DashboardView {
...
     .then(() => this.handleFetchedSchema());
+
+    const { className, entityId, relationName } = this.props.params;
+
     if (!this.props.params.className && this.props.schema.data.get('classes')) {
       this.redirectToFirstClass(this.props.schema.data.get('classes'));
     } else if (this.props.params.className) {
+      let filters = new List();
       if (this.props.location.query && this.props.location.query.filters) {
-        let filters = new List();
```
flovilmart: This change dont seem necessary

src/dashboard/Data/Browser/Browser.react.js
Beingbook: Can I force-push to fix it?

flovilmart: No need to force push, we squash upon merge.

src/dashboard/Data/Browser/Browser.react.js
Beingbook: I'm just curious at that how to change this? new commit? or force-push after rebase? what's right?

flovilmart:
The overall code look great, and the feature is 💯. I'm no react expert but it seems we're bloating the Browser component with a lot of corner case Logic for the relations. Can we compose or inherit instead? What do you think @peterdotjs @drew-gross

Beingbook:
@flovilmart please make it better.

src/dashboard/Data/Browser/Browser.react.js
flovilmart: Just add a new commit, we prefer not encouraging force pushes :)

flovilmart:
I'm asking for advice here. Want to make sure we keep the code base maintainable. Otherwise I'm impressed! 👍

Beingbook:
I want to submit a PR which makes this project more maintainable but this project is too big to patch this myself. There are too many legacies...

src/dashboard/Data/Browser/AttachRowsDialog.react.js
```diff
@@ -0,0 +1,95 @@
+import * as Filters from 'lib/Filters';
+import Modal from 'components/Modal/Modal.react';
+import React from 'react';
+import Field from 'components/Field/Field.react';
+import Label from 'components/Label/Label.react';
+import TextInput from 'components/TextInput/TextInput.react';
+import { List, Map } from 'immutable';
```
Beingbook: This component doesn't need immutable structures. Please remove this before squashing.

flovilmart: That's your code :) you should do the changes

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -370,7 +379,7 @@ export default class Browser extends DashboardView {
     } else {
       const source = this.props.params.className;
       const _filters = JSON.stringify(filters.toJSON());
-      const url = `browser/${source}` + (filters.size === 0 ? '' : `?filters=${encodeURIComponent(_filters)}`);
+      const url = `browser/${source}` + (filters.size === 0 ? '' : `?filters=${(encodeURIComponent(_filters))}`);
```
Beingbook: strange string concat...

src/dashboard/Data/Browser/Browser.react.js
Beingbook: This code is needed to pass filter to fetchRelation below.

facebook-github-bot-2:
@Beingbook updated the pull request.

facebook-github-bot-1:
@Beingbook updated the pull request.

facebook-github-bot-1:
@Beingbook updated the pull request.

src/dashboard/Data/Browser/Browser.react.js
flovilmart: if so it's declared in the wrong place as you exit the `if` block and expect the filters still to be set

peterdotjs:
@Beingbook @flovilmart generally we'll prefer composition over inheritance but in this case Browser already has several states for various dialogs that it seems okay to keep it in the same component but may be worth looking at in a future refactor PR as it gets larger.

RE: flow and eslint, we had discussed around the time it was open sourced but there wasn't enough request for it to justify the refactor. For future maintainability I totally agree that it would be beneficial. eslint will be a sizable task to refactor but with flow we can go with the approach of any new file that is touched to be flowified. If you're interested in taking this on we're totally supportive of it.

Looks great overall though!

src/components/Toolbar/Toolbar.react.js
```diff
@@ -7,31 +7,59 @@
...
-    <div className={styles.title}>
-      <div className={styles.section}>{props.section}</div>
-      <div>
-        <span className={styles.subsection}>
-          {props.subsection}
-        </span>
-        <span className={styles.details}>
-          {props.details}
-        </span>
+const goBack = () => history.goBack();
```
peterdotjs: will this functionality be any different than hitting browser back?

src/dashboard/Data/Browser/AttachRowsDialog.react.js
```diff
@@ -0,0 +1,92 @@
...
+  render() {
+    const {
+      relation,
+      onCancel,
+      onConfirm,
+      errors,
+    } = this.props;
+    let errorStatus;
+    if (errors) {
+      const errorList = [];
```
peterdotjs: nit: `let`

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -472,10 +585,71 @@ export default class Browser extends DashboardView {
...
+    this.setState({
+      showAttachRowsDialog: false,
+    });
+  }
+
+  confirmAttachRows(objectIdsString) {
+    const relation = this.state.relation;
+    const query = new Parse.Query(relation.targetClassName);
+    const parent = relation.parent;
+    this.state.attachRowsErrors = [];
```
peterdotjs: did you mean `setState` ?

Beingbook: this property is used below 634, 644 line

src/dashboard/Data/Browser/Browser.react.js
Beingbook: `if` and `else` both block needs `filters`. So I declared `filters` upper scope.

src/dashboard/Data/Browser/Browser.react.js
peterdotjs: what I mean is you don't want to mutate state directly. ref: https://facebook.github.io/react/docs/component-api.html

src/components/Toolbar/Toolbar.react.js
Beingbook: I know it matters. I just added this because relation viewer has url statement. If a user access relation viewer first, it will matter. But I have no idea to implement this better. Can you suggest about it?

src/dashboard/Data/Browser/AttachRowsDialog.react.js
Beingbook: you means unused variable? I don't know what's `nit`

src/dashboard/Data/Browser/Browser.react.js
Beingbook: I thought `setState` function and mutating state directly flip 180...

src/dashboard/Data/Browser/AttachRowsDialog.react.js
peterdotjs: I meant since `errorList` is being mutated we should use `let` rather than `const`.

src/dashboard/Data/Browser/Browser.react.js
peterdotjs: right so when you want to change the value of any state `setState` should be used. If you're trying to clear the errors we should use `setState` rather than directly modifying it.

src/dashboard/Data/Browser/Browser.react.js
flovilmart: Yeah right, the if block is so big, I didn't see where it ends. Consider extracting maybe

facebook-github-bot-1:
@Beingbook updated the pull request.

facebook-github-bot-1:
@Beingbook updated the pull request.

src/dashboard/Data/Browser/Browser.react.js
Beingbook: Yes, it needs to be separating. I will make some functions as async to make it shorter with including this.

Beingbook:
what's next step?

peterdotjs:
@Beingbook I haven't had a chance to patch it and try it out but looks good otherwise. I'll update once I've gotten a chance to test it out.

src/dashboard/Data/Browser/Browser.react.js
Beingbook: linted at 0ee762d

flovilmart:
@Beingbook you're in good hands with @peterdotjs!

src/dashboard/Data/Browser/AttachRowsDialog.react.js
```diff
@@ -0,0 +1,98 @@
+import * as Filters from 'lib/Filters';
```
Beingbook: it is imported but not used.

facebook-github-bot-1:
@Beingbook updated the pull request.

Beingbook:
oops. I did something wrong... I guess rebase and force push made ref tree twisted...

facebook-github-bot-1:
@Beingbook updated the pull request.

facebook-github-bot-1:
@Beingbook updated the pull request.

PR #509, opened 2016-08-25
Made the encrypted passwords an option

dvanwinkle:
Discovered that my previous encrypted password implementation allowed the user to type the hash and be authenticated. I feel like that isn't going to be expected, so, I added an options config option to enable encrypted passwords.


dvanwinkle:
@drew-gross Not sure if there is a way to remove my stupidity, but, I accidentally committed our master keys (which are now reset anyway). If there is a way, can you completely delete this PR?

drew-gross:
Unfortunately it can't be deleted, except by github staff. Sorry.

PR #510, opened 2016-08-25
Made the encrypted passwords an option

dvanwinkle:
Discovered that my previous encrypted password implementation allowed the user to type the hash and be authenticated. I feel like that isn't going to be expected, so, I added an options config option to enable encrypted passwords.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,13 +231,11 @@ You can configure your dashboard for Basic Authentication by adding usernames an
       "pass":"pass"
     }
   ],
-  "options": {
-    "useEncryptedPasswords": true | false
-  }
+  "useEncryptedPasswords": true | false
 }
 ```
 
-You can store the password in either `plain text` or `bcrypt` formats. To use the `bcrypt` format, you must set the config `options.useEncryptedPasswords` parameter to `true`.
+You can store the password in either `plain text` or `bcrypt` formats. To use the `bcrypt` format, you must set the config `useEncryptedPasswords` parameter to `true`.
 You can encrypt the password using any online bcrypt tool e.g. [https://www.bcrypt-generator.com](https://www.bcrypt-generator.com).
 
 ### Separating App Access Based on User Identity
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -230,11 +230,15 @@ You can configure your dashboard for Basic Authentication by adding usernames an
       "user":"user2",
       "pass":"pass"
     }
-  ]
+  ],
+  "options": {
```
drew-gross: I don't think we need this additional nesting level.

dvanwinkle: I was trying to leave it open for future changes. Would you suggest that I just bring the setting up, or something else?

drew-gross: Exactly. The whole config file is already a dict of options, it doesn't need another dict of options within it.

drew-gross:
Would you mind also adding a test for this? We've had a few annoying regressions recently due to lack of testing so I'd like to start stepping up our game in the department.

Zertz:
Not sure why the bot pinged me, I just contributed to docs once or twice ;-)

facebook-github-bot-7:
@dvanwinkle updated the pull request - [view changes](https://github.com/ParsePlatform/parse-dashboard/pull/510/files/badd45031e369f9fca5e660e04b700d279bebb5c..4f384232f4e59b140eb499f283df61b10b916065)

dvanwinkle:
@drew-gross Moved Authentication into it's own module and added some tests. This is really my first time writing Javascript tests, so, please, be nice in your review ;-)

drew-gross:
Looks solid to me! Thanks for the tests :)

johnnydimas:
Nice feature! Although, I had to add "use strict" to the top of Parse-Dashboard/Authentication.js to get this to build for me. Otherwise it was returning a server error page.

dvanwinkle:
@johnnydimas Most likely you are using an older version of node. Looks like the minimum in the `package.json` is `4.3`. That does require the `"use strict"`, so, I will add that and submit another PR unless @drew-gross wants to bump the `package.json` up to a later version ;-)

flovilmart:
Let'S stick with 4.3, the use strict has been added :p

Issue #515, opened 2016-08-31
October missing from calendar widget on date columns


alsutton:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [Y] You're running version >=2.1.4 of Parse Server.

- [ Y] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup

F8 Sample app running on a Linux server

#### Steps to reproduce

- Find a date column (e.g. Agenda class, startTime column)
- Select an existing value
- Using the left/right (prev/next month) buttons on the calendar widget try to find October
- October is skipped (Month goes from Sept to Nov, and Nov to Sept).

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

Not a hard error. Issue is with UI.

HorsieSoftware:
I would like to extend this bug report and politely request that it's urgency be raised, because in the global app config (as opposed to the data browser), date fields are _only_ editable via the calendar widget. Thus while you can work around these UI issues in the data browser by manually editing the date string, there is no such opportunity in the app config.

**Calendar Widget Issues:**
1. As noted, the month of October is missing entirely - a rather pressing concern at this time of year.
2. Selecting a particular date results in the day BEFORE that date being entered as the value. So for example, selected September 30th at 23.59 would result in September 29th 23.59 being entered as the value.
3. The number of days in a month is not guaranteed to be correct. E.g. 31 days shown for September, though September has only 30.

Aside from having the widget work correctly, it would be beneficial to have the option to enter the date value manually in the config (like you can in the data browser) and to have the system validate its format.

Issue #387, opened 2016-05-31
Date picker shows wrong day and numbers of days


iStopped:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=2.1.4 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup

ParseDashboard 1.0.11 

#### Steps to reproduce

May has 31 days not 30. Days are wrong!
Just open a date picker, screenshot attached.

<img width="245" alt="screen shot 2016-05-31 at 16 39 02" src="https://cloud.githubusercontent.com/assets/910971/15678418/91b1ead8-274e-11e6-8a59-bec3c4d0fd75.png">

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

drew-gross:
Sorry, this doesn't reproduce for me, I see 31 days. Can yo utry updating to 1.0.12?

steven-supersolid:
This should now be fixed in 1.0.19. The days were being calculated from the previous month in certain timezones.

PR #555, opened 2016-10-17
Fix calendar month/current selection and clicking

steven-supersolid:
Code in DateUtils uses local time (e.g. getMonth() instead of getUTCMonth() etc.) so for compatibility state.currentMonth is also now considered local time.

props.local is used in only 2 places:
1. When constructing state.currentMonth to decide if props.value is in local or UTC time
2. When sending an outgoing onChange event to convert currentMonth back to local or UTC time

There is no code path where props.local is true so this has not been tested. I can't think why this would be required so would like to remove in a non-bugfix PR.

Have tested in timezones UTC+1, UTC-12, UTC+14 (in which case local time was one day after UTC)

### DIFF
--- a/src/components/Calendar/Calendar.react.js
+++ b/src/components/Calendar/Calendar.react.js
@@ -22,14 +22,14 @@ export default class Calendar extends React.Component {
     super();
     let now = props.value || new Date();
     this.state = {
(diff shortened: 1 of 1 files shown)


steven-supersolid:
Fixes #515 and #387

PR #567, opened 2016-10-26
[issue-543] - fix download link

pivanov:


kwstasna:
I can't even remember what `Downloads` was for!
Does it have to be there actually now? 
I think it has to be only to download sample projects.

flovilmart:
Thanks @pivanov !

PR #568, opened 2016-10-29
Add eslint and editorconfig for linting and codestyle

JeremyPlease:
I attempted to base this configuration on the current code style.

`.editorconfig`:
- indent with 2 spaces
- trim trailing whitespace
- `lf` as end of line
- always end file with empty line

Eslint config:
- node, es6, and browser
- React eslint plugin
- properly handles and lints JSX
- allows `console.___` and case declaration.

Let's discuss! Anyone have any feedback, additions, or recommendations?


.eslintrc.json
```diff
@@ -0,0 +1,26 @@
...
+        "sourceType": "module"
+    },
+    "plugins": [
+        "react"
+    ],
+    "rules": {
+        "react/jsx-uses-vars": 1,
+        "react/jsx-uses-react": 1,
+        "react/react-in-jsx-scope": 1,
+        "no-console": 0,
```
flovilmart: do we really need that?

.eslintrc.json
```diff
@@ -0,0 +1,26 @@
...
+        "no-case-declarations": 0
```
flovilmart: that can be fixed by wrapping declarations inside cases within blocks (did it for parse-server)

flovilmart:
@JeremyPlease awesome work! Can we try to reduce to eslint:recommended with the additional plugins?

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
JeremyPlease: > **insert_final_newline**: set to true to ensure file ends with a newline when saving and false to ensure it doesn't.

([editorconfig.org](http://editorconfig.org))

.eslintrc.json
JeremyPlease: I didn't actually fix any linting errors. I noticed the `no-case-declarations` errors a lot so I disabled that rule.

JeremyPlease:
@flovilmart With the current configuration there are 243 linting errors. Should I fix all linting errors in order for us to get the code into good shape and then add to travis `before_script`?

Also can you clarify what you mean by:
> Can we try to reduce to eslint:recommended with the additional plugins?

.editorconfig
JeremyPlease: 😆 How ironic.

flovilmart:
@JeremyPlease that's tedious, but I did it for parse-server, no rush though.

> Also can you clarify what you mean by:
> Can we try to reduce to eslint:recommended with the additional plugins?

Like in parse-server, we tried to stick with [eslint:recommended](https://github.com/ParsePlatform/parse-server/blob/master/.eslintrc.json#L3)

with very few additional rules to reduce the amount of debate :)

.editorconfig
flovilmart: Yeah that made me particularly laugh :)

JeremyPlease:
@flovilmart I can probably do a little marathon of lint error fixing next week.

When I go through and fix errors I'll remove `no-console` and `no-case-declarations`. But, we should keep the `eslint-plugin-react` and 3 current react related rules. Linting react with jsx is sad an inaccurate without these.

flovilmart:
>  Linting react with jsx is sad an inaccurate without these.

Makes total sense, It took me a few hours to fix the linting for parse-server, it was less tedious than expected ;)

JeremyPlease:
I'm sorry that I never came back to this and cleaned up all the linting errors 😞

flovilmart:
Ah that's fine! Let's move ahead with that, lint is not running on the CI, that gives you a chance to fix them till then!

Issue #561, opened 2016-10-20
Cant send push to specific user


gateway:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!
- [x] You're running version >=1.0.18 of Parse Dashboard.
- [ x] You're running version >=2.2.18 of Parse Server.
- [ x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.
#### Environment Setup

Heroku
#### Steps to reproduce
1. Get a user object id and verify they are in the _Installation class
2. Go to push and create a new audience
3. Select iOS (in my case)
4. Select user equals and then the users object id

![image](https://cloud.githubusercontent.com/assets/519563/19570996/1fb8f8fe-96b1-11e6-9f5f-75cb61a6577f.png)
#### Logs/Trace

Basic:
`Oct 20 10:33:51 blah-stage app/web.1:  info: warning: error while sending push code=107, message=You cannot use [object Object] as a query parameter.`

VERBOSE

```
Oct 20 10:40:23 test-stage heroku/router:  at=info method=POST path="/parse/push" host=test-stage.herokuapp.com request_id=475549a9-df33-451a-be0b-8381c99014b4 fwd=“xxx” dyno=web.1 connect=0ms service=95ms status=200 bytes=569 
Oct 20 10:40:24 test-stage app/web.1:  verbose: REQUEST for [POST] /parse/push: { 
Oct 20 10:40:24 test-stage app/web.1:    "where": { 
Oct 20 10:40:24 test-stage app/web.1:      "user": { 
Oct 20 10:40:24 test-stage app/web.1:        "email": “xxx@xxx.com", 
Oct 20 10:40:24 test-stage app/web.1:        "username": "gate1", 
Oct 20 10:40:24 test-stage app/web.1:        "displayName": "gate1", 
Oct 20 10:40:24 test-stage app/web.1:        "_email_verify_token": “xxx”, 
Oct 20 10:40:24 test-stage app/web.1:        "emailVerified": false, 
Oct 20 10:40:24 test-stage app/web.1:        "_email_verify_token_expires_at": "2016-10-20T22:22:21.699Z", 
Oct 20 10:40:24 test-stage app/web.1:        "createdAt": "2016-10-20T17:22:21.416Z", 
[...]

PR #570, opened 2016-10-31
#561: Fixes issue with installation conditions that are pointers

dvanwinkle:
Fixes issue #561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Parse Dashboard Changelog
 
+### NEXT RELEASE
+
+* Fix: Can't send push to specific user (#561)
+
 ### 1.0.19
 
 * New: Support for trusting proxies w/ HTTPS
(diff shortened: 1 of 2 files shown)


steven-supersolid:
lgtm. Not sure if we want to save the changelog update until the actual release? Could get confusing @flovilmart ?

flovilmart:
There are multiple schools there, Updating the changelog as the PR goes is nice too, so the PR releaser doesn't have to do it :) Many OSS projects do it that way, this document the summary of the changes which is nice. I'd be all in to go in that direction for the dashboard. (and obviously put in in the CONTRIBUTING guide).

dvanwinkle:
I'll happily put it in the CONTRIBUTING guide. I'm a big fan of making this as easy as possible for you all so that we can move faster on releases. I'm also more than happy to help with Release management if you need more people.

flovilmart:
now that travis does the releases for us, it's just a matter of tagging :)

flovilmart:
But yeah @dvanwinkle I would definitely love any additional involvement :)

dvanwinkle:
Updated CONTRIBUTING in #574

dvanwinkle:
@flovilmart Anytime, just let me know what you need.

gateway:
@dvanwinkle thanks for this contribution, been waiting for it for a while, simple small little line of code :)

mario:
Folks, can we get a release with this fix anytime soon? Because this is hitting people a lot :P

PR #575, opened 2016-11-02
Updated to pass masterKey for Config get request

### LINKED ISSUE #339: Error loading Config in Dashboard from apps pointing to api.parse.com/1
- [x] Version 2.2.7 of Parse Server
- [x] Same app hosted also on Parse.com
#### Steps to reproduce

Open the Config section in the dashboard for a Parse app, and make sure that a few items are there. Also, make a `GET` on `/config` (make sure that the Master toggle is on) and see the underlying JSON. Now, still in your hosted dashboard, open the same app, but pointing to the server on `api.parse.com/1`. When executing `/config` with master enabled, the results come normally, meaning that the master key is setup correctly. 

I opened the web inspector and saw this:

captura de tela 2016-05-12 as 22 02 39
captura de tela 2016-05-12 as 22 15 47

You can see that Safari and Chrome show the type of the request as `POST`, even though that [`ConfigStore`](https://github.com/ParsePlatform/parse-dashboard/blob/57ca0e003364e16116a92bb4b6ee57a25fecf457/src/lib/stores/ConfigStore.js#L25) requests a get from the JS SDK. 

Could that be an issue on the Parse JS SDK?

brndmg:
I believe this will resolve #339.  When the masterKey is not being passed on the FETCH, then the server is returning a 401.

I had the same issue with version 1.0.19 of the dashboard and 2.2.24 of the server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### NEXT RELEASE
 
 * Fix: Can't send push to specific user (#561)
+* Fix: Config FETCH results in 401 (#339), thanks to [Matt Simms](https://github.com/brndmg)
 
 ### 1.0.19
 
--- a/src/lib/stores/ConfigStore.js
+++ b/src/lib/stores/ConfigStore.js
@@ -22,8 +22,13 @@ function ConfigStore(state, action) {
   action.app.setParseKeys();
   switch (action.type) {
     case ActionTypes.FETCH:
-      return Parse.Config.get().then(({ attributes }) => {
-        return Map({ lastFetch: new Date(), params: Map(attributes) });
+      return Parse._request(
+        'GET',
+        'config',
+        {},
+        { useMasterKey: true }
+      ).then((result) => {
+        return Map({ lastFetch: new Date(), params: Map(result.params) });
       });
     case ActionTypes.SET:
       return Parse._request(


brndmg:
I also have this issue with the Jobs store, even though it isn't functional in this edition, the masterKey still needs to be passed to avoid the 401.
https://github.com/ParsePlatform/parse-dashboard/blob/master/src/lib/stores/JobsStore.js#L33

dvanwinkle:
I'm not sure why you would need to pass in the Master Key here. Are you not using your Master Key in the config? It should pass the Master Key every time.

flovilmart:
Config should be readable without the masterKey also!

brndmg:
@flovilmart and @dvanwinkle You just contradicted each other. 
Maybe parse-dashboard is highlighting an issue with parse-server. I will see if I can get it back to a bad state outside of parse-dashboard.

dvanwinkle:
Sorry, I was referring to parse.com, not `parse-server`. The original ticket that was referenced was parse.com related, which DOES need **A** key. It cannot be called without any key. I just realized that you were fixing something related to `parse-server`, in which yes, it should be able to be read anonymously.

brndmg:
I think is is because I added a clientKey to my parse-server to be able to use it from the Android SDK.  
Before I added the clientKey, I could read from the /config without any extra headers (just the appId).
Once I added the clientKey and restarted the parse-server, parse-dashboard could no longer read the config without the masterKey (or the clientKey).

dvanwinkle:
You can use the Client SDKs without adding the `clientKey` to the Parse server.

With that said, would a better solution be to accept the client keys from the config and then register those with the Node Parse SDK?

brndmg:
It's hard to say.  I know it is possible to avoid using the clientKey, but shouldn't the dashboard work regardless?

dvanwinkle:
I don't know the *right* answer. Thinking of `parse.com`, you are required to have either a `masterKey` or one of the SDK keys to have access to the config. The parse server acts identically, except when you don't have and SDK key, you aren't required to pass in any Client keys.

There isn't any harm in passing the `masterKey` here, so I think that it is okay to merge this in. I am just trying to think of the easiest way to solve all the places that this is happening. You said that the jobs store still had the issue. I am sure that there are a few more places with the same issue.

I am leaning more towards fixing all of the places this issue exists as being the *right* way, so, if @flovilmart agrees, I don't see any issues with merging this in, but, we should probably create and Issue regarding at least the jobs store.

j-koenig:
Any updates on this issue? I'm facing the same problem with `parse-server` v2.2.25 and `parse-dashboard` v1.0.19. We're also using a `clientKey` in our `parse-server` configuration and simply removing it won't be an appropriate solution.

redenz:
Is the only conflict to this issue changelog? Can we get this merged in please?

flovilmart:
We can't get it merged as there is a conflict.

natanrolnik:
@flovilmart now we can fix this PR's conflict here on the web and merge it. As the file is only the Changelog, it's pretty easy to do it.

NatanRolnik:
@codegefluester can we fix the conflict and merge this? Now that the Parse.com dashboard is gone, if Parse-Server has any client keys configured, the dashboard doesn’t access config data. Even though config is public, if the request from dashboard doesn’t have any key, it will fail because if the server requires at least one key.

PR #609, opened 2016-12-04
Check usernameOnly parameter in passwordMatches

lsohn:
When authenticating to dashboard using encrypted passwords, bcrypt throws an error in the authenticate function that userToTest.pass is undefined. This happens when the userToTest object does not contain a password, which happens every time deserializeUser is called. The solution is to check the usernameOnly flag when initializing passwordMatches to avoid passing an undefined value to bcrypt.

### DIFF
--- a/Parse-Dashboard/Authentication.js
+++ b/Parse-Dashboard/Authentication.js
@@ -88,7 +88,7 @@ function authenticate(userToTest, usernameOnly) {
     this.validUsers.find(user => {
       let isAuthenticated = false;
       let usernameMatches = userToTest.name == user.user;
-      let passwordMatches = this.useEncryptedPasswords ? bcrypt.compareSync(userToTest.pass, user.pass) : userToTest.pass == user.pass;
+      let passwordMatches = this.useEncryptedPasswords && !usernameOnly ? bcrypt.compareSync(userToTest.pass, user.pass) : userToTest.pass == user.pass;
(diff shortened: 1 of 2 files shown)


70mmy:
Hi,

I just encountered this bug and got to the same conclusion, the passwordMatches check should not be done if usernameOnly is true

The solution in this pull request fixes the bug, but it actually forces to check if userToTest.pass is equal to user.pass which is still wrong since userToTest.pass is still undefined.

I would replace that line with:

`
let passwordMatches = false;
      if (!usernameOnly) {
        passwordMatches = this.useEncryptedPasswords ? bcrypt.compareSync(userToTest.pass, user.pass) : userToTest.pass == user.pass;
      }
`

lsohn:
Yeah either way is fine, I was just going for less is more. The statement user.pass == undefined evaluates to a boolean value, not undefined. If usernameOnly is true, then it doesn't matter whether passwordMatches is true or false since the (usernameOnly || passwordMatches) statement on the following line will always hold true.

70mmy:
You are right, both have the same result, I just thought my solution is easier to understand :D

flovilmart:
@JeremyPlease, feel free to merge!

JeremyPlease:
@flovilmart I don't have merge access on this repo 😞

flovilmart:
And you can review :/ whatever GitHub :)

PR #652, opened 2017-02-03
Data browser updates object count when table is filtered

### LINKED ISSUE #334: Display count when data is filtered in data browser

rizzomichaelg:
Fixes #334

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -51,6 +51,7 @@ export default class Browser extends DashboardView {
 
       relation: null,
       counts: {},
+      filteredCounts: {},
       clp: {},
       filters: new List(),
       ordering: '-createdAt',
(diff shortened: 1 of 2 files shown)


NatanRolnik:
@rizzomichaelg this looks good to me, beside the changelog:

1 - Can you please add your name, [like in the other features](https://github.com/ParsePlatform/parse-dashboard/blob/master/CHANGELOG.md)?
2 - Can you please add the number of this PR (652) instead of the issue it fixes (334)?

NatanRolnik:
Looks good to me. @steven-supersolid wanna take a look?

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -320,9 +321,21 @@ export default class Browser extends DashboardView {
     return data;
   }
 
+  async fetchParseDataCount(source, filters) {
+    const query = queryFromFilters(source, filters);
+    const count = await query.count({ useMasterKey: true });
```
steven-supersolid: Potential performance issue - if a filter uses non-indexed fields then count() will perform a colscan. This could also happen in fetchParseData so may not be a concern, however count has no option to early exit by hitting a limit whereas fetchParseData does.

I'm not sure how to solve this, perhaps have to rely on parse-server having databaseOptions.maxTimeMS specified?

rizzomichaelg: You raise a great point.  

Besides relying on the parse-server timeout, we could also add a restriction on when the count will update and only update if the total number of objects in the table is less than a certain constant.  Not super excited about this option, though.  Might be best to rely on the server's timeout.

steven-supersolid: I think - let's merge it for now and look at optimizations in a separate PR, unless you want to try anything different?

NatanRolnik: @steven-supersolid I want to release a new version including the Jobs fix, so I'm ok merging this and improving later.

PR #654, opened 2017-02-06
AppsIndex: apps are sorted by name now.

scthi:
The apps list in `src/dashboard/Apps/AppsIndex.react.js` is sorted by `name` now (asc).

It was sorted by the `createdAt` attribute.

### DIFF
--- a/src/dashboard/Apps/AppsIndex.react.js
+++ b/src/dashboard/Apps/AppsIndex.react.js
@@ -134,7 +134,6 @@ export default class AppsIndex extends React.Component {
         </div>
       );
     }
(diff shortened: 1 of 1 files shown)


NatanRolnik:
@scthi thanks for your contribution. [I've noticed that `AppsManager` already returns `apps()` sorted by name.](https://github.com/ParsePlatform/parse-dashboard/blob/master/src/lib/AppsManager.js#L22). In your dashboard, your apps are not sorted?

scthi:
@NatanRolnik Thanks for the hint. That makes my PR even simpler... :-)

NatanRolnik:
Looks good to me. Maybe @steven-supersolid wants to take a look also?

scthi:
Thx guys.

PR #658, opened 2017-02-10
Remove unsupported features in Parse Server

NatanRolnik:
The initial motivation for this are that a few request generate 403 and 404. They are:
* http://serverurl/parse/releases/latest
* http://serverurl/parse//apps/appname/cloud_code/jobs/data

Also removed: in ParseApp.js, all code related to database migration/updating connection string, releases deploy.

CloudCode.react.js, that was responsible for showing Cloud Code files, were removed.

JobsFrom, JobsData, JobEdit were removed, as Parse Server doesn’t support them.


NatanRolnik:
@dvanwinkle, @flovilmart, and @JeremyPlease - I would love your feedback in this PR.

steven-supersolid:
I think it makes sense to remove migration code and disable features that are not supported with a comment or flag, but would recommend keeping the underlying code to minimise work required to get working again, in case someone wants to pick up the task - unless there is no intention of ever supporting?

For cloud code I think there was some discussion and it's not easy to upload new cloud code, but might it still be useful to see current cloud code?

For jobs, I think some work was started on this?

Not sure about what the release section of logs is supposed to denote.

flovilmart:
Reading status, starting a job should work as is. Only scheduling is not supported, and we could very well support it though an adapter if one want to use an existing scheduling library

JeremyPlease:
It's always nice to clean up and remove unused code. I agree that anything that _might_ eventually end up being implemented should be kept and commented. When I implemented form based auth I had to do a commit search to pull back out relevant code/files that already had most of the work done.

@NatanRolnik perhaps you could update this PR to keep all/most of the code with appropriate comments to say it's disabled or not supported?

flovilmart:
The cloud code files read is quite 'trivial' to do, it's just a fs search to run :)

NatanRolnik:
Thank you all for your comments. I'll probably open a new PR or update this taking into account what you said.

NatanRolnik:
Ok, I've opened #660 instead! Thanks guys!

PR #659, opened 2017-02-11
Key concepts for Push Schedule, Master App and App Store Reviews

felipemobile:
This is an attempt to introduce the concept of push schedule and one
master app to handle multiple applications.

What's in there?

- Push schedule concept
- Master app concept to dynamically handle multiple apps on dashboard
- Reviews side bar

Push Schedule

By default all scheduled messages will be saved in the _PushStatus
class with a new "scheduled" status.
The consumer will look for scheduled messages and cancel them direct
from the Dashboard.
You will need a new adapter or a cron to dispatch all scheduled
messages.

Reviews

This is one new module derived from the master app concept.
You will need to execute a python script crawler to download all
reviews from the App Store.

Example: parse-dashboard-config.json

{
"apps": [{
"appName": "Parse Dashboard Master App",
"masterApp": true,
...
}, {
"appName": "Your app name",
"appStoreId": "866492605",
}]
}


.gitignore
```diff
@@ -9,4 +9,10 @@ npm-debug.log
 // vim .swp
 *.swp
 .env
-.idea/
\ No newline at end of file
+<<<<<<< HEAD
```
dvanwinkle: Merge conflict

src/dashboard/Push/PushDetails.react.js
```diff
@@ -45,6 +46,15 @@ let getMessage = (payload) => {
   return '';
 }
 
+let getScheduleStatus = (status) => {
```
dvanwinkle: This entire code block could be simplified to:

```js
return status == "scheduled";
```

flovilmart:
Why don't we get the reviews from the AppStore on parse-server?

flovilmart:
@felipemobile there are too many things at once in that PR. I can't really wrap my head around it.

You seem to have added 3 features that are independant, can you open one Pr per feature please?

src/dashboard/Push/PushDetails.react.js
felipemobile: Hahaha... Shame of me ;-)

felipemobile:
@flovilmart @dvanwinkle First of all, sorry for this PR. I'm 99% sure that this PR was not the appropriated choice. I've been thinking in these concepts, doing some minor prototypes but in order to notify you guys about the changes I decided to share what I had in my mind.

**Scheduled Push**
Why don't we use the _PushStatus to save the push data? By default all scheduled messages can be saved in the _PushStatus class with a new "scheduled" status. We can add a new adapter to ParseServer or at least a cronjob to dispatch the messages. I've prototyped a python script where I list, check, delete before sending the scheduled messages.

**Reviews and Master App**
My intention here is to create, delete, update and manage all login and apps by using one unique master app. For the apps we can use a dynamic NGINX config file and reload it after changes.

I will keep this discussion logically distinct in a new PR but I've created a [python script](https://gist.github.com/felipemobile/f5f4133d49961e0762d39041324104b0) where I get all reviews from the App Store in an unique Parse App. 

screen shot 2017-02-11 at 1 26 20 pm
screen shot 2017-02-11 at 1_fotor
screen shot 2017-02-11 at 1 26 13 pm

flovilmart:
For your master app concept, also, this is very specific to your use case, I'm not sure it belongs here. I could belong to parse-server but again, this is deployment specific, and I'm not sure it belongs here as this won't work for 100% of the use cases and deployments.

felipemobile:
@flovilmart Regarding the master app, I mean a 1-click setting up tool for the parse-server/dashboard instead of having to write JSON files.

I was thinking about the Push schedule, so better to write a python script I was thinking about to write a new push adapter with a pooling mechanism to dispatch schedule messages. What do you think?

PR #689, opened 2017-04-10
Changed links that point to parse.com to parseplatform.org

jcguarinpenaranda:


jcguarinpenaranda:
I am sorry, didn't notice the links were updated to parseplatform.org//docs.. instead of parseplatform.org/docs. 

I will create a new one

PR #693, opened 2017-04-12
Push schedule

felipemobile:
This is an attempt to introduce the concept of push schedule in Parse
Dashboard.

What's in there?

- Push schedule and cancelation

By default all scheduled messages will be saved in the _PushStatus
collection with a "scheduled" status. You will be able to cancel them
directly from the Dashboard.

NOTICE: You will need a new adapter or a cron to dispatch all scheduled
messages stored in _PushStatus collection.

### DIFF
--- a/src/dashboard/Push/PushDetails.react.js
+++ b/src/dashboard/Push/PushDetails.react.js
@@ -30,6 +30,7 @@ import subscribeTo            from 'lib/subscribeTo';
 import tableStyles            from 'components/Table/Table.scss';
 import Toggle                 from 'components/Toggle/Toggle.react';
 import Toolbar                from 'components/Toolbar/Toolbar.react';
+import Button                 from 'components/Button/Button.react';
 import { Directions }         from 'lib/Constants';
 import { Link }               from 'react-router';
 import { Promise }            from 'parse';
@@ -45,12 +46,46 @@ let getMessage = (payload) => {
   return '';
 }
 
+let getScheduleStatus = (status) => {
+  return status == "scheduled";
+}
+
 let getFormattedTime = ({ time, is_local }) => {
   let formattedTime = DateUtils.yearMonthDayTimeFormatter(new Date(time), !is_local);
   if(is_local){
     formattedTime += ' Local Time'
   }
   return formattedTime;
+}
+
+let getScheduleInfo = (sendTime, expiration) => {
+  //expiration unit is in seconds :(
+  if (!sendTime){
+    return '';
+  }
+
+  let fmtSendTime = getFormattedTime({time: sendTime});
+  let fmtExpiration = expiration ? getFormattedTime({time: expiration * 1000}) : null;
+  if (expiration){
+    return `Sent ${fmtSendTime} and expires ${fmtExpiration}`;
+  } else {
+    return `Sent ${fmtSendTime}`;
+  }
+}
+
+let getSentScheduleInfo = (sendTime, expiration) => {
+  //expiration unit is in seconds :(
+  if (!sendTime){
+    return '';
+  }
+
+  let fmtSendTime = getFormattedTime({time: sendTime});
+  let fmtExpiration = expiration ? getFormattedTime({time: expiration * 1000}) : null;
+  if (expiration){
+    return `It will be sent ${fmtSendTime} and expires ${fmtExpiration}`;
+  } else {
+    return `It will be sent ${fmtSendTime}`;
+  }
 }
 
 let getSentInfo = (sendTime, expiration) => {
@@ -512,27 +547,53 @@ export default class PushDetails extends DashboardView {
         </div>
       );
     } else {
-      res = (
-        <div>
-          <div className={styles.groupHeader}>
-            <div className={styles.headerTitle}>MESSAGE SENT</div>
-            <div className={styles.headline}>{getMessage(pushDetails.get('payload'))}</div>
-            <div className={styles.subline}>
-              {getSentInfo(pushDetails.get('pushTime'), pushDetails.get('expiration'))}
+      if(getScheduleStatus(pushDetails.get('status'))) {
+        res = (<div className={styles.groupHeader}>
+                  <div className={styles.headerTitle}>MESSAGE SCHEDULED</div>
+                  <div className={styles.headline}>{getMessage(pushDetails.get('payload'))}</div>
+                  <div className={styles.subline}>
+                      {getSentScheduleInfo(pushDetails.get('pushTime'), pushDetails.get('expiration'))}
+                  </div>
+                  <div className={styles.headline}>
+                    <Button
+                      value='Cancel schedule'
+                      primary={true}
+                      onClick={() => {
+                          let promise = this.context.currentApp.cancelPushSchedule(this.state.pushDetails.id);
+                          promise.then((push) => {
+                            push.destroy({ useMasterKey: true }).then(() => {
+                              history.push(this.context.generatePath('push/activity'));
+                            });
+                          });
+                      }}>
+                    </Button>
+                  </div>
+              </div>
+          );
+      } else {
+        res = (
+          <div>
+            <div className={styles.groupHeader}>
+                <div className={styles.headerTitle}>MESSAGE SENT</div>
+                <div className={styles.headline}>{getMessage(pushDetails.get('payload'))}</div>
+                <div className={styles.subline}>
+                    {getSentInfo(pushDetails.get('pushTime'), pushDetails.get('expiration'))}
+                </div>
             </div>
+            {prevLaunchGroup}
+            {experimentInfo}
+            <PushOpenRate
+              numOpened={pushDetails.get('numOpened') || 0}
+              numSent={pushDetails.get('numSent')}
+              customColor={this.state.standardColor} />
           </div>
-          {prevLaunchGroup}
-          {experimentInfo}
-          <PushOpenRate
-            numOpened={pushDetails.get('numOpened') || 0}
-            numSent={pushDetails.get('numSent')}
-            customColor={this.state.standardColor} />
-        </div>
-      );
+        );
+      }
     }
 
     return res;
   }
+
 
   renderAnalytics() {
     if (Object.keys(this.state.chartData).length > 0) {
--- a/src/dashboard/Push/PushNew.react.js
+++ b/src/dashboard/Push/PushNew.react.js
@@ -39,7 +39,12 @@ import { Directions }          from 'lib/Constants';
 import { Promise }             from 'parse';
 
 const PARSE_SERVER_SUPPORTS_AB_TESTING = false;
-const PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH = false;
+
+const PARSE_SERVER_SUPPORTS_PUSH_RICH_MEDIA = false;
+
+const PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH = true;
+const PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH_EXPIRATION = false;
+const PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH_TIMEZONE = false;
 
 let formatErrorMessage = (emptyInputMessages, key) => {
   let boldMessages = emptyInputMessages.map((message) => {
@@ -59,6 +64,13 @@ let isValidJSON = (input) => {
   } else {
     return false;
   }
+}
+
+let isValidURL = (input) => {
+  var urlPattern = "(https?|ftp)://(www\\.)?(((([a-zA-Z0-9.-]+\\.){1,}[a-zA-Z]{2,4}|localhost))|((\\d{1,3}\\.){3}(\\d{1,3})))(:(\\d+))?(/([a-zA-Z0-9-._~!$&'()*+,;=:@/]|%[0-9A-F]{2})*)?(\\?([a-zA-Z0-9-._~!$&'()*+,;=:/?@]|%[0-9A-F]{2})*)?(#([a-zA-Z0-9._-]|%[0-9A-F]{2})*)?";
+  urlPattern = "^" + urlPattern + "$";
+  var regex = new RegExp(urlPattern);
+  return regex.test(input);
 }
 
 let LocalizedMessageField = ({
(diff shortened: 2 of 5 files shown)


felipemobile:
push-schedule
push-schedule-cancel
push-schedule-new

src/dashboard/Push/PushNew.react.js
```diff
@@ -670,21 +714,40 @@ export default class PushNew extends DashboardView {
...
+      {PARSE_SERVER_SUPPORTS_PUSH_RICH_MEDIA ? this.renderRichMediaContent(fields, setField) : null}
+    </Fieldset> : null;
+
     const timeFieldsLegend = PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH ?
       'Choose a delivery time' :
       'Choose exiry';
 
     const timeFieldsDescription = PARSE_SERVER_SUPPORTS_SCHEDULE_PUSH ?
-      'We can send the campaign immediately, or any time in the next 2 weeks.' :
+      'We can send the campaign immediately, or any time in the next weeks.' :
```
dvanwinkle: Should this be changed to `We can send the campaign immediately, or any time in the future.`?

src/dashboard/Push/PushNew.react.js
```diff
@@ -201,27 +213,39 @@ export default class PushNew extends DashboardView {
...
-      //navigate to push index page and clear cache once push store is created
-      if (error) {
-        promise.reject({ error });
-      } else {
-        //TODO: global success message banner for passing successful creation - store should also be cleared
-        const PARSE_SERVER_SUPPORTS_PUSH_INDEX = false;
-        if (PARSE_SERVER_SUPPORTS_PUSH_INDEX) {
-          history.push(this.context.generatePath('push/activity'));
+    
+    if(changes.push_time_type === 'now') {
```
dvanwinkle: Not to be nit-picky, and obviously not high priority, can you add a space between `if` and `(` everywhere that you've omitted it? For consistency.

src/dashboard/Push/PushNew.react.js
```diff
@@ -670,21 +714,40 @@ export default class PushNew extends DashboardView {
       {this.renderExperimentContent(fields, setField)}
     </Fieldset> : null;
 
+    const richmediaFieldsLegend = PARSE_SERVER_SUPPORTS_PUSH_RICH_MEDIA ?
+      'Rich media push notification' :
+      'Rich notifications';
+
+    const richmediaFieldsDescription = PARSE_SERVER_SUPPORTS_PUSH_RICH_MEDIA ?
+      'We can send images and videos directly to your app.' :
+      "If your push hasn't been send with rich media, it won't get setup by your dev.";
```
dvanwinkle: Using double quotes here but single everywhere else.

Also, should probably be something like: `Rich media is not currently supported.`

dvanwinkle:
This could obviously be a future improvement, but, for situations where the cron job goes down, it'd be nice to have a min/max time. We send pushes to ~1 million devices, and sometimes have windows of time that we're able to send them.

This may be for @flovilmart, but, I'm guessing this wouldn't work with push adapters like AWS SQS?

flovilmart:
That feature is a huge help, on the front end side, you should probably gate it's access with the feature availability API we use somewhere else to make sure no one reports scheduled push as non working because the feature is not enabled / configured on the server.

felipemobile:
Scheduled push can work without any server setup. A background job could be used in the parse-server-example to enable the listing of scheduled push (_PushStatus collection) and delivery at the right time. I know that front end is just the superficial evidence of a much larger problem but we need to go further in this matter.

flovilmart:
@felipemobile and that's the main issue, I prefer having the server having a flag/configuration that says scheduled push is ON, instead of having complaints that push are not sent when scheduled!

felipemobile:
@flovilmart Sorry for the dumb question but why a flag configuration on parse-server since the processing will be done in another module.

dvanwinkle:
@felipemobile I'd assume because Parse server will not respect the schedule quite yet and will try to send immediately? (This is a real assumption, I'm not 100% sure if it's accurate)

felipemobile:
My intention with this pull request is simple: get the push stored in _PushStatus collection flagged as 'scheduled'. The mechanism to delivery the message can be implemented on a push-adapter for example (yes, the row needs to be erased from _PushStatus before/after sending the request to dispatch the message to parse-server).

flovilmart:
And this is why, it should still go through the push route and ignored by the server. parse-server should effectively store the push status with the schedule in, and ingore the sending. The _PushStatus collection should not be manipulated directly by the clients, even if it comes from the dashboard.

flovilmart:
Also, the configuration flags on the server is how we handle the availability of a feature or another on the dashboard. There are plenty of examples of them being used and processed.

We do this to provide a consistant and nice degraded experience, and it gives a chance to show a guide or a link to a guide in the dashboard.

Again, if tomorrow, the next release of the dashboard has this schedule, 100% of the users will fail sending scheduled pushes and we'll have to deal with those non-issues. 

I don't believe it's a good practice to expose a broken feature.

felipemobile:
I completely agree. Sorry for any inconvenience .. I'm just freaking out with the lack of this feature.

dvanwinkle:
@felipemobile You could add the scheduling feature to the Parse server. I think you're on the right track with it.

flovilmart:
Not even thinking that we can run scheduling on parse-server which imply having timers etc on all instances etc... But, at least, add the option to say that this server supports scheduling.

felipemobile:
@dvanwinkle If I follow this path I should start implementing the scheduling feature in each SDK because there's no row to support scheduling in the Parse.Push.send({ }). Am I right?

@flovilmart If we can't run scheduling on parse-server what's the implication of the parse-server saying that scheduling is supported? It will be never supported by the parse-server.

felipemobile:
@flovilmart ok ok. Let's start again... :)

felipemobile:
@flovilmart Please let me know if you have any special request before merging into master.

src/dashboard/Push/PushNew.react.js
```diff
@@ -158,7 +167,7 @@ export default class PushNew extends DashboardView {
       options).then(() => {
       this.setState({ pushAudiencesFetched :true });
     });
-
+
```
flovilmart: nit, remove white space

src/lib/ParseApp.js
```diff
@@ -403,6 +403,12 @@ export default class ParseApp {
     return query.first({ useMasterKey: true });
   }
 
+  cancelPushSchedule(objectId) {
+    var query = new Parse.Query("_PushStatus");
+    query.equalTo('objectId', objectId);
+    return query.first({ useMasterKey: true });
```
flovilmart: this just fetches the object, shouldn't we destroy it?

src/dashboard/Push/PushDetails.react.js
```diff
@@ -53,6 +58,36 @@ let getFormattedTime = ({ time, is_local }) => {
   return formattedTime;
 }
 
+let getScheduleInfo = (sendTime, expiration) => {
```
flovilmart: const

src/dashboard/Push/PushDetails.react.js
```diff
@@ -53,6 +58,36 @@ let getFormattedTime = ({ time, is_local }) => {
...
+  //expiration unit is in seconds :(
+  if (!sendTime){
+    return '';
+  }
+
+  let fmtSendTime = getFormattedTime({time: sendTime});
```
flovilmart: const

src/dashboard/Push/PushDetails.react.js
```diff
@@ -53,6 +58,36 @@ let getFormattedTime = ({ time, is_local }) => {
...
+  let fmtExpiration = expiration ? getFormattedTime({time: expiration * 1000}) : null;
```
flovilmart: const

src/dashboard/Push/PushDetails.react.js
```diff
@@ -53,6 +58,36 @@ let getFormattedTime = ({ time, is_local }) => {
...
+  }
+}
+
+let getSentScheduleInfo = (sendTime, expiration) => {
+  //expiration unit is in seconds :(
+  if (!sendTime){
+    return '';
+  }
+
+  let fmtSendTime = getFormattedTime({time: sendTime});
```
flovilmart: const

src/dashboard/Push/PushDetails.react.js
```diff
@@ -736,4 +797,4 @@ export default class PushDetails extends DashboardView {
       </div>
     );
   }
-}
+}
```
flovilmart: add blank line

src/dashboard/Push/PushNew.react.js
```diff
@@ -881,4 +941,4 @@ export default class PushNew extends DashboardView {
         return <div>{deviceNote}{timeNote}</div>;
       }}/>;
   }
-}
+}
```
flovilmart: blank line

src/dashboard/Push/PushDetails.react.js
felipemobile: omg! :)

src/lib/ParseApp.js
felipemobile: Yes, sure.

src/dashboard/Push/PushDetails.react.js
felipemobile: @flovilmart Did I miss anything?

flovilmart: I'll have a deeper look tomorrow, thanks for hanging in there .)

src/lib/ParseApp.js
```diff
@@ -403,6 +403,12 @@ export default class ParseApp {
     return query.first({ useMasterKey: true });
   }
 
+  cancelPushSchedule(objectId) {
+    var query = new Parse.Query("_PushStatus");
+    query.equalTo('objectId', objectId);
+    return query.first({ useMasterKey: true }).destroy({ useMasterKey: true });
```
flovilmart: not sure that works :/

should be:
```
const pushStatus = new Parse.Object('_PushStatus');
pushStatus.id = objectId;
return pushStatus.destroy({ useMasterKey: true });
```

felipemobile: fixed

felipemobile:
@flovilmart What can I do to include this PR in the next 1.0.26 version?

felipemobile:
@flovilmart Sorry, this PR was not working as it should. 

I've applied a fix as you may seen above. 

PS: You can't send push messages immediately with parse-server 2.4.0, it is returning 504 GATEWAY_TIMEOUT.

flovilmart:
I'll have a look for the issue of 2.4.0, did you use an audience or something?

felipemobile:
Even using an audience the push got stuck in this state.

felipemobile:
Eager to see this feature enabled in parse-dashboard :-)

kamioon:
Perfect feature, is there any plan for releasing new version ?

flovilmart:
@kamioon the scheduling won't actually happen, you'll need a cron job to run the push sending on time.

flovilmart:
@dvanwinkle can you give the final look please, (code-wise) and merge when you see fit!

felipemobile:
@flovilmart Push scheduling is already working without pooling _PushStatus!!! Please, show me the magic! :-) 🥇

flovilmart:
it doesn't make sense

felipemobile:
Yes, that's exactly what I thought!  It doesn't make sense :-)

flovilmart:
What do you mean by it's working? they are run at the right time? that's odd...

felipemobile:
If the pushTime is set to an date before now it is sent immediately.
If you set the pushTime ahead they run at the right time.

felipemobile:
This behavior started after I ran the following cloud function (maybe it is still running in background).

````
curl -X POST \
  -H "X-Parse-Application-Id: {$app_id}" \
  -H "X-Parse-REST-API-Key: {$rest_id}" \
  -H "Content-Type: application/json" \
  -d '{}' \
  http://localhost:1363/1/functions/checkPushSchedule
````

```
Parse.Cloud.define("checkPushSchedule", function(request, response) {
    // Get all message from the past year until now
    var now = new Date();
    var pastYear = new Date();
    pastYear.setDate(pastYear.getDate() - 365);
    // Query _PushStatus
    var query = new Parse.Query("_PushStatus");
    query.equalTo("status", "scheduled");
    query.lessThanOrEqualTo('pushTime', now);
    query.greaterThan('pushTime', pastYear);
    query.find({ useMasterKey: true }).then(results => {
        if (results.length > 0) {
            for (var i = 0; i < results.length; i++) {
                var payload = results[i].get("payload");
                var where = results[i].get("query");
                Parse.Push.send({
                    where: JSON.parse(where),
                    data: JSON.parse(payload)
                }, { useMasterKey: true}).then(() => {
                    // Push was successful
                }, (e) => {
                    // Push was not sent
                    console.log(e);
                });
            }
            response.success("Your push was sent to " + results.length + " devices.");
        } else {
            response.success("You have no messages to be delivered at this time.");
        }
    });
});
```

flovilmart:
Ah so it's not by itself right !

natanrolnik:
This line in th code you posted seems to me wrong:

`response.success("Your push was sent to " + results.length + " devices.");`

Results is an array of scheduled _PushStatus, not devices.

felipemobile:
No it's by itself since I've executed this cloud function one unique time !

@natanrolnik thanks!

felipemobile:
I've restarted Parse-server 3.4.1 and I have to confirm that push scheduling is working and delivering messages at the right time. That's disturbing.

flovilmart:
It doesn't make any sense... Did you migrate that app? @drew-gross @hramos do you still have workers running at parse.com on the migrated databases????

felipemobile:
Surely You're joking, @flovilmart. It's working for two non migrated apps and my security group on port 27017 is tight :-)

flovilmart:
if you schedule an hour ahead, will it be sent? Or is it just soon enough?

felipemobile:
Worked for an hour ahead. I'll check if I've developed some worker or crontab while I was asleep :-)

flovilmart:
This doesn't make sense at all... did you sent any other push after?

felipemobile:
All day long... 10min/1hour/etc

flovilmart:
Uhm... perhaps the worker is pulling your scheduled pushes??

felipemobile:
Overnight pushes didn't work. I will check three hours ahead.

felipemobile:
After opening the dashboard my push set to 06:00 AM (GMT -0300) was immediately sent. Is there an api endpoint that I can ping to get my scheduled messages sent?

felipemobile:
@flovilmart I'm sorry about this mess! I feel embarrassed and ashamed by the way I acted.
Finally I've found the cron set in one of my servers. In conclusion I really had developed this when I was asleep (the server stopped after midnight due to the lack of usage). That's the only explanation for that :-/ 👎

u3amar:
What's the status of this? Would be great to have this feature.

PR #701, opened 2017-04-26
_PushStatus add to SpecialClasses

gimdongwoo:
I love parse-server & parse-dashboard.

I have found that the _PushStatus class is specially handled, but that's the definition is missing.

It's a simple changes, but I'm glad to be able to send PR to parse-dashboard.

Thank you!

### DIFF
--- a/src/lib/Constants.js
+++ b/src/lib/Constants.js
@@ -51,6 +51,7 @@ export const SpecialClasses = {
   _Role: 'Role',
   _Product: 'Product',
   _Session: 'Session',
+  _PushStatus: 'PushStatus',
 };
 
 export const DefaultColumns = {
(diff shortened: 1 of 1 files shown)


natanrolnik:
@gimdongwoo thanks for your contribution!
What's the result in the Dashboard for having `_PushStatus` as a special class? It shows by default in the available classes in the menu?

If so, I'm not sure that this is something desired, considering that we can have the **Push** section in the menu as well, that shows the `_PushStatus` in a better way.

gimdongwoo:
Yes, _PushStatus is exposed by default.


Of course, you can also see it on the Push tab.

But this is not searchable, it is difficult to handle push results as data.

It's better way to give it a searchable name, such as _User, _Installation, _Session, etc.

flovilmart:
I'm not sure it's really something we want to be exposed like that, perhaps we could have a 'browser' view into the Past Pushes instead of adding the _PushStatus to the tables. _PushStatus is an internal table that is not managed by the user. This should be read only, even for the masterKey.

gimdongwoo:
It's already exposed basically,
rather than being exposed as like a bug named _PushStatus, I thought it was right to give it a name.

For those exposed as accessible classes, I thought it seems better way to have the PushStatus exposed, like Session or Installation is exposed.

natanrolnik:
I actually think we could have them displayed in the Dashboard, for easier querying, and having all the fields with edit disabled, read only.

@flovilmart what do you think?
@gimdongwoo do you think that would be possible to do?

flovilmart:
Alright let's go! I actually had to query in it recently, so that's a nice addition!

gimdongwoo:
@flovilmart Thank you. It's my honor 👍 .

PR #710, opened 2017-05-22
Enforces lint on test

flovilmart:
@JeremyPlease that's it!


flovilmart:
Was actually faster to do than parse-server !

Issue #713, opened 2017-05-26
Parse file column can not be edited in dashbaord


monajafi:
It's not possible to edit and update existing parsefile coulmn in dashbaord. Just new file can be uploaded .

r1cebank:
You can already do something like this, is this what you are trying to find?
<img width="387" alt="screenshot at may 27 23-20-02" src="https://cloud.githubusercontent.com/assets/261671/26526531/120c2bac-4333-11e7-8b34-8de460ea3fdb.png">

monajafi:
In my dashboard I can't find those delete and Replace file. My dashbaord version is 1.0.25
here is my image file screenshot, when I click it is opening image file:
<img width="149" alt="screen shot 2017-05-29 at 19 18 05" src="https://cloud.githubusercontent.com/assets/6666486/26554134/f5c83568-44a3-11e7-9de2-335069129a60.png">

natanrolnik:
Try double clicking twice in the white part of the cell

monajafi:
Thanks a lot man. you saved my day

natanrolnik:
Glad I could help, @monajafi. I'll try to make a PR to improve this.

natanrolnik:
@monajafi I've did a few changes to improve this behavior:

#716

PR #716, opened 2017-05-30
Moves download option into file editor

natanrolnik:
Instead of clicking in the file cell to download, and double clicking in the white part of the cell to edit/delete, this change adds download into the editor, so it’s easier to edit if needed, while keeping the download functionality.

captura de tela 2017-05-30 as 12 48 36

This addresses the problems raised in #713

### DIFF
--- a/src/components/FileEditor/FileEditor.react.js
+++ b/src/components/FileEditor/FileEditor.react.js
@@ -52,13 +52,15 @@ export default class FileEditor extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


src/components/FileEditor/FileEditor.react.js
```diff
@@ -52,13 +52,15 @@ export default class FileEditor extends React.Component {
   }
 
   render() {
+    let file = this.props.value;
```
acinader: prefer const

Issue #720, opened 2017-06-07
Can’t search within an `Object` type column


oferRounds:
Would be great if we’d be able to search for a specific key/value pair within an `Object` type column (same as we can do with `Array`).

oferRounds:
@natanrolnik

ssamuli:
I have a working prototype for this. Will clean up and submit a PR later so you can have a look if it fits this purpose.

natanrolnik:
Closed in #727

PR #724, opened 2017-06-12
Added/fixed a filtering option "contains string" for String fields.

ssamuli:
There are options "startsWith" and "endsWith" in the filter input for string fields, but "contains string" option is for some reason missing, so I made this change. In my opinion it is pretty important feature for people who use dashboard to browse their parse data.

It does filtering case insensitively for now. Could maybe add later a checkbox for selecting case sensitivity.

screen shot 2017-06-12 at 14 13 34

### DIFF
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -60,6 +60,8 @@ function addConstraint(query, filter) {
       query.greaterThan(filter.get('field'), filter.get('compareTo'));
       break;
     case 'containsString':
+      query.matches(filter.get('field'), filter.get('compareTo'), 'i');
+      break;
     case 'containsNumber':
       query.equalTo(filter.get('field'), filter.get('compareTo'));
       break;
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -60,8 +60,6 @@ function addConstraint(query, filter) {
       query.greaterThan(filter.get('field'), filter.get('compareTo'));
       break;
     case 'containsString':
-      query.matches(filter.get('field'), filter.get('compareTo'), 'i');
-      break;
     case 'containsNumber':
       query.equalTo(filter.get('field'), filter.get('compareTo'));
       break;
@@ -71,6 +69,9 @@ function addConstraint(query, filter) {
       break;
     case 'containedIn':
       query.containedIn(filter.get('field'), filter.get('array'));
+      break;
+    case 'stringContainsString':
+      query.matches(filter.get('field'), filter.get('compareTo'), 'i');
       break;
   }
   return query;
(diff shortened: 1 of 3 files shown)


natanrolnik:
Thanks for your PR, @ssamuli. I left one small note.

src/lib/queryFromFilters.js
```diff
@@ -60,6 +60,8 @@ function addConstraint(query, filter) {
       query.greaterThan(filter.get('field'), filter.get('compareTo'));
       break;
     case 'containsString':
+      query.matches(filter.get('field'), filter.get('compareTo'), 'i');
```
natanrolnik: The documentation in [Parse.query](http://parseplatform.org/Parse-SDK-JS/api/classes/Parse.Query.html#methods_matches) states only two parameters: key and regex. What's the `'i'` string you pass here?

ssamuli: Its an undocumented "modifiers" parameter to "matches" function of ParseQuery, 'i' means case insensitive: [https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L714](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L714l)

In retrospective, looking at the ParseQuery code, seems to be possible to pass 'case insensitive' inside the second parameter too.

ssamuli:
Is there something you'd like me to change or add to this feature to facilitate the merging process?

natanrolnik:
One last question I have before merging, as I was working on something very similar (allowing to filter if a pointer exists or not): before your changes, both `containsString` and `containsNumber` had the same query method on them, but it didn't matter because `containsString` wasn't activated in `src/lib/Filters.js`. Is my understanding correct?

ssamuli:
You are correct, it was not activated in Filters.js. 

Also the query was not correct because it fell through to `case 'containsNumber'`, which does an equalTo query. This works for finding strings and numbers from arrays but not from inside other strings. For strings it would only find ones that are a 100% match.

Maybe to make things more clear and not change how the query behaves with arrays (although this seems to work with them too), I should add a new `Constraint` in Filters.js? `stringContainsString` or similar?

ssamuli:
Should be better and safer like this. Didn't think of the array query much when I initially sent the PR.

natanrolnik:
Also, @ssamuli, I was trying recently to do something to fix #720 .
I played a little bit with it, and it was harder than what I imagined.

Initially, we could add at least an option to have **exists** and **does not exist** for pointers - that's currently not available and works out of the box with the current `query.exists(filter.get('field'));` or `query.doesNotExist(filter.get('field'));`, just by adding `Object` to `FieldConstraints`.

What do you say? Could you help us with this?

ssamuli:
Thanks! Looks like that's something I can help with.

natanrolnik:
Bear in mind that you should use dot notation for this:
`query.equalTo("column.numberProp", 42)`

ssamuli:
@natanrolnik OK, here it is: https://github.com/parse-community/parse-dashboard/pull/727

PR #725, opened 2017-06-12
Sort config data according to parameter names

natanrolnik:


natanrolnik:
Wait, that's actually wrong.. I'll fix it and reopen

natanrolnik:
Reopened as #726

PR #727, opened 2017-06-13
Add ability to search Object columns

ssamuli:
As requested in issue #720 

What do you think, is this something you were looking for?

screen shot 2017-06-13 at 20 19 45
screen shot 2017-06-13 at 20 19 17

### DIFF
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -21,6 +21,13 @@ export default function queryFromFilters(className, filters) {
   return query;
 }
 
+function addQueryConstraintFromObject(query, filter, constraintType) {
+  let compareTo = JSON.parse(filter.get('compareTo'));
+  for (let key of Object.keys(compareTo)) {
+    query[constraintType](filter.get('field')+'.'+key, compareTo[key]);
+  }
+}
+
 function addConstraint(query, filter) {
   switch (filter.get('constraint')) {
     case 'exists':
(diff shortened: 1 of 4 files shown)


natanrolnik:
OMG this is fantastic!!!
This is good, but considering we are dealing with this, do you honk we could have does not contain, greater than/greater than or equal to, less than/less than or equal to? That would be perfect!!
If not, we can merge and improve it later.

oferRounds:
@ssamuli - THANK YOU, just saw it now, looks awesome! That would really be useful!

ssamuli:
@natanrolnik good idea. OK, now this should cover all.

ssamuli:
@oferRounds happy to be of help!

src/lib/queryFromFilters.js
```diff
@@ -21,6 +21,13 @@ export default function queryFromFilters(className, filters) {
   return query;
 }
 
+function addQueryConstraintFromObject(query, filter, constraintType) {
+  let compareTo = JSON.parse(filter.get('compareTo'));
+  for (let key of Object.keys(compareTo)) {
+    query[constraintType](filter.get('field')+'.'+key, compareTo[key]);
```
natanrolnik: Interesting, I didn't know about the `query['equalTo'](..., ...)` syntax...

ssamuli: It's an old javascript feature, I'm just accessing the function by its name through bracket notation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#Bracket_notation

natanrolnik: Good to know! Thank you.

natanrolnik:
@ssamuli thanks once again for your contribution, this is amazing. I'm sorry to be picky here, but I think it's important to get it right before merging - I have two notes:

1 - the _key exists_ and _key does not exist_ are not working for me, or at least I don't understand how I should write the key. Not only that, for these options, it doesn't save the state when I open the Filter popup again. 

I guess the problem is in the following code:

```
let compareTo = JSON.parse(filter.get('compareTo'));
     for (let key of Object.keys(compareTo)) {
```

And I guess you need a function different than:
`function addQueryConstraintFromObject(query, filter, constraintType) {`

2 - Do you think you could add a _exists_ and _does not exist_ option also, as the first ones, for the object itself? I mean, if I want to search for Parse Object where a object column exists or not, in this current implementation it's not possible.

ssamuli:
1. Key exists and does not exist works with {"key.im.looking.for":1} for example, did you try that? Should work, at least it did last night when I tried it.

2. Sure, I'll add those too.

natanrolnik:
I didn't try this way, but I would expect it to work just with a plain string. The example you gave requires understanding how the underlying code works, and that's why I suggested a different function to modify the query that could handle only a string and not a JSON object for **_key exists_** and **_key does not exist_**.

ssamuli:
Yes it's a bit tricky to use but it's kind of consistent with how the rest of the object search works, that's why I thought it was acceptable. I'll change that to work with strings as you suggested.

ssamuli:
There you are 🙂

ssamuli:
Btw. would be great feature if those search options had placeholders showing the user how the input should look like. Not just for this but for everything.

natanrolnik:
Tested now and it's perfect. Let's 🚢 it!
Thanks @ssamuli!

ssamuli:
Thank you! 🎉

natanrolnik:
Closes #720

PR #728, opened 2017-06-14
Restore jobs scheduling interface and capabilities

flovilmart:
This goes side by side with https://github.com/parse-community/parse-server/pull/3927

This restores the `Scheduled Jobs` section, keeps the `All Jobs` to run them manually, modernizes the network calls by using Parse._request instead of AJAX.

### DIFF
--- a/src/dashboard/Data/Jobs/Jobs.react.js
+++ b/src/dashboard/Data/Jobs/Jobs.react.js
@@ -27,7 +27,7 @@ import Toolbar         from 'components/Toolbar/Toolbar.react';
 
 let subsections = {
   all: 'All Jobs',
-  /*scheduled: 'Scheduled Jobs',*/
+  scheduled: 'Scheduled Jobs',
   status: 'Job Status'
 };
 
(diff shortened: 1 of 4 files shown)


flovilmart:
@natanrolnik what do you  think?

natanrolnik:
In general I think it looks good. I just think we need to make it super clear to users that it's the user's responsibility to actually trigger the jobs, differently than what it used to be with Parse.com and how people would expect it to work out of the box.
We should maybe add a message in the same area as the jobs scheduling and add it to the server docs?

flovilmart:
Yes that's a great idea!

natanrolnik:
@flovilmart waiting for your change to merge this one, as parse-community/parse-server#3927 was already merged

flovilmart:
Yeppy yay!

flovilmart:
@natanrolnik actually, I kinda struggle setting that up, me no good with React!\

natanrolnik:
I’ll try to do it in the next few days then.

natanrolnik:
Let's get this in, @ssamuli will help us later on adding the message.

PR #731, opened 2017-06-14
Fix display bug on safari when table has empty cells ('')

ssamuli:
The tables look broken on Safari when there are empty cells:

screen shot 2017-06-14 at 18 06 36

And after this patch:

screen shot 2017-06-14 at 18 07 00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### NEXT RELEASE
 
 * _Contributing to this repo? Add info about your change here to be included in next release_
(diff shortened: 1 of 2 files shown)


ssamuli:
This bug has been bothering me for quite some time. Am I the only one? 🙂

flovilmart:
Nice, I didn't notice :)

natanrolnik:
That bothered me as well!
Thanks again, @ssamuli!
Can you just add the change to the change log?

ssamuli:
@natanrolnik OK, done.

PR #734, opened 2017-06-21
Feature: When editing Object or Array fields the data is displayed in a prettier format and the textarea is resizable

ssamuli:
Makes editing objects a lot more fun.

jsonedit_resizable


natanrolnik:
Awesome. Thanks @ssamuli !

Issue #539, opened 2016-09-21
PushDetails not rendering if the push name is an object.


benjamingrobertson:
- [x] You're running version >=1.0.18 of Parse Dashboard.
- [x] You're running version >=2.2.18 of Parse Server.
- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.
#### Environment Setup

I've got a parse server and a parse dashboard setup at two different heroku apps. I'm sending push notifications to the server from the Parse PHP SDK. The notifications are delivered, I'm just not able to see them in the dashboard if they are an object. Text notifications sent via the dashboard are delivered and I can view their details just fine.
#### Steps to reproduce

Navigate to Push / Past Pushes. Click on a Push Notification that has a name like: `{"title":"Test Push","body":"Push Notification"}`

Details do no load. JS console reports: "Uncaught Invariant Violation: Objects are not valid as a React child (found: object with keys {body, title}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of `PushDetails`"
#### Logs/Trace

Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

<pre>
<code>
Uncaught Invariant Violation: Objects are not valid as a React child (found: object with keys {body, title}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of `PushDetails`.

invariant   @   invariant.js?bb9b:38
traverseAllChildrenImpl @   traverseAllChildren.js?fc13:136
traverseAllChildren @   traverseAllChildren.js?fc13:164
instantiateChildren @   ReactChildReconciler.js?660f:71
_reconcilerInstantiateChildren  @   ReactMultiChild.js?87d2:189
mountChildren   @   ReactMultiChild.js?87d2:228
_createInitialChildren  @   ReactDOMComponent.js?f4f2:698
mountComponent  @   ReactDOMComponent.js?f4f2:523
mountComponent  @   ReactReconciler.js?9da7:47
obj.(anonymous function)    @   backend.js:8386
mountChildren   @   ReactMultiChild.js?87d2:240
_createInitialChildren  @   ReactDOMComponent.js?f4f2:698
mountComponent  @   ReactDOMComponent.js?f4f2:523
mountComponent  @   ReactReconciler.js?9da7:47
[...]

PR #735, opened 2017-06-24
Updated iOS alert object to display title and body detail

repertus:
This fix addresses the issue outlined in #539 by being able to now render the title and body keys from the alert object.

### DIFF
--- a/src/dashboard/Push/PushDetails.react.js
+++ b/src/dashboard/Push/PushDetails.react.js
@@ -39,7 +39,13 @@ const EXP_STATS_URL = 'http://docs.parseplatform.org/ios/guide/#push-experiments
 let getMessage = (payload) => {
   if(payload) {
     let payloadJSON = JSON.parse(payload);
-    return payloadJSON.alert ? payloadJSON.alert : payload;
+		if (payloadJSON.alert.body) {
+			return payloadJSON.alert.body;
+		} else if (payloadJSON.alert) {
+			return payloadJSON.alert;
+		} else {
+			return payload;
+		}
   }
   return '';
 }
--- a/src/dashboard/Push/PushDetails.react.js
+++ b/src/dashboard/Push/PushDetails.react.js
@@ -522,7 +522,7 @@ export default class PushDetails extends DashboardView {
 							{
 								(typeof alert === 'object') ?
 									<div>
-								 		<div className={styles.headline}>{alert.title}</div>
+										<div className={styles.headline}>{alert.title}</div>
 										<div className={styles.headline}>{alert.body}</div>
 									</div>:
 									<div className={styles.headline}>{alert}</div>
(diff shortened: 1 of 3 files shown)


natanrolnik:
Thanks for the PR, @repertus !

Would you mind showing a screenshot of the result?

Also, could you please update the Changelog.md file as well, to contain your improvement?

repertus:
@natanrolnik 

Here are the screenshots of the results and I have updated the Changelog.md to reflect the fix details:

Alert entered as a JSON object:
screen shot 2017-06-28 at 6 22 21 am

Title now reflected for the alert object containing title in the "Past Pushes":
screen shot 2017-06-28 at 6 23 41 am

Message now reflected in the "Pash Pushes" detail:
screen shot 2017-06-28 at 6 24 05 am

repertus:
@natanrolnik,

I will need you help looking at why the build failed in Travis with one of the newer versions of node.js.

src/dashboard/Push/PushDetails.react.js
```diff
@@ -39,7 +39,13 @@ const EXP_STATS_URL = 'http://docs.parseplatform.org/ios/guide/#push-experiments
 let getMessage = (payload) => {
   if(payload) {
     let payloadJSON = JSON.parse(payload);
-    return payloadJSON.alert ? payloadJSON.alert : payload;
+		if (payloadJSON.alert.body) {
```
natanrolnik: There is no need to check the type of `payloadJSON.alert` like in the other changed file? Also, in case there is a body, could you also add the title?

repertus: So under PushDetails if an alert object contains title and body, would you like me to show both the title and message just below the title in smaller font size?

natanrolnik: That would be ideal! Do you think you can do it? I think it makes sense to do it the best way we can if you are already touching these files.

And did you see my question? Don't we need to check the type of `payloadJSON.alert`, or it will always be an object here?

natanrolnik: @repertus Robert, we wanted to make a release in the next few days - if you could do these changes, it will enter in the next release.

pungme:
Hey this feature is neat, @repertus do you have time to wrap it up? or can we take over? How does it work? @natanrolnik 

Cheers

src/dashboard/Push/PushDetails.react.js
repertus: @natanrolnik,

Sorry I took so long to circle back, but I've been tied up. I was able to modify the component to display both the title and the body as part of detail by checking the existance of an tyeof objective to keep consistant as per your observation. Here is a screenshot.
screen shot 2017-08-09 at 8 29 55 am

repertus: @natanrolnik,

Can you outline why the build is failing in Travis.

natanrolnik: @repertus Sorry, I was afk for a few hours. Thanks for working on it!

natanrolnik:
Merged! Thank you so much for your efforts, @repertus!

Issue #557, opened 2016-10-18
Cant filter for negative values


rahilsharma:
The input box for filter option less than does not take negative values. 
Eg : I want to find a column with values less -2. It does not allow taking in negative values.

PR #749, opened 2017-07-13
fix filter for negative and empty number input

miguel-s:
Contributing to #557

Fixes two bugs with number input in filters:
- when input has a negative number (to reproduce in master: typing a "-" char will lead to "NaN" being displayed)
- when input is empty (to reproduce in master: typing a valid number and deleting with backspace will lead to "NaN" being displayed)

### DIFF
--- a/src/components/BrowserFilter/FilterRow.react.js
+++ b/src/components/BrowserFilter/FilterRow.react.js
@@ -49,7 +49,18 @@ function compareValue(info, value, onChangeCompareTo, active) {
     case 'Boolean':
       return <ChromeDropdown color={active ? 'blue' : 'purple'} value={value ? 'True' : 'False'} options={['True', 'False']} onChange={(val) => onChangeCompareTo(val === 'True')} />;
     case 'Number':
-      return <input type='text' value={value} onChange={(e) => onChangeCompareTo(validateNumeric(e.target.value) ? parseFloat(e.target.value) : (parseFloat(value) || ''))} />;
+      return (
+        <input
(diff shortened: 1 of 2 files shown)


src/components/BrowserFilter/FilterRow.react.js
```diff
@@ -49,7 +49,18 @@ function compareValue(info, value, onChangeCompareTo, active) {
...
       return <ChromeDropdown color={active ? 'blue' : 'purple'} value={value ? 'True' : 'False'} options={['True', 'False']} onChange={(val) => onChangeCompareTo(val === 'True')} />;
     case 'Number':
-      return <input type='text' value={value} onChange={(e) => onChangeCompareTo(validateNumeric(e.target.value) ? parseFloat(e.target.value) : (parseFloat(value) || ''))} />;
+      return (
+        <input
+          type='text'
+          value={value}
+          onChange={(e) => {
+            let val = value;
+            if (!e.target.value.length) val = '';
```
natanrolnik: Please don't use `if` and `else` without brackets. This might lead to bugs. And already did in the history of programming many times :) Also, don't keep the `val` assignment in the same line.

miguel-s:
@natanrolnik fixed the formatting and added the line to the Changelog file

flovilmart:
LGTM! Thanks!

pgiguere1:
This seems to have broken number filtering entirely for any 1-digit number.

https://github.com/parse-community/parse-dashboard/issues/831

PR #754, opened 2017-07-23
Converted UTC date to local time

imMatt:
For our development environment we wanted dates listed in the local time instead of UTC, unsure if anyone else would find it useful so submitting here.

It should most likely be configuration, or enabled by a flag but I'm not 100% sure on the best practice for doing that within the project is, so if anyone could give some pointers it'd be much appreciated.

### DIFF
--- a/src/components/BrowserCell/BrowserCell.react.js
+++ b/src/components/BrowserCell/BrowserCell.react.js
@@ -6,6 +6,7 @@
  * the root directory of this source tree.
  */
 import { dateStringUTC }         from 'lib/DateUtils';
+import { dateStringLocale }      from 'lib/DateUtils';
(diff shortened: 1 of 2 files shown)


natanrolnik:
Hey, @Matt-Allen44!
I think that we should be able to pass this from the `config.json` file, or as a parameter when starting with the command line tool.
What do you think?

Matt-Allen44:
I was having a look into doing it as a command line flag, but I wasn't sure what the best practice for retrieving it from within BrowserCell was, as I had a look around and pretty much all of the flags seemed to be handled on startup with the main JS file.

Feel as if parameter and config would be the best way to go.

adolfdsilva:
Hi even we require this feature.. can we merge this?

PR #759, opened 2017-07-28
Add support for Polygon type

madsb:
Dashboard currently breaks when viewing classes that contain a `Polygon` type. This pull request adds support for the new type.


flovilmart:
Thanks @madsb, can you post a small screenshot of the fix?

madsb:
Certainly! Here are some:
polygon_cell
new_polygon_column
edit_polygon

flovilmart:
Sweet! Looking good!

Issue #756, opened 2017-07-25
Parse-dashboard modifies the data of an array containing pointers into a different format


Apptation:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [ ] You're running version >=1.0.23 of Parse Dashboard.

- [ ] You're running version >=2.3.2 of Parse Server.

- [ ] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
- Parse Server v2.5.3 on AWS EB
- NodeJS v4.2.0 (also tested on NodeJS v3.3.1)
- Parse Dashboard v1.1.0
- mongod version: 3.2.14 (MMAPv1) on MLAB
- Safari 10.1.1 on MacOS 10.12.5

#### Steps to reproduce
In cloud code, I'm storing data in the __User table as an array of pointers with a createdDate like this:
`var newsPointer = {"__type":"Pointer","className":"News","objectId":"5hRuMhTxFz"};
var todayNow = new Date();
var userEvent = 
{
	"news": newsPointer,
	"createdAt": todayNow
};
userObject.add("newsEvents", userEvent);`

**1.** This saves the data as expected when viewed in MLAB, or another raw data viewer:
![img_4217](https://user-images.githubusercontent.com/7025214/28554268-91abf016-70c6-11e7-938e-42102be6c054.PNG)

**2.** Then if viewed in Parse-Dashboard, the data appears in this format (no longer a pointer)
![screen shot 2017-07-24 at 10 56 51 pm](https://user-images.githubusercontent.com/7025214/28554312-ca48c48a-70c6-11e7-9e0e-c19885da7ed0.png)

**3.** If you then double click the cell data in "newsEvents", the data will be saved in that incorrect, modified format (screenshot taken with Parser app, after viewing the cell data in Parse-Dashboard):
![img_4218](https://user-images.githubusercontent.com/7025214/28554455-8a4edd0a-70c7-11e7-88fe-14413d021f65.PNG)

I can reproduce the above every time using Parse-Dashboard. To make things worse, intermittently the "news" data is shown as the whole object (showing all fields/data in the "news", where I expect to see the pointer data only).

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

Apptation:
[...]

Issue #590, opened 2016-11-17
Editing an array of dates converts these to an array of Strings


glentregoning:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] You're running version >=1.0.19 of Parse Dashboard.

- [X] You're running version >=2.2.24 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Parse Server 2.2.24
Parse Dashboard 1.0.19

#### Steps to reproduce

1. Add an object that includes an array of Dates to parse-server using something other than the dashboard. E.g. via CURL:
```
curl -X POST -H "X-Parse-Application-Id: <your-app-id>"\
 -H "Content-Type: application/json" \
 -d '{"dates": [{ "__type": "Date", "iso": "2013-11-14T07:37:49.838Z" }] }' http://localhost:1337/parse/classes/TestObject/
```

2. Open up the parse dashboard, go to Browse, select 'TestObject', then double click on the 'dates' row entry.

Observe:

Dashboard makes the following request to the database updating the record with an array of strings instead of dates:
E.g. (taken from verbose parse server logs)

```
verbose: REQUEST for [PUT] /parse/classes/TestObject/5QtnjZmFnC: {
  "dates": [
    "2016-11-14T07:37:49.838Z"
  ]
} method=PUT, url=/parse/classes/TestObject/5QtnjZmFnC, host=localhost:1337, connection=keep-alive, content-length=248, origin=http://localhost:4040, user-agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36, content-type=text/plain, accept=*/*, referer=http://localhost:4040/apps/App/browser/TestObject, accept-encoding=gzip, deflate, br, accept-language=en-US,en;q=0.8, dates=[2016-11-14T07:37:49.838Z]
verbose: RESPONSE from [PUT] /parse/classes/TestObject/5QtnjZmFnC: {
  "response": {
    "updatedAt": "2016-11-17T06:24:42.619Z"
  }
} updatedAt=2016-11-17T06:24:42.619Z
```

Expected:
You should be able to edit an array of dates without it corrupting the structure. 

#### Logs/Trace
[...]

PR #770, opened 2017-09-06
Updating array of Dates no longer converts it to array of strings

rihadavid:
This PR fixes issue #590  (but does not fix issue #756 )

It might not be the best solution (the problem might be deeper in the system), but it is a 
one-line workaround that just does what is needed and prevents dashboard users from creating fatal damage to their Date arrays just by opening the field and losing focus thus saving changes.

datearray

BTW: I had troubles running the jest tests on windows, spend a few hours figuring out how to do it but I was not able to. It really drives me crazy. Could someone please try it on windows and update CONTRIBUTING.md with info how to do it? First I was getting `'NODE_PATH' is not recognized as an internal or external command`, then I found out I need to do some changes in package.json:
```
"build": "SET NODE_ENV=production && webpack --config webpack/production.config.js && webpack --config webpack/PIG.config.js",
    "test": "SET NODE_PATH=./node_modules && jest",
```
..and maybe also setting windows NODE_PATH environment variable and running `npm install jest` was needed, I don't know what exactly was needed, tried a lot of things. But then I ended up on 
```
Using Jest CLI v12.1.1, jasmine2
No tests found for "".
```
and I was not able to solve this issue. So the PR is not tested, can someone please do it?

### DIFF
--- a/src/dashboard/Data/Browser/BrowserTable.react.js
+++ b/src/dashboard/Data/Browser/BrowserTable.react.js
@@ -201,7 +201,15 @@ export default class BrowserTable extends React.Component {
             value = '';
           } else if (type === 'Array') {
             if (value) {
-              value = value.map(val => val instanceof Parse.Object ? val.toPointer() : (typeof val.getMonth === 'function' ? { __type: "Date", iso: val.toISOString() } : val));
+              value = value.map(val => {
+                  if (val instanceof Parse.Object) {
+                      return val.toPointer();
+                  } else if (typeof val.getMonth === 'function') {
+                      return { __type: "Date", iso: val.toISOString() };
+                  }
+
+                  return val;
+              });
             }
           }
           let wrapTop = Math.max(0, this.props.current.row * ROW_HEIGHT);
--- a/src/dashboard/Data/Browser/BrowserTable.react.js
+++ b/src/dashboard/Data/Browser/BrowserTable.react.js
@@ -201,7 +201,7 @@ export default class BrowserTable extends React.Component {
             value = '';
           } else if (type === 'Array') {
             if (value) {
-              value = value.map(val => val instanceof Parse.Object ? val.toPointer() : val);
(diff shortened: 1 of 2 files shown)


src/dashboard/Data/Browser/BrowserTable.react.js
```diff
@@ -201,7 +201,7 @@ export default class BrowserTable extends React.Component {
             value = '';
           } else if (type === 'Array') {
             if (value) {
-              value = value.map(val => val instanceof Parse.Object ? val.toPointer() : val);
+              value = value.map(val => val instanceof Parse.Object ? val.toPointer() : (typeof val.getMonth === 'function' ? { __type: "Date", iso: val.toISOString() } : val));
```
natanrolnik: I would prefer not having nested ternary operators here. What do you think?
And when the user edits the date in the array, is it saved back correctly?

rihadavid: Yes, it is saved correctly. Even the previous version was able to save provided `{"__type":"Date","iso":"2017-09-05T18:06:06.943Z" }` correctly as a valid Date, but it was then again displayed (after refresh) as `"2017-09-05T18:06:06.943Z"`. Now it is still displayed as the iso string, but when you open the cell, it adds this `{"__type":"Date","iso": .... }` so that it is saved correctly when update is confirmed :-)

How would you recommend to avoid nested ternary operators?

natanrolnik: I would do something like this: 

```
value = value.map(val => {
    if (val instanceof Parse.Object) {
        return val.toPointer;
    } else if (typeof val.getMonth === 'function') {
        return { __type: "Date", iso: val.toISOString() };
    }

    return val;
});
```

rihadavid:
@natanrolnik  Ok, I updated the code with the suggested formatting and tested there are no typos, is it OK now, can it be a part of the next parse dashboard release? :-) Thanks.

flovilmart:
Thanks @rihadavid looking great! Releasing a new version soon!

PR #771, opened 2017-09-12
Send audience ID with push body.

aontas:
Conditionally add an audience_id property if a saved audience is used for a push on the dashboard.


aontas:
@flovilmart Here is a dashboard PR for the audience_id updates on the parse-server.

flovilmart:
Thanks!

Issue #751, opened 2017-07-16
Since 1.0.20 it's impossible to use an external `passport` authentication


romanbsd:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.

- [x] You're running version >=2.3.2 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
parse-dashboard 1.1.0

#### Steps to reproduce
We're using an external authentication (Google OAuth2 with passport), since the commit https://github.com/parse-community/parse-dashboard/commit/be9f498efe78211e6478ec9ce2d4e2cb959830d0 
it's impossible to use external `passport` strategy.
The easiest fix would be to allow an external `authInstance` key in `config` which will provide `authInstance.initialize(app)` method, and use it instead of the Authentication if provided.

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

flovilmart:
Do you want to provide a PR for it?

flovilmart:
@romanbsd would you wanna do a PR for your solution? I'm of for abstracting away 1 step, as we could probably use the same logic.

mtrezza:
Closing due to inactivity

PR #772, opened 2017-09-18
Added support for not using the built-in Authentication

BenniPlejd:
Currently there is no way of disabling the built-in Basic Authentication that the Parse-Dashboards uses. This causes issues if you want to use your own custom Authentication middleware instead.

I've added a new Configuration option named `noAuthentication`. If its value is set to `true`, then the built-in Authentication (~/Parse-Dashboard/Authentication.js) is not initialized.


flovilmart:
@BenniPlejd I'd like to help on this one, but disabling authentication seems be be a bad idea, I'd rather have a new authentication provider, compatible with passport, instead of disabling it altogether. Then you're free to have security holes on your dashboard, but at least, we've provided a best effort to prevent users from inadvertently expose their dashboard. What do you think?

flovilmart:
Also, if you want to disable your authentication, and use your own middleware, just don't specify any authentication info in it, that would be easier I believe no?

romanbsd:
We used a bit different approach:
https://github.com/parse-community/parse-dashboard/issues/751

yamitzky:
I agree with this PR. I am using the parse dashboard in a private network, so it must be completely secure without any authentication.

yamitzky:
@flovilmart I agree with this PR. I am using the parse dashboard in a private network, so it must be completely secure without any authentication.

README.md
```diff
@@ -147,6 +147,8 @@ Parse Dashboard supports adding an optional icon for each app, so you can identi
 
 You can set `appNameForURL` in the config file for each app to control the url of your app within the dashboard. This can make it easier to use bookmarks or share links on your dashboard.
 
+If you wish to use a custom authentication system, then you can disable Parse Dashboard's built-in authentication system by setting `noAuthentication` to `false`. The default value is false if not specified.
```
acinader: i think this is backward.  

"you can disable Parse Dashboard's built-in authentication system by setting `noAuthentication` to `false`. "

should be true, no?

acinader:
While I defer to your judgment, my $.02 would be that an option to turn off authentication is not a bad idea.  I can imagine a few legitimate use cases: running on your local box or in a VPN environment.

The distinction between this and allowing email to be returned in an unauthenticated rest call is that I don't think that a lazy developer would use this feature just to avoid doing some work.

flovilmart:
If you don’t provide any users, there’s no auth...

acinader:
oh.  nevermind.

PR #774, opened 2017-09-23
Adds ability to set the cookie session secret

flovilmart:


flovilmart:
Yeah, reported a few times :(

PR #793, opened 2017-10-27
Hotfix

flovilmart:


flovilmart:
@montymxb I actually need to force push on master as reverting the bad commit count not help.

montymxb:
@flovilmart I see...that's not so great :/. Ideally we should just release a **1.1.2** rather than trying to retroactively patch the existing release, at least that's how it's usually recommended (even if the release is bugged).

flovilmart:
I unpublished 1.1.1 didn't do the force push, found a better way.

PR #794, opened 2017-10-27
⚡ Release 1.1.2

flovilmart:
- Reverts the GreenKeeper update.
- Fixes an issue affecting the read-only users.
- Bumps version for immediate release.


flovilmart:
The 1st commit is the rollback of the deps updates (which is causing all hell breaking loose).
The 2nd commit is a hot fix (that ensuring config.apps is not mutated)
The 3rd commit is just the version bump.

flovilmart:
Yeah react-router is breaking everything from 2 to 4, that's the big bad boy, perhaps we should try to bump everything but this one.

montymxb:
If the break is incredibly evident than we could either modify the tests to catch it, or just run through it ourselves to verify the dashboard still functions as expected. I would love to have a test that could potentially catch this however, just so it's automated.

montymxb:
One more thing too, I'm presuming this will rectify the docker build issue?

flovilmart:
Test would be quite complex to setup at this point but we could have a look from the broken commit. I assume docker should be fine

ctrongminh:
@flovilmart, you are so fast,

I was about to open a PR for the second commit (An issue introduced when using readOnlyMasterKey would make all users readOnly after one has logged in).

Thanks

PR #795, opened 2017-10-27
Include PushAudience query

marvelm:
Currently, selecting an audience results in push notifications being sent to all installations because the query is `{}`. This commit makes sure that the Push query (`{}`) is replaced with the query found in the Audience object.


montymxb:
Nice catch @marvelm

PR #796, opened 2017-10-27
Removes forcing sort on createdAt

flovilmart:
When sorting columns, an additional ordering was always added on createdAt which cause issues on large collection.

This PR removes this additional descending, therefore, index management should be easier.


davimacedo:
What do you think about automatically creating index for createdAt in Parse Server instead of removing the default order by?

flovilmart:
It’s not the default orderBy, but a secondary order on every single quet

davimacedo:
Got it. I haven't noticed that. It seems a good idea to remove it.

Issue #811, opened 2017-11-14
Can't open Push tab


Pikorua:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] You're running version >=1.0.23 of Parse Dashboard.

- [X] You're running version >=2.3.2 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
PARSE DASHBOARD 1.1.2
PARSE SERVER 2.6.5

#### Steps to reproduce
Open Parse Dashboard, click on PUSH in the left menu on the bottom.
Nothing seems to happen. 
When opening developer tools, in the console there is an error. This error happens when the push menu on the left is clicked. 

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

dashboard.bundle.js:47 Uncaught TypeError: Cannot read property 'filter' of undefined
    at t.value (dashboard.bundle.js:47)
    at d.performInitialMount (dashboard.bundle.js:32)
    at d.mountComponent (dashboard.bundle.js:32)
    at Object.mountComponent (dashboard.bundle.js:10)
    at d.performInitialMount (dashboard.bundle.js:32)
    at d.mountComponent (dashboard.bundle.js:32)
    at Object.mountComponent (dashboard.bundle.js:10)
    at d.performInitialMount (dashboard.bundle.js:32)
    at d.mountComponent (dashboard.bundle.js:32)
    at Object.mountComponent (dashboard.bundle.js:10)
value @ dashboard.bundle.js:47
performInitialMount @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:10
performInitialMount @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:10
performInitialMount @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:32
mountComponent @ dashboard.bundle.js:10
updateChildren @ dashboard.bundle.js:32
_reconcilerUpdateChildren @ dashboard.bundle.js:33
_updateChildren @ dashboard.bundle.js:33
updateChildren @ dashboard.bundle.js:33
_updateDOMChildren @ dashboard.bundle.js:33
[...]

PR #812, opened 2017-11-15
Fixes missing supportedPushLocales in #811

montymxb:
The newly added `supportedPushLocales` causes the **Push** section of the dashboard to break if it is not present. This adds a one-line check to set an empty array if push locales are not set. This does not remove the option to set a push locale however, it just provides no options as can be seen below.

localize

### DIFF
--- a/src/lib/ParseApp.js
+++ b/src/lib/ParseApp.js
@@ -60,7 +60,7 @@ export default class ParseApp {
     this.serverURL = serverURL;
     this.serverInfo = serverInfo;
     this.icon = iconName;
-    this.supportedPushLocales = supportedPushLocales;
+    this.supportedPushLocales = supportedPushLocales ? supportedPushLocales : [];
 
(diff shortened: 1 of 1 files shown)


montymxb:
@flovilmart you mean [this](https://github.com/parse-community/parse-dashboard#configuring-localized-push-notifications)?

flovilmart:
Yessir!

montymxb:
Cool, we should be good the if there's nothing else that comes to mind?

flovilmart:
I meant as an alert inside the section, explaining how to set it up.

montymxb:
Ohhhh, you mean as an actual log entry if the localizations are found not to be set?

montymxb:
@flovilmart Just bumping here. Added a console warning to indicate in the logs if push localization is missing altogether, along with a link to the relevant section in the docs.

PR #813, opened 2017-11-18
Push Audiences improvements

flovilmart:
- Fix issue affecting bad requests with audience_limit
- Adds support for sizing audiences in the dashboard running counts on _Installation table

### DIFF
--- a/src/lib/ParseApp.js
+++ b/src/lib/ParseApp.js
@@ -374,13 +374,17 @@ export default class ParseApp {
   }
 
   fetchPushSubscriberCount(audienceId, query) {
-    let path = '/apps/' + this.slug + '/dashboard_ajax/push_subscriber_count';
-    let urlsSeparator = '?';
-    if (query){
-      path += `?where=${encodeURI(JSON.stringify(query))}`;
-      urlsSeparator = '&';
+    let promise;
+    if (!query) {
+      promise = new Parse.Query('_Audience').get(audienceId, { useMasterKey: true }).then(function(audience) {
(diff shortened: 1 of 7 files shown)


src/lib/ParseApp.js
```diff
@@ -374,13 +374,17 @@ export default class ParseApp {
...
 
   fetchPushSubscriberCount(audienceId, query) {
-    let path = '/apps/' + this.slug + '/dashboard_ajax/push_subscriber_count';
-    let urlsSeparator = '?';
-    if (query){
-      path += `?where=${encodeURI(JSON.stringify(query))}`;
-      urlsSeparator = '&';
+    let promise;
+    if (!query) {
+      promise = new Parse.Query('_Audience').get(audienceId, { useMasterKey: true }).then(function(audience) {
```
flovilmart: we may wanna add caching for those counts as may be quite expensive on the DB.

flovilmart:
@montymxb i’d Like to get this one in for the release too

montymxb:
Yep yep, just saw this. Sorry I'm behind by a day or so and just going through everything.

src/lib/ParseApp.js
montymxb: @flovilmart do we want to add in caching with this PR or do we want to handle this later on?

montymxb: Also if this runs okay on your end this would be great to get in. About time we should put up a dashboard release so we can get past #811 .

flovilmart: Yep, caching would need more work, not required

montymxb: Ehhh, maybe in this case we should cache. Although we could take it as is it would be nice to not have to come back and do that later.

Issue #831, opened 2018-01-10
Regression: Filtering on Number attributes broken in 1.1.2. Caused by bad query string.


pgiguere1:
Using any of the dashboard's filter on a Number attribute (`equals`, `less than`, `greater than`...) will always yield zero results.

Reverting to 1.1.0 fixes the issue.

The issue is in the query string building. 1.1.2 and 1.1.0 will not build the same query string for the same filter. Pasting an URL generated by 1.1.0 in dashboard 1.1.2 will make the filter work as expected.

___

Here's the URL generated for class `MyClass` filtering on `myAttribute` equals `1`.

1.1.0 (correct):
`https://mydomain.com/apps/MyApp/browser/MyClass?filters=%5B%7B%22field%22%3A%22myAttribute%22%2C%22constraint%22%3A%22eq%22%2C%22compareTo%22%3A1%7D%5D`

1.1.2 (incorrect):
`https://mydomain.com/apps/MyApp/browser/MyClass?filters=%5B%7B%22field%22%3A%22myAttribute%22%2C%22constraint%22%3A%22eq%22%2C%22compareTo%22%3A%221%22%7D%5D`

The decoded equivalent:

1.1.0 (correct):
```
filters=[  
   {  
      "field":"myAttribute",
      "constraint":"eq",
      "compareTo":1
   }
]
```

1.1.2 (incorrect):
```
filters=[  
   {  
      "field":"myAttribute",
      "constraint":"eq",
      "compareTo":"1"
   }
]
```

The `compareTo` value shouldn't be enclosed in quotation marks.

Appears to have been broken by https://github.com/parse-community/parse-dashboard/commit/fcd9e51ced30ec7fb29b52051fd14fae77271f33.

pgiguere1:
Fixed by https://github.com/parse-community/parse-dashboard/pull/832

PR #832, opened 2018-01-10
Fix filtering with a 1-digit number (#831)

pgiguere1:
Fixes a bug introduced in 1.1.1 causing any 1-digit number filter to yield no results.

https://github.com/parse-community/parse-dashboard/issues/831


montymxb:
I want to take a look at this just to double check it all works as expected. We can merge after if it's all good.

montymxb:
Thanks again @pgiguere1 !

PR #836, opened 2018-01-28
Fix Databrowser shows correct count of filtered objects

engel:


montymxb:
@engel can you elaborate on this? .length should suffice...

engel:
After upgrading to parse-dashboard 1.1.2 from 1.0.28 I found that after filtering a table, the object count was not calculated correctly anymore. See screenshot. 

bildschirmfoto 2018-02-01 um 22 31 19

This PR fixes this by using List.size instead of List.length (as List.length is not defined in the List (Collection) class of the immutable package)

Additionally I found that v1.0.28 includes the "immutable": "~3.7.5" package while v1.1.2 uses     "immutable": "~3.8.1".
I assume that changes to the immutable package are the reason why filters.length not work anymore.

montymxb:
@engel just confirmed this, and went to their docs and found [size listed there](https://facebook.github.io/immutable-js/docs/#/List/of). I did not find any reference to length however, not sure if that ever was a valid member or not.

Issue #681, opened 2017-03-13
Setting different themes (colors) for different apps


oferRounds:
Would be awesome if we could the dashboard would support setting theme for different apps. I have 2 DBs for my app — one for Production, one for Stage — which are sometimes open simultaneously on the browser. So this would come handy as a mechanism to prevent mistakes (editing a row on one env, when you think you’re on the other one)

warrenca:
+1

PR #838, opened 2018-02-15
add background color in AppCard

AreyouHappy:
### add background color
Easily distinguish your app cards.

```
{
  "apps": [{
    "serverURL": "",
    "appId": "",
    "masterKey": "",
    "appName": "",
    "iconName": "",
    "backgroundColor":"#000000"
  }],
  "iconsFolder": "icons"
}
```

### DIFF
--- a/Parse-Dashboard/parse-dashboard-config.json
+++ b/Parse-Dashboard/parse-dashboard-config.json
@@ -4,7 +4,8 @@
     "appId": "",
     "masterKey": "",
     "appName": "",
-    "iconName": ""
+    "iconName": "",
+    "backgroundColor":"#000000"
   }],
   "iconsFolder": "icons"
 }
--- a/Parse-Dashboard/parse-dashboard-config.json
+++ b/Parse-Dashboard/parse-dashboard-config.json
@@ -5,7 +5,8 @@
     "masterKey": "",
     "appName": "",
     "iconName": "",
-    "backgroundColor":"#000000"
+    "primaryBackgroundColor": "",
+    "secondaryBackgroundColor": ""
   }],
   "iconsFolder": "icons"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Fix: Databrowser shows correct count of filtered objects, thanks to [Tom Engelbrecht](https://github.com/engel)
 
 ### 1.1.3
-* Feature: Add background color in AppCard, thanks to [AreyouHappy](https://github.com/AreyouHappy)
+* Feature: Add primaryBackgroundColor and secondaryBackgroundColor in AppCard, thanks to [AreyouHappy](https://github.com/AreyouHappy)
 
 ### 1.1.2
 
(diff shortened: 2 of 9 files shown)


dplewis:
Can you provide a screenshot?

AreyouHappy:
Here!

dplewis:
Thanks for getting started on this featured request! https://github.com/parse-community/parse-dashboard/issues/681

Can you add the background color to the data browser?

untitled

AreyouHappy:
Hmm ... 
I do not recommend it.

natanrolnik:
Thanks for your PR, @AreyouHappy .
If no value is configured, what value is used? Can you make a condition that, if `backgroundColor` isn't set, or if it's set to an empty string, it is not used?

flovilmart:
In my side I’m not sure I’m convinced by the feature.

dplewis:
@flovilmart It helps people like me with multiple servers. I can't count how many times I've changed production thinking it was on local. Adding colors would help.

AreyouHappy:
If the backgorundcolor is not entered, it is replaced with the default. ( It is CSS style)
The colors at Data brower don't look good and aren't in now.
Please let me know if you have a problem with that commit!
Thank you for your feedback.

dplewis:
It’s not that it doesn’t look good it should be an indicator that a different server is being used. 

Changing the color of the app card is the same as looking at the URL.

I would like to see something more semi-themed to avoid confusion.

AreyouHappy:
Hmm .. I don't have good sense of design.
Do you want the Data browser part to change color, too?
[1] Core part
[2] Classes part (Browser)
I'm concerned about how to make the difference between [1] and [2] colors.
Do you have any good ideas?

dplewis:
@AreyouHappy You could have a `primaryBackgroundColor` and `secondaryBackgroundColor`  with defaults for both.

AreyouHappy:
Would you try it?

dplewis:
@AreyouHappy Thanks for the update. I'll have some time today to try it out!

Parse-Dashboard/parse-dashboard-config.json
```diff
@@ -4,7 +4,8 @@
     "appId": "",
     "masterKey": "",
     "appName": "",
-    "iconName": ""
+    "iconName": "",
+    "backgroundColor":"#000000"
```
dplewis: Can you change background color here to primary and secondary defaults?

dplewis:
@AreyouHappy Just tried it out and it works great. Just a small change. 

Defaults are working without adding these fields.

AreyouHappy:
```
{
  "apps": [{
    "serverURL": "",
    "appId": "",
    "masterKey": "",
    "appName": "",
    "iconName": "",
    "primaryBackgroundColor": "",
    "secondaryBackgroundColor": ""
  }],
  "iconsFolder": "icons"
}
```
I changed it as follows.

dplewis:
@natanrolnik How does this look?

natanrolnik:
@dplewis I understand the need, and in my opinion, it's good: we still have the default version, and whoever wants to customize, it's his/her responsibility to do so. Where are the primary and secondary colors applied, in the end?

dplewis:
@natanrolnik In the app cards and databrowser. The above attached images.

natanrolnik:
Thanks @AreyouHappy and @dplewis.
Release 1.2.0 is already available :simple_smile:

PR #854, opened 2018-05-15
Fix Docker build

damarte:
Version 1.2.0 no build in https://hub.docker.com/r/parseplatform/parse-dashboard/builds/.


flovilmart:
Thanks! Let’s try it out

Issue #891, opened 2018-08-30
Don't make Everyone the default audience on Push


DevBakura:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] You're running version >=1.0.23 of Parse Dashboard.

- [X] You're running version >=2.3.2 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce
Open dashboard push section.
Go to send new push.
Everyone is selected as default audience.

#### Issue
Unfortunately today I made this mistake of sending a push to over 20k users.
To make is worse the push contained json as text.
This was clearly my fault. But I believe this can be prevented.

My suggestion is to don't have the audience selected to "Everyone" by default.
Or maybe require a confirmation window when sending a push to "Everyone" or over a X limit of users.

Also, and I think this is a bit more complex is to detect when the message contains json and is set to text. Maybe show a warning to the user.

flovilmart:
Is it something you want to implement?

PR #910, opened 2018-10-04
Issue: Don't make Everyone the default audience on Push

paulovitin:
I believe is the most simple solution for this #891


flovilmart:
Good idea!

Issue #830, opened 2018-01-05
Docker hasn't been building


JonDum:
https://hub.docker.com/r/parseplatform/parse-dashboard/builds/

> The command '/bin/sh -c cd /src  && npm install  && npm run build  && npm cache clear  && rm -rf ~/.npm  && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 1

Running it locally results in tons of module not found errors related to lodash: 

```
..  ModuleNotFoundError: Module not found: Error: Can't resolve 'lodash/without' in '/src/node_modules/react-dnd-html5-backend/lib'
  ModuleNotFoundError: Module not found: Error: Can't resolve 'lodash/memoize' in '/src/node_modules/react-dnd-html5-backend/lib'
...
 ModuleNotFoundError: Module not found: Error: Can't resolve 'disposables' in '/src/node_modules/react-dnd/lib'
...
ModuleNotFoundError: Module not found: Error: Can't resolve 'lodash/isArray' in '/src/node_modules/react-dnd/lib/utils'
      at factoryCallback (/src/node_modules/webpack/lib/Compilation.js:264:39)
      at /src/node_modules/webpack/lib/NormalModuleFactory.js:247:20
      at /src/node_modules/webpack/lib/NormalModuleFactory.js:65:21
      at /src/node_modules/webpack/lib/NormalModuleFactory.js:138:21
      at /src/no
...
etc
```

I'm guessing that has something to with what how NPM installs deps in node:argon-slim.
Perhaps using node:latest or node:alpine-latest will result in a clean build.

I will try and submit PR if successful.

flovilmart:
Thanks!

JonDum:
Ok I updated the Dockerfile to be much leaner and faster. 73mb total vs 396mb from the old one.
I also added non pid 1 (root) exec, proper signal handling via tini and node_env prod as a default.

https://github.com/JonDum/parse-dashboard/commit/9fc72a2b4211c52001117e0ce1430e13ab6565f9

I'm not entirely sure that I have the proper set up though. What's the relationship between ./Parse-Dashboard and production/bundles? Ideally you'd `npm run build` before hand and commit the results in the docker image, but I'm not sure if that would work. Need some guidance there.

you can play around with the resultant files like this:

```
docker-build -t parse-db .
docker-run -it parse-db sh
```
Everything is in /home/node as is custom from the `node:<>` images from upstream.

flovilmart:
[...]

Issue #640, opened 2017-01-25
Upgrade Dockerfile


mrmarcsmith:
The dockerfile still uses `ENTRYPOINT ["npm", "run", "dashboard"]` which as I understand is no longer how we are supposed to do things. so would an alternative ENTRYPOINT command be `ENTRYPOINT ["parse-dashboard"]` ?
I'm not very familiar with npm so sorry if this is a simple question.
original Dockerfile:
```
FROM node:4.4.2
WORKDIR /src
ADD . /src
RUN cd /src \
 && npm install \
 && npm run build \
 && npm cache clear \
 && rm -rf ~/.npm \
 && rm -rf /var/lib/apt/lists/*
ENTRYPOINT ["npm", "run", "dashboard"]
```
my current workaround which I know doesn't build from the src folder.
```
FROM node:4.4.2
RUN npm install -g parse-dashboard
ENTRYPOINT ["parse-dashboard"]
```

flovilmart:
> ["npm", "run", "dashboard"] which as I understand is no longer how we are supposed to do things

Where did you read that? I'm curious.

mrmarcsmith:
in issue https://github.com/ParsePlatform/parse-dashboard/issues/165 @drew-gross said

> It seems like you are using the old npm run dashboard method of using the dashboard. We overhauled the installation process recently to use npm install instead, I recommend you try that method. You can find more details in the readme.

I found that thread because I was experiencing the same error as the op when using docker.

The errors:
```
Container#eachRule is deprecated. Use Container#walkRules instead.
Container#eachDecl is deprecated. Use Container#walkDecls instead.
Node#removeSelf is deprecated. Use Node#remove.
30% 82/240 build modulesNode#before is deprecated. Use Node#raws.before
```

PR #912, opened 2018-10-12
Updated Dockerfile to use a multi-stage build.

bgsandan:
**Summary:**

This pull request updates the Dockerfile for the project to improve the build artifacts and deployment speed of the built container.  This is my first PR so if it needs formatting differently please let me know :)

**Related issues:** 

#830 and #640 

**Summary of changes:**

- Switched to alpine linux base image for smaller image sizes
- Created a multi-stage build allowing final image to only contain production dependencies
- All webpack compilation occurs during the docker build phase rather than some occurring when the container is executed.  This drastically reduces the time-to-live for the deployed dashboard

**Performance Impact**

Whilst initial container build time is increased, the container size and launch time are decreased.  The refactored Dockerfile should benefit from layer-caching when used in an iterative development cycle

Metrics (collected from 'cold' on a clean [Play with Docker](https://labs.play-with-docker.com/) instance):

|                | Container Build Time | Container Launch Time | Container Size |
|----------------|----------------------|-----------------------|----------------|
| Current Master | 1m 47.608s            | 0m 25.552s             | 320MB          |
| This PR        | 2m 42.572s            | 0m 1.328s              | 74.6MB         |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,35 @@
-FROM node:8-slim
-ENV NPM_CONFIG_LOGLEVEL error
+#
+# --- Base Node Image ---
+FROM node:8-alpine AS base
+
 WORKDIR /src
-ADD . /src
-RUN cd /src \
- && npm install \
- && npm run build \
- && npm cache clear --force \
- && rm -rf ~/.npm \
- && rm -rf /var/lib/apt/lists/*
-ENTRYPOINT ["npm", "run", "dashboard"]
+
+# Copy package.json first to benefit from layer caching
+COPY package*.json ./
+RUN npm install --only=production
+# Copy production node_modules aside for later
+RUN cp -R node_modules prod_node_modules
+# Install remaining dev dependencies
+RUN npm install
+
+COPY . /src
+
+# Run all webpack build steps
+RUN npm run prepublish && npm run build
+
+
+#
+# --- Production Image ---
+FROM node:8-alpine AS release
+WORKDIR /src
+
+# Copy production node_modules
+COPY --from=base /src/prod_node_modules /src/node_modules
+COPY --from=base /src/package*.json /src/
+
+# Copy compiled src dirs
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,11 +1,35 @@
...
+
+# Copy production node_modules
+COPY --from=base /src/prod_node_modules /src/node_modules
+COPY --from=base /src/package*.json /src/
+
+# Copy compiled src dirs
+COPY --from=base /src/Parse-Dashboard/ /src/Parse-Dashboard/
+
+
+EXPOSE 4040
```
flovilmart: 4040 is only the default port. Is it relevant if the users has a config that listens to another port?

bgsandan: Yep, good spot!  I don't think there is much point a user changing the listening port when running in Docker seeing as they can map the container port to a port of their choice on the host system, but to support this use case (and people porting their existing config to Docker) I've removed the default 4040 from the Dockerfile

Dockerfile
```diff
@@ -1,11 +1,33 @@
...
+WORKDIR /src
+
+# Copy production node_modules
+COPY --from=base /src/prod_node_modules /src/node_modules
+COPY --from=base /src/package*.json /src/
+
+# Copy compiled src dirs
+COPY --from=base /src/Parse-Dashboard/ /src/Parse-Dashboard/
+
+ENTRYPOINT ["npm", "run", "start"]
```
flovilmart: npm start is enough, no need to use `run`.

Also, I believe it is better to use node directly than using npm to start the server so we should use:

`node ./Parse-Dashboard/index.js` for the ENTRYPOINT

bgsandan: I've updated the entrypoint and everything seems to still be working nicely :)

flovilmart:
Thanks for the Pull Request! This surely will make the dashboard image lighter!

Issue #482, opened 2016-07-29
Time zone support


benpackard:
It would be great to have some way to specify the timezone to use when displaying dates in Parse Dashboard `createdAt`, `updatedAt`, etc.

Vortec4800:
Commenting on this as a vote, I would like to see time zone support as well.

tarekskr:
+1

mistahgandhi:
+1

yomybaby:
+1

AnChiChang:
+1

Matt-Allen44:
+1

xcarpentier:
@flovilmart any idea on this topic?

onosol:
Bump - this would be an ideal feature. +1000

jenlai1345:
+1 please

bfjfm2:
+1

ananfang:
+1

AlexShul:
+1

adolfdsilva:
+1

vanics:
+1

JEKos:
+1

alxmllr:
+1

Simone-cogno:
+1

hyamsj:
+1

pash2048:
Hi @mtrezza, May I work on this?

mtrezza:
I think you would make a lot of developers happy, looking at the +1s here.
Do you have any idea on how you want to implement this?

pash2048:
Dear @mtrezza, I need this feature too, so very cool. Thumbs up! I start working on this :))
at the first, I need to add the `time_zone` parameter to the `parse-dashboard-config.json`, then only show the date-time in the selected timezone, for store date, send UTC date time to the server.
everything is alright?

mtrezza:
I think that's a good approach - to start with the most simple form of the functionality. Later on, if someone wants to, they can add a UI to change that param in the dashboard.

What if someone enters a date via the date picker in the data browser. Should that be the local date and converted to UTC, or sent as it?

benpackard:
@pash2048 Here's something you might want to consider (just chipping in as the original opener of this issue, happy to see it might get picked up!)

I suspect most people would prefer to see these dates and times in one of two ways:

- UTC (existing behavior)
- Their current timezone

[...]

PR #914, opened 2018-10-13
Timezone configuration

paulovitin:
#482


flovilmart:
This is not a feature we want to support in the future. This can lead to mis understandings, and major fuck ups. DB dates should always be stored and shown in UTC.

mtrezza:
Absolutely. 

> This is not a feature we want to support in the future. This can lead to mis understandings, and major fuck ups. DB dates should always be stored and shown in UTC.

For future reference, a place where time zone localization makes sense is time references in data that is derived (and thus clearly separated) from the original data, for example in Analytics.

PR #920, opened 2018-10-26
Add distinct / unique filter

dplewis:
Closes: https://github.com/parse-community/parse-dashboard/issues/888

I don't really know the conventions for react, so any help would be appreciated.

screen shot 2018-10-26 at 2 25 03 pm

More testing will be needed as I may have missed something.


dplewis:
@flovilmart Quick update

Add / Delete rows are disabled if filtering by unique.

All fields are readonly

It works for all fields except arrays. Discussed here https://github.com/parse-community/parse-server/pull/4195#issuecomment-331300980

I could use aggregate to make array work.

I could also add not unique filter, discussed here https://github.com/parse-community/parse-server/issues/5019

This haven't been tested on Postgres

flovilmart:
Can you quickly test over on Postgres? Worst case add a flag on parse server features

dplewis:
@flovilmart A quick fix for PG and everything works

flovilmart:
@natanrolnik what do you think? Not well versed in this area ;)

natanrolnik:
@flovilmart sorry the late reply.
Although react is also not my strongest skill, the changes look fine to me.
@dplewis would you mind adding a few more screenshot?

dplewis:
@natanrolnik What would you like to see?

dplewis:
Date

screen shot 2018-11-01 at 10 19 22 am

Object

screen shot 2018-11-01 at 10 19 07 am

Pointer

screen shot 2018-11-01 at 10 18 42 am

screen shot 2018-11-01 at 10 18 49 am

natanrolnik:
Looks good. My only comment:

From what I see here, the “Add a column” button is still enabled. Do you think it makes sense to disable it? I think it does.

dplewis:
Distinct queries don't have filtering fyi

screen shot 2018-11-01 at 1 46 04 pm

dplewis:
@natanrolnik @flovilmart I just tried this against a Collection with 305k unique records and received a 504 (Gateway Timeout).

acinader:
@dplewis what do you want to wit this?

dplewis:
@acinader Waiting for review. I did find that on large datasets this can cause a timeout, if you haven't set timeout / index the field.

acinader:
what do you think about that?

dplewis:
A popup box with a warning or wait until indexes get added to dashboard

acinader:
can you catch the timeout?

dplewis:
After updating my server config maxTimeMS I don't get that timeout error.

PR #936, opened 2018-12-08
Add tvOS support to push targeting

toto:
This goes together with the update done here https://github.com/parse-community/Parse-SDK-iOS-OSX/pull/1375 to add support for tvOS to the iOS/OSX SDK. 

It basically just adds the `tvos` audience to the push UI.


flovilmart:
Thanks! That sounds good!

PR #949, opened 2019-01-10
Activate stalebot

dplewis:


dplewis:
@flovilmart It still has to be configured for the project.

flovilmart:
Yep i’ll Have a look or can you?

dplewis:
I can't, organization owner only. That would be you.

flovilmart:
Looks active!

PR #952, opened 2019-01-10
Unhandled Promise Rejection fix

dplewis:
If there is an error connecting to the server, don't run count operations

Updates package-lock.json with react-helmet


flovilmart:
LGTM!

PR #960, opened 2019-01-17
Add preventSchemaEdits option to apps and hide or disable buttons to manipulate classes and columns

cyrilchandelier:
In a project using Parse Dashboard as a middleware of a Parse Server, I want to disable the actions modifying the schema.

I know there is way to fail these actions when using the `readOnlyMasterKey`, but I need to let my user add / edit / delete rows, I just want them to not change the schema. The best way I found so far is to hide the buttons:
- "Create a class" in the Browser
- "Add a column" in the BrowserTable
- "Security" and all options related to columns and classes from the Edit dropdown

To support that, I added a new `preventSchemaEdits` option to apps, defaulted to `false`.

Eg:
```
{
  "apps": [{
    "serverURL": "http://localhost:1337/parse",
    "appId": "hello",
    "masterKey": "world",
    "appName": "",
    "iconName": "",
    "primaryBackgroundColor": "",
    "secondaryBackgroundColor": "",
    "preventSchemaEdits": true
  }],
  "iconsFolder": "icons"
}
```

Any feedback welcome.

### DIFF
--- a/src/components/DataBrowserHeaderBar/DataBrowserHeaderBar.react.js
+++ b/src/components/DataBrowserHeaderBar/DataBrowserHeaderBar.react.js
@@ -16,7 +16,7 @@ export default
 @DragDropContext(HTML5Backend)
 class DataBrowserHeaderBar extends React.Component {
   render() {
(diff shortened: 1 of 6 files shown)


flovilmart:
Thanks for the PR!

PR #962, opened 2019-01-19
Adds dev flag

flovilmart:
when running with docker, as the local IP address is different from the dashboard
all the security features would light up.

This --dev flag let anyone run the dashboard locally easily


Parse-Dashboard/index.js
```diff
@@ -35,6 +36,8 @@ const mountPath = program.mountPath || process.env.MOUNT_PATH || '/';
 const allowInsecureHTTP = program.allowInsecureHTTP || process.env.PARSE_DASHBOARD_ALLOW_INSECURE_HTTP;
 const cookieSessionSecret = program.cookieSessionSecret || process.env.PARSE_DASHBOARD_COOKIE_SESSION_SECRET;
 const trustProxy = program.trustProxy || process.env.PARSE_DASHBOARD_TRUST_PROXY;
+const dev = program.dev;
+console.log(dev);
```
acinader: remove console.  

Also, should probably make an env key too. PARSE_DEV?

flovilmart:
This is different as it allows for bypassing user authentication requirement

acinader:
right.

PR #966, opened 2019-01-25
Fix sidebar app sort

nafsimao:


flovilmart:
You guys should really get it together. Publishing your private forks like that...

Issue #968, opened 2019-01-30
Default to empty value for column removal dropdown


JStonevalley:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [ x ] You're running version >=1.0.23 of Parse Dashboard.

- [ x ] You're running version >=2.3.2 of Parse Server.

- [ x ] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Not Applicable

#### Steps to reproduce
We have by accident removed columns from our database when manually removing rows due to missclicks and the fact that the Remove column dialog dropdown is preselected with a random column. I think a better alternative would be that the dropdown is empty by default to force the user to have to pick the correct column for removal.

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

JStonevalley:
I would be happy to submit a pull request for this if that helps.

flovilmart:
That's a good idea!

JStonevalley:
@flovilmart Do you want me to submit a pull request and if so are there any guidelines that I should read first. I guess I also need some kind of permission.

flovilmart:
@JStonevalley yes you can go ahead with the pull request. We do not have any particular guidelines besides making sure the style is consistent and automated checks pass. Thanks!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

JStonevalley:
@flovilmart #969 PR is live.

PR #969, opened 2019-02-01
feat(RemoveColumnDialog): Remove default value and add placeholder

JStonevalley:
#968 Remove default value in 'Remove column modal' add a placeholder and disable button if no column is selected.


flovilmart:
LGTM Thanks!

PR #984, opened 2019-02-12
Fix #983 verify the whole array to check Parse.Objects into generic objects

### LINKED ISSUE #983: The dashboard does not work well with nested objects
Hi.

I have the next issue:

When I save a list of generic objects and one of their children is a Parse.Pointer the dashboard does not interpret it well

Example
Original array with nested object
```JSON
[
    {
        "text": "Hello",
        "myPointerObject": {
            "__type": "Pointer",
            "className": "Producto",
            "objectId": "K4vUSfxkMx"
        }
    }

]
```

And parse dashboard made the next evaluation:

Bad formed array with nested object

```JSON
[
    {
        "text": "Hello",
        "myPointerObject":{
            "objectId": "K4vUSfxkMx"
        }
    }

]
```

The problem comes when you try to see the whole json, and the auto-save sistema of dashboard overwrite the original object to the bad formed.

I already fix the issue in the parse dashboard code, tomorrow i will make a PR with the solution.

dannywolfmx:
Recursive function to deal with nested objects and check if some children is a Parse.Object type


dplewis:
Can you post a screenshot?

dannywolfmx:
Hi sure.

"Producto is a parse object"

After

bug 1

bug

And this is my solution

after1
after 2


The problem with the actual dashboard is when you try to edit the cell, and the dashboard overwrite the parse server object keeping only the id and losing the reference

dplewis:
@dannywolfmx Thanks for starting this PR. There are two parts to this issue, displaying and editing.

This PR handles displaying properly.

There is an error with editing an array of array of objects.

`testObject.set('arrayField', [[object]])`

It stores correctly in the database as `[[ '__type': 'Pointer' ...]]` but if you edit it will save the full object.

https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Data/Browser/BrowserTable.react.js#L204

dannywolfmx:
Interesting, let's me check the PR to solve the issues.

dplewis:
I added some nits make sure you pull it down

dplewis:
@dannywolfmx Can we get this one in? For the editing you can use similar recursion.

dplewis:
@dannywolfmx @masonlee Can you guys test this out? I added a fix for editing

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -43,7 +43,21 @@ let BrowserCell = ({ type, value, hidden, width, current, onSelect, onEditChange
   } else if (type === 'Boolean') {
     content = value ? 'True' : 'False';
   } else if (type === 'Array') {
-    content = JSON.stringify(value.map(val => val instanceof Parse.Object ? val.toPointer() : val))
+    const detectObject = (value) => {
```
acinader: can you pull this function out so its not in the middle of this huge if/else?

src/dashboard/Data/Browser/BrowserTable.react.js
```diff
@@ -203,15 +203,18 @@ export default class BrowserTable extends React.Component {
...
-              value = value.map(val => {
-                  if (val instanceof Parse.Object) {
-                      return val.toPointer();
-                  } else if (val && typeof val.getMonth === 'function') {
-                      return { __type: "Date", iso: val.toISOString() };
-                  }
-
-                  return val;
-              });
+              const detectObject = (value) => {
```
acinader: same?

dannywolfmx:
Hi guys, sorry for being late.

Unfortunately I had to work on other projects so I could not devote much time to review this code.

I hope this weekend to be able to finish fixing the bug commented by @dplewis.

It is correct @acinader, it is better to move that block of code out of the if / else so that it can be reused.

I Have a question, Wouldn't it be better to allow the developers to place comments so that other users understand better than the code does?, I find it complex understanding that makes each module to be able to locate the bug.

dplewis:
@dannywolfmx I went ahead and fixed the bug I mentioned. Just need another set of eyes for testing.

> Wouldn't it be better to allow the developers to place comments so that other users understand better than the code does?

Can you elaborate?

acinader:
@dannywolfmx there are no comments in that file, but there are in others.

definitely feel free to comment the code!

dannywolfmx:
@dplewis Ok, I'll review it this weekend

dannywolfmx:
mmm, acinader you are right. If the user makes an array date types, it will be processed by the "JSON.stringify" and not by the dateStringUTC.

dannywolfmx:
We Could separate that great block of if/else in another function (setContent or setValue), this way would be easier to do the same thing that does  "detectArray " but including all types

dplewis:
Dates are currently handled (with the fixes)

Before Edit (From Database)

```
"arrayTest": [
        [
            {
                "__type": "Pointer",
                "className": "TestObject",
                "objectId": "agBGsAr7uf"
            }
        ],
        {
            "$date": "2019-04-02T20:07:33.057Z"
        },
        [
            {
                "__type": "Date",
                "iso": "2019-04-02T20:07:33.057Z"
            }
        ]
    ],
```
Screen Shot 2019-04-02 at 4 06 35 PM

After Edit

```
"arrayTest": [
        [
            {
                "__type": "Pointer",
                "className": "TestObject",
                "objectId": "agBGsAr7uf"
            }
        ],
        {
            "$date": "2019-04-02T20:07:33.057Z"
        },
        [
            {
                "__type": "Date",
                "iso": "2019-04-02T20:07:33.057Z"
            }
        ]
    ],
```

Screen Shot 2019-04-02 at 4 10 28 PM

If you refresh it goes back to how it looked in before edit (note the data in the database is the same even though they look different for view / edit.

dplewis:
Closing via https://github.com/parse-community/parse-dashboard/pull/1223

@dannywolfmx Thanks for getting started on this!

Issue #1054, opened 2019-04-11
The single app redirect does not keep the root path of the dashboard


Helmikku:
- [x] You're running version >=1.0.23 of Parse Dashboard.
- [x] You're running version >=2.3.2 of Parse Server.
- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup

Parse Dashboard 1.3.0
Parse Server 2.8.4
express 4.16.4
node 10.15.3

#### Steps to reproduce

```
const Application = require('express')();
const ParseDashboard = require('parse-dashboard');
const ParseServer = require('parse-server').ParseServer;
Application.use('/parse', new ParseServer ({ ... }));
Application.use('/dashboard', new ParseDashboard ({ ... }));
require('http').createServer(Application).listen(process.env.PORT || '80', () => {
	console.log('Parse Server is at your service.');
});
```

Then `http://localhost/dashboard/apps` will redirect to `http://localhost/apps/nameOfApp`

Any shortcut such as:

http://localhost/apps/nameOfApp/browser/_User => fails
http://localhost/dashboard/apps/nameOfApp/browser/_User => redirects to http://localhost/apps/nameOfApp/browser/_Installation

Thanks for your help

zenz:
Same situation here.
Parse-Server:3.2.3
Parse-Dashboard:1.3.0

dplewis:
Closing via https://github.com/parse-community/parse-dashboard/pull/1100

Please update to the latest version of the dashboard

grx7:
Can you re-open? Since parse/apps will redirect to parse/apps/app-name/browser/_Role 

Dahsboard 2.0.5
v10.19.0

PR #1067, opened 2019-04-30
Fix dashboard not applying mounting point when linking or redirecting to another page

cyrilchandelier:
TL;DR: single app redirection is not the root of our issue here, all mounted dashboards seem broken when using 1.3.0.

When upgrading to the latest version of the dashboard, I encountered the same issue as folks in #1054: app redirects and doesn't keep the mounting point.

Since I made the single app redirection changes in the first place (#958) and it was just a couple lines of code, I started investigating and I'm now 100% confident the problem was misattributed to the single app redirection in the first place.

Unless I'm missing something, something changed between 1.2.0 and 1.3.0. Commenting the code I wrote in #958 didn't fix the issue. I tracked the bad redirection to the following line:
```
// Dashboard.js#312
<Redirect from='/' to='/apps' />`
```

But that's not it! Basically, it looks like the mounting point is lost and the navigation is totally broken when the mountPath is not `/`. I've no idea how it was even working before since the whole app is referring/redirecting to `/apps` everywhere, without referencing the eventual mount path.

This PR is the result of a painful work trying to add the mountPath everywhere I found it was needed, but to be honest, I had no idea was I was doing the whole time, there might be a simple solution.

@flovilmart: I would love your assistance and opinion on this, thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2377,7 +2377,6 @@
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
-      "optional": true,
       "requires": {
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
@@ -3659,8 +3658,7 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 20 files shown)


flovilmart:
I believe the problem is somewhere else and that rewriting it all was not necessary. I may be wrong. I don’t maintain this project anymore, so check with @acinader, he may be able to help

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

acinader:
@dplewis would you be willing to look at this?  I don't have the problem...

dplewis:
@acinader Just looked into it. A lot of dependencies were updated. The problem is in the `history` package. You have to set a basename (base url / mountpath) to fix this issue.

I can do a separate PR with a sample server for testing.

@cyrilchandelier Thanks for getting started on this, I know it must have taken a lot of time.

dplewis:
Superced via https://github.com/parse-community/parse-dashboard/pull/1070

GoMino:
Can we hope for a new release with the fix? @dplewis

TomWFox:
@GoMino I've opened a PR for a new release - #1083

Issue #1049, opened 2019-04-04
Uncaught SyntaxError: Unexpected token <


Treverr:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] You're running version >=1.0.23 of Parse Dashboard.

- [X] You're running version >=2.3.2 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce
npm install -g parse-dashboard
sudo parse-dashboard --config /Users/Trever/parsedash.json --allowInsecureHTTP=1

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

```
> parse-dashboard@1.1.2 dev /usr/local/lib/node_modules/parse-dashboard
> node ./Parse-Dashboard/index.js & webpack --config webpack/build.config.js --devtool eval-source-map --progress --watch

module.js:557
    throw err;
    ^

Error: Cannot find module '/usr/local/lib/node_modules/parse-dashboard/webpack/build.config.js'
    at Function.Module._resolveFilename (module.js:555:15)
    at Function.Module._load (module.js:482:25)
    at Module.require (module.js:604:17)
    at require (internal/module.js:11:18)
    at module.exports (/usr/local/lib/node_modules/parse-dashboard/node_modules/webpack/bin/convert-argv.js:80:13)
    at Object.<anonymous> (/usr/local/lib/node_modules/parse-dashboard/node_modules/webpack/bin/webpack.js:39:40)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! parse-dashboard@1.1.2 dev: `node ./Parse-Dashboard/index.js & webpack --config webpack/build.config.js --devtool eval-source-map --progress --watch`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the parse-dashboard@1.1.2 dev script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

PR #1070, opened 2019-05-02
Fix issue regarding lost mount path

dplewis:
Can be tested by setting process.env.MOUNT_PATH.

Test Added


dplewis:
@acinader How does this look? That TODO is a reminder for future e2e tests.

I’d recommend giving it a shot when you have time. I ran into a few issues when I tried to mount a parse server.

acinader:
@dplewis I've taken a first (second and third) pass at testing and I have run into problems.  I don't know if it is because of this change combined with how I run the dashboard or if it is my unfamiliarity with how to test the dashboard.  I suspect the later.

Here's my setup:

I have a repo with package.json that includes the dependency:  `parse-dashboard: latest`.  I have a `parse-dashboard-config.json` file that has entries for all my pars servers.

The package.json has this script:

`"start": "parse-dashboard --config parse-dashboard-config.json",`

When I try to change the dependency from `latest` to `github:parse-community/parse-dashboard#mount-fix`

I get a blank white page with following js error in the console: `Uncaught SyntaxError: Unexpected token <`

if I click through to login.bundle.js in the inspector, they look very different, so I suspect the issue is that I don't know how to test a branch and build it, etc.  

I'll give it another try in a few hours when I have some time to look into it.

dplewis:
@acinader There is a known issue with that https://github.com/parse-community/parse-dashboard/issues/1049 I fixed that in my last commit try to install from branch again.

acinader:
ahh, much better!  thanks.

.gitignore
```diff
@@ -11,3 +11,6 @@ npm-debug.log
 *.swp
 .env
 .idea/
+
+logs/
+test_logs
```
acinader: add trailing cr

acinader:
ps: thanks for spoon feeding me.  much appreciated...

larcoamir:
> @acinader There is a known issue with that #1049 I fixed that in my last commit try to install from branch again.

Could please you suggest me git commands to install from the branch?

mtrezza:
@larcoamir I've noticed you've left several comments on years old threads. Your chances to receive support this way are rather slim as we use GitHub only for code related issues, not for support. For help with Parse Platform we recommend our [community forum](https://community.parseplatform.org).

PR #1098, opened 2019-06-06
Update Dockerfile

dplewis:
Adds git to the $PATH, this may increase the build size.

https://cloud.docker.com/u/parseplatform/repository/registry-1.docker.io/parseplatform/parse-server/builds/112a03a6-ef2e-4c6b-bb8f-9802aca574a9

Screen Shot 2019-06-06 at 4 24 02 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,9 @@
 # --- Base Node Image ---
 FROM node:8-alpine AS base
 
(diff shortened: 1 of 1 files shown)


dplewis:
@acinader I randomly found the build error under the parse-server build errors for some reason.

Issue #1097, opened 2019-06-05
can't build on Windows


oallouch:
@dplewis Could you do on the dashboard what you did on the server, which is using cross-env ?
Even if I don't use the dashboard, I'd like to contribute.

dplewis:
I'll investigate it soon, unless you want to give it a shot?

oallouch:
It's the same as what you've done on the server. But, if you really don't have time, I can try.

dplewis:
I use a Mac, should get my hands on a windows computer soon

PR #1115, opened 2019-06-19
Support building on Windows

dplewis:
Closes: https://github.com/parse-community/parse-dashboard/issues/1097


dplewis:
@oallouch Can you check if this works for you?

oallouch:
"npm run dashboard" : the server starts, but it doesn't update when I change a file (I tried to modify the sidebar title in SidebarHeader.react.js)
"npm run build" : seems to work, but I don't see my modification
"npm run pig" : localhost:4041 gives my a blank page

dplewis:
`npm run dashboard` should also call --watch. Did you refresh the browser / clear cache?
`npm run pig` I ran into the same issue, it will need investigation.

oallouch:
ok, "npm run build" does indeed work. I just need to clear the cache.
but --watch doesn't

The PIG issue seems to be module related. Maybe there should be "export default" instead of "module.exports = "

JeromeDeLeon:
I'm using windows and I just run `npm install` and `npm run pig` access `localhost:4041` but got invalid response.
image

dplewis:
@JeromeDeLeon Can you open this in a separate issue? It might get lost here.

Also include any error messages you received. Does `npm run dashboard` work for you?

PR #1123, opened 2019-06-26
GraphQL API playground

douglasmuraoka:
Adds the `GraphQLConsole` component under the `API Console` category, which allows users to test the `parse-server` GraphQL API. The component should be mounted with `X-Parse-Application-Id` and `X-Parse-Master-Key` headers, and the `parse-server`'s GraphQL API endpoint set. It means it will be ready to execute queries/mutations as soon as it is opened.

All the information required (`applicationId`, `masterKey` and `graphQLServerURL`) must be defined on the `ParseApp` instance in order to work properly.

Parse Dashboard GraphQL Playground


davimacedo:
What happens if graphQLServerURL is not set? I wanted to make sure that either the GraphQL console will not show up or we will have a friendly message introducing GraphQL as most of the developers have not yet setup GraphQL for their projects and no errors should be displayed in this case.

douglasmuraoka:
@davimacedo I was thinking about using the `EmptyState` component. The same used when there are no `Jobs` scheduled, for instance. It will show a title, a message (which contains a link to the Parse documentation, GraphQL section), and an icon.

What do you think?

Issue #1045, opened 2019-04-03
Can't switch (dropdown list) between servers


danibjor:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard. v1.3.0

- [x] You're running version >=2.3.2 of Parse Server. v3.2.3

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Running dashboard for cli/npm

#### Steps to reproduce
- add two or more servers to the dashboard config
- select a server, open class browser
- try changing Parse server in the dropdown list below the logo/header

Dropdown don't show up as expected. Chrome dev tools/console shows 2 js errors.
 
#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

`dashboard.bundle.js:22 Error: Invariant failed`
`dashboard.bundle.js:50 Uncaught Error: Invariant failed`

![image](https://user-images.githubusercontent.com/2149381/55479400-4e0d1a80-561e-11e9-9dd9-72a5b94926d9.png)

acinader:
I can reproduce this.

j4zz:
Can confirm that it broke for us with 1.3.0 as well.

Rgf11111:
1.3.0 get same problem too. Downgrade to 1.2.0 is worked.

acinader:
@douglasmuraoka FYI :).

douglasmuraoka:
> @douglasmuraoka FYI :).

I'll try to fix it as soon as I finish the GraphQL console :)

douglasmuraoka:
@acinader could you review my commit?

PR #1125, opened 2019-06-28
Fix: Can't switch (dropdown list) between servers #1045

douglasmuraoka:


davimacedo:
fixes https://github.com/parse-community/parse-dashboard/issues/1045

PR #1142, opened 2019-07-15
feat: Add client key into GraphQL console by default

douglasmuraoka:
Adds the `clientKey` into `GraphQLConsole` component as `X-Parse-Client-Key` header.


src/dashboard/Data/ApiConsole/GraphQLConsole.react.js
```diff
@@ -32,6 +32,7 @@ export default class GraphQLConsole extends Component {
     } else {
       const headers = {
         'X-Parse-Application-Id': applicationId,
+        'X-Parse-Client-Key': clientKey,
```
davimacedo: The client key may be not available. Can you please check and only add the header in the case it exists?

douglasmuraoka: Done!

PR #1164, opened 2019-07-19
fix: AppsMenu broken layout when app name is too long

alencarlucas:
This PR fixes the layout issue that occurs when an app name is too long.

The screenshot of the issue:
image

and after this PR:
image


acinader:
@alencarlucas nice!  thanks.

PR #1172, opened 2019-07-22
fix: Can't access method 'hasOwnProperty' from target object

alencarlucas:


davimacedo:
Reference: https://eslint.org/docs/rules/no-prototype-builtins

PR #1180, opened 2019-07-23
feat: Include "count" CLP option

### LINKED ISSUE #1165: Include `count` CLP option in Parse Dashboard
According to https://github.com/parse-community/parse-server/blob/master/spec/Schema.spec.js#L260 it is possible to define a `count` CLP rule. But this option is not available to be set through Parse Dashboard.

douglasmuraoka:
Allows users to set/unset CLP to run `count` queries through the `SecurityDialog` component.

Screenshot from 2019-07-23 12-58-08

Fixes #1165

### DIFF
--- a/src/components/PermissionsDialog/PermissionsDialog.react.js
+++ b/src/components/PermissionsDialog/PermissionsDialog.react.js
@@ -24,12 +24,13 @@ import {
 let origin = new Position(0, 0);
 
 function renderAdvancedCheckboxes(rowId, perms, advanced, onChange) {
-  let get = perms.get('get').get(rowId) || perms.get('get').get('*');
-  let find = perms.get('find').get(rowId) || perms.get('find').get('*');
+  const get = perms.get('get').get(rowId) || perms.get('get').get('*');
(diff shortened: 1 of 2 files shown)


alencarlucas:
There is something weird with the state update. If you apply some changes in CLP and open the PermissionsDialog again the shown state is still the previous, but I'm not sure if it's related to this PR as well.

davimacedo:
I've just checked here and it is not related to this PR. I am merging this one and we can tackle the problem on a separate issue.

davimacedo:
Open the issue to fix the dialog state: https://github.com/parse-community/parse-dashboard/issues/1182

johncodeos:
I have Parse Dashboard 2.0.3 and Parse Server 3.9.0 but still don't get the 'count' CLP option. I'm using PostgresDB. Does it work for PostgresDB databases or only for MongoDB? I have a problem with my app saying 'Permission denied for action 'count' on class <Name of my Class>' when I'm trying to count the items inside the class and I found out that I need to enable this option to fix it.

PR #1193, opened 2019-07-30
feat: Add required and default value options

alencarlucas:
This PR means to add the feature developed in [#5835](https://github.com/parse-community/parse-server/pull/5835) to Parse-Dashboard, that includes the option to set required fields and default values to the schema. This feature is available for Parse-Server version >= 3.7.0 and all field types, except Relation.

Some screenshots:

- String
image

- Boolean
image

- Field creating with an invalid default value
image

- Handle error on object creation
image

### DIFF
--- a/src/dashboard/Data/Browser/AddColumnDialog.react.js
+++ b/src/dashboard/Data/Browser/AddColumnDialog.react.js
@@ -115,7 +115,7 @@ export default class AddColumnDialog extends React.Component {
           formattedValue = new Parse.Polygon(JSON.parse(defaultValue))
           break
         case 'GeoPoint':
-          formattedValue = new Parse.GeoPoint(defaultValue)
+          formattedValue = new Parse.GeoPoint(JSON.parse(defaultValue))
           break;
         case 'Pointer':
           formattedValue = await this.handlePointer(defaultValue, target)
--- a/src/dashboard/Data/Browser/AddColumnDialog.react.js
+++ b/src/dashboard/Data/Browser/AddColumnDialog.react.js
@@ -17,8 +17,8 @@ import DateTimeInput      from 'components/DateTimeInput/DateTimeInput.react';
 import SegmentSelect      from 'components/SegmentSelect/SegmentSelect.react';
 import FileInput          from 'components/FileInput/FileInput.react';
 import styles             from 'dashboard/Data/Browser/Browser.scss';
-import Parse from 'parse'
-import validateNumeric from 'lib/validateNumeric';
+import Parse              from 'parse'
+import validateNumeric    from 'lib/validateNumeric';
 import {
   DataTypes,
   SpecialClasses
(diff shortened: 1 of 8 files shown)


acinader:
wow!  I'll leave the review to someone knowledgable, but this is great.  Thank you.

src/dashboard/Data/Browser/AddColumnDialog.react.js
```diff
@@ -74,21 +193,42 @@ export default class AddColumnDialog extends React.Component {
...
+            Allow include require fields and default values if the parse-server
+            version is greater than or equal 3.7.0, that is the minimum version
+            support this feature and check if the field is not a relation
+          */
+          semver.valid(this.props.parseServerVersion) &&
+          semver.gte(this.props.parseServerVersion, '3.7.0') &&
+          this.state.type !== 'Relation' ?
+            <>
+              <Field
+                label={<Label text='What is the default value?' description='If no value is specified for this column, it will be filled with its default value' />}
```
TomWFox:
```suggestion
                label={<Label text='What is the default value?' description='If no value is specified for this column, it will be filled with its default value.' />}
```

src/dashboard/Data/Browser/AddColumnDialog.react.js
```diff
@@ -74,21 +193,42 @@ export default class AddColumnDialog extends React.Component {
...
+                input={this.renderDefaultValueInput()}
+                className={styles.addColumnToggleWrapper} />
+              <Field
+                label={<Label text='Is it a required field?' description={'When true this field must be filled when a new object is created'} />}
```
TomWFox:
```suggestion
                label={<Label text='Is it a required field?' description={'When true this field must be filled when a new object is created.'} />}
```

src/dashboard/Data/Browser/AddColumnDialog.react.js
```diff
@@ -54,11 +92,92 @@ export default class AddColumnDialog extends React.Component {
...
+      case 'GeoPoint':
+        return <TextInput
+          placeholder='Set here a default value'
+          multiline={true}
+          onChange={async (defaultValue) => await this.handleDefaultValueChange(defaultValue)} />
+      case 'Number':
+      case 'String':
+      case 'Pointer':
+        return <TextInput
+          placeholder={type === 'Pointer' ? 'Set here a valid object ID' : 'Set here a default value'}
```
TomWFox:
```suggestion
          placeholder={type === 'Pointer' ? 'Set a valid object ID here' : 'Set a default value here'}
```

alencarlucas: @TomWFox thanks for the review! All suggestions were applied.

dplewis:
I just pull it down and tested it. Can you add a editColumnDialog with required and default value?

davimacedo:
@dplewis it is actually not so easy, because Parse Server does not allow modifying existing fields. We will tackle this issue later, though. But first in Parse Server side, then here. We've discussed some of the next steps in this thread: https://github.com/parse-community/parse-server/pull/5835

I will create a project with each of the tasks.

Do you think that we could go ahead like this and address the modification later? We will also have to decide what to do in the case a not required field becomes a required one...

dplewis:
We can go ahead with this. But I think we should wait until the next server release.
https://github.com/parse-community/parse-server/pull/5868 was just merged.

davimacedo:
I agree. I am planning to release a new version of Parse Server as soon as the new version of @parse/push-adapter gets merged to solve the push bug in the latest version. So this one here will probably not seat for a long while.

davimacedo:
@dplewis I think we can merge, right?

kevflynn:
This is great. It doesn't seem to work for objects as a default value - is that by design? If so, should remove the default value field when specifying an object. 
To try it - create a default value of say {"mon": 0} or {mon: 0} - neither works.

davimacedo:
@kevflynn do you mind to open an issue about this bug you found?

PR #1211, opened 2019-08-09
fix: Breaking changes in react-dnd version 9.x

alencarlucas:
The `DragDropContext` was removed in version 9.0.0 in react-dnd, as we can see [here](https://github.com/react-dnd/react-dnd/releases/tag/v9.0.0).


davimacedo:
Thanks!!!

PR #1219, opened 2019-08-13
 feat(Database Browser): Select all objects

douglasmuraoka:
Allows users to select all visible objects. If lazy loading fetches more data, the select all checkbox is unset.

### DIFF
--- a/src/components/DataBrowserHeaderBar/DataBrowserHeaderBar.react.js
+++ b/src/components/DataBrowserHeaderBar/DataBrowserHeaderBar.react.js
@@ -14,12 +14,12 @@ import { DndProvider }     from 'react-dnd'
 
 export default class DataBrowserHeaderBar extends React.Component {
   render() {
-    let { headers, onResize, selectAll, onAddColumn, updateOrdering, readonly, preventSchemaEdits } = this.props;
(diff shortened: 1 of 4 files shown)


douglasmuraoka:
> Like we talked, I think that a better approach would be selecting what is currently visible.

Changes done!

davimacedo:
@douglasmuraoka it worked pretty good to me but I think we need just one more small fix. If I have, for example, 400 rows, select all (it will select the first 200 ones - because they are the visible ones), then delete these rows, everything works fine and the dashboard count shows me that I have 200 rows left, but they don't show up in the table. I think we should manage to load more rows every time that a delete operation happens. Can you do this?

douglasmuraoka:
> @douglasmuraoka it worked pretty good to me but I think we need just one more small fix. If I have, for example, 400 rows, select all (it will select the first 200 ones - because they are the visible ones), then delete these rows, everything works fine and the dashboard count shows me that I have 200 rows left, but they don't show up in the table. I think we should manage to load more rows every time that a delete operation happens. Can you do this?

Done!
Whenever the user deletes objects and the remaining objects on the table are lesser than the maximum allowed objects, we fetch more data.

PR #1222, opened 2019-08-14
fix(Database Browser): Editing values not possible on mobile

### LINKED ISSUE #953: Parse Dashboard Hover Issue on Mobile Browsers
I can't edit the values on mobile browsers. 
I think it's a hover issue in the dashboard.bundle.js file.

douglasmuraoka:
Fixes #953


davimacedo:
So how does it work? In a mobile device, do I need to touch a first time to select the cell, then a second time to enable editing? It is not working for ACL. Can you please take a look?

dplewis:
I just tried on iOS and it works great.

I was also able to change boolean but since there isn't a tap outside event it could only be changed by selecting a different field.

When I tried to upload a file I got `Unhandled Promise Rejection: Error: Tried to encode an unsaved file` but the file was still uploaded when I refreshed.

davimacedo:
@dplewis did the ACL popup work for you? It was the only thing that didn't work properly for me, but I tested in Google Chrome.

dplewis:
It worked for me on Safari, Chrome and iOS.


IMG_2559TRIM

davimacedo:
Nice. I've just tested again and it worked as well. I think in the first time I haven't noticed the popup was outside the screen I had to scroll in order to see it. This is probably something to be improved later. LGTM!

dplewis:
I had the same issue when I first used it. That’s why I zoomed in on the recording.

Could improve later

PR #1223, opened 2019-08-14
fix(Database Browser): Avoid Parse transformations on array and object fields

### LINKED ISSUE #983: The dashboard does not work well with nested objects

douglasmuraoka:
Fixes #983

When rendering/editing `Array` or `Object` type fields on `Database Browser`, we must avoid any transformations of generic data to the `ParseObject` format. Since these fields are generic, we can't assume it will contain Parse related content.

### DIFF
--- a/src/components/BrowserCell/BrowserCell.react.js
+++ b/src/components/BrowserCell/BrowserCell.react.js
@@ -52,9 +52,7 @@ let BrowserCell = ({ type, value, hidden, width, current, onSelect, onEditChange
     content = dateStringUTC(value);
   } else if (type === 'Boolean') {
     content = value ? 'True' : 'False';
-  } else if (type === 'Array') {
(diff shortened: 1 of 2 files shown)


dplewis:
I’ll pull it down now

PR #1224, opened 2019-08-15
fix(FileEditor): Error: Tried to encode an unsaved file

dplewis:


douglasmuraoka:
I don't think this is the right approach to solve this bug. We'll lose the `Uploading...` component state, and also block the UI until the file upload is done.
From what I've investigated, it is more related to the `obj.toJSON` that I'm using to get the raw object and array data than the file upload itself. I'll try to find some alternatives to resolve this.

dplewis:
Thanks for looking into this. I’ll also try to find an alternative.

dplewis:
Closing via https://github.com/parse-community/parse-dashboard/pull/1225

PR #1225, opened 2019-08-15
fix(Database Browser): JSON conversion breaking file upload process

douglasmuraoka:


douglasmuraoka:
@dplewis could you test this PR and see if the file upload is working?

PR #1233, opened 2019-08-20
Add masterkey parameters

mtrezza:
Adds a property to Parse Config parameter to make it only retrievable with master key.

### DIFF
--- a/src/dashboard/Data/Config/ConfigDialog.react.js
+++ b/src/dashboard/Data/Config/ConfigDialog.react.js
@@ -178,6 +178,7 @@ export default class ConfigDialog extends React.Component {
     this.props.onConfirm({
       name: this.state.name,
       value: GET_VALUE[this.state.type](this.state.value),
+      masterKeyOnly: this.state.masterKeyOnly
     });
   }
 
--- a/src/dashboard/Data/Config/ConfigDialog.react.js
+++ b/src/dashboard/Data/Config/ConfigDialog.react.js
@@ -108,13 +108,15 @@ export default class ConfigDialog extends React.Component {
     this.state = {
       value: null,
       type: 'String',
-      name: ''
+      name: '',
+      masterKeyOnly: false
     };
     if (props.param.length > 0) {
       this.state = {
         name: props.param,
         type: props.type,
         value: props.value,
+        masterKeyOnly: props.masterKeyOnly
       };
     }
   }
@@ -228,6 +230,18 @@ export default class ConfigDialog extends React.Component {
               description='Use this to configure your app. You can change it at any time.' />
           }
           input={EDITORS[this.state.type](this.state.value, (value) => { this.setState({ value }) })} />
+        <Field
+          label={
+            <Label
+              text='Master key only'
+              description='The parameter is returned only when requested with the master key. You can change it at any time.' />
+          }
+          input={
+            <Toggle
+              type={Toggle.Types.TRUE_FALSE}
+              value={this.state.masterKeyOnly}
+              onChange={(masterKeyOnly) => this.setState({ masterKeyOnly })} />
+          } />
       </Modal>
     );
   }
--- a/src/dashboard/Data/Config/ConfigDialog.react.js
+++ b/src/dashboard/Data/Config/ConfigDialog.react.js
@@ -18,6 +18,7 @@ import React           from 'react';
 import TextInput       from 'components/TextInput/TextInput.react';
 import Toggle          from 'components/Toggle/Toggle.react';
 import validateNumeric from 'lib/validateNumeric';
+import styles          from 'dashboard/Data/Browser/Browser.scss';
 
 const PARAM_TYPES = [
   'Boolean',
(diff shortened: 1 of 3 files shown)


mtrezza:
This feature requires https://github.com/parse-community/parse-server/pull/5954.

davimacedo:
@mtrezza nice job! Could you please just centralize the field and set the background color to the same of the other fields? Below how it current is:
Screen Shot 2019-08-20 at 11 12 41 PM

It would be good if it had the same style of the other similar fields. An example below:
Screen Shot 2019-08-20 at 11 14 24 PM

davimacedo:
Also.. in the other places this field is used the labels are `No` and `Yes`

mtrezza:
@davimacedo Done, looks like this now:

Screen Shot 2019-08-21 at 12 15 47

src/dashboard/Data/Config/ConfigDialog.react.js
```diff
@@ -228,6 +232,21 @@ export default class ConfigDialog extends React.Component {
               description='Use this to configure your app. You can change it at any time.' />
           }
           input={EDITORS[this.state.type](this.state.value, (value) => { this.setState({ value }) })} />
+        <Field
+          label={
+            <Label
+              text='Requires master key?'
+              description='The parameter is returned only when requested with the master key. You can change it at any time.' />
```
TomWFox: Probably good to be super explicit...
```suggestion
              description='When true the parameter is returned only when requested with the master key. You can change it at any time.' />
```

mtrezza:
@TomWFox, as @davimacedo suggested, we are going away from the notion of “true/false” in favor of “yes/no”. Do you want to change the wording in your suggestion?

src/dashboard/Data/Config/ConfigDialog.react.js
mtrezza: As @davimacedo suggested, we are going away from the notion of “true/false” in favor of “yes/no”. Do you want to change the wording in your suggestion?

mtrezza: As @davimacedo suggested, we are going away from the notion of “true/false” in favor of “yes/no”. Do you want to change the wording in your suggestion?

mtrezza: As @davimacedo suggested, we are going away from the notion of “true/false” in favor of “yes/no”. Do you want to change the wording in your suggestion?

TomWFox: Sure

TomWFox:
Could you check that the text is displayed correctly with the extra words?

mtrezza:
Actually I am thinking about adding a parse server version check for that new property. So the dashboard is backwards compatible with older parse server versions. @davimacedo which parse server version should I add as the minimum? 3.7.3?

TomWFox:
Lots of changes for the next release so it should be 3.8.0

mtrezza:
> Could you check that the text is displayed correctly with the extra words?

Yes, text fits.

mtrezza:
All changes done, ready for merge I believe.

src/dashboard/Data/Config/ConfigDialog.react.js
```diff
@@ -228,6 +233,30 @@ export default class ConfigDialog extends React.Component {
               description='Use this to configure your app. You can change it at any time.' />
           }
           input={EDITORS[this.state.type](this.state.value, (value) => { this.setState({ value }) })} />
+        
+        {
+          /*
+            Add `Requires master key` field if parse-server version >= 3.8.0,
+            that is the minimum version that supports this feature.
+          */
+          semver.valid(this.props.parseServerVersion) && semver.gte(this.props.parseServerVersion, '3.8.0')
```
davimacedo: We use to check the [server info](https://github.com/parse-community/parse-server/blob/master/src/Routers/FeaturesRouter.js) to decide if a feature is available or not at Parse Dashboard. What do you think?

mtrezza: I took this approach from here:
https://github.com/parse-community/parse-dashboard/blob/169e13b2328fc5adff2afffa2f6e855eb7cc638e/src/dashboard/Data/Browser/AddColumnDialog.react.js#L217-L218

mtrezza: And as I understand it, `app.serverInfo` is set by calling the `/serverInfo` endpoint when the dashboard is launched:
https://github.com/parse-community/parse-dashboard/blob/169e13b2328fc5adff2afffa2f6e855eb7cc638e/src/dashboard/Dashboard.js#L129-L131

davimacedo:
@dplewis do you want to take a look as well?

mtrezza:
@davimacedo This feature has been released with parse server 3.8. Can you create a dashboard release so the feature can be used, since there are also no open dashboard PRs?

davimacedo:
I've just sent the PR for the new version: https://github.com/parse-community/parse-dashboard/pull/1263/files

PR #1241, opened 2019-08-23
refactor(Database Browser): Table performance improvements

douglasmuraoka:
Improves the general performance of the `Browser`component and its nested components, in order to provide a smoother experience.
* Improved table scrolling UX by increasing the number of elements rendered on screen (from 60 to 200) and increasing the number of rows needed to trigger a rerender of the entire table (from 20 to 160). Now users can scroll way more without stuttering (needs more testing on older devices).
* Removed several unnecessary renderings during initialization. `Browser`component initialization has become faster.
* Removed several unnecessary table row and cell renderings whenever a table cell is selected, causing stuttering when holding an arrow key for navigation (instead of updating only the old/new selected cell, it was updating the entire table).


dplewis:
I noticed that sometimes I can’t scroll pass two-three pages if I go really fast. That includes using arrow keys

douglasmuraoka:
@dplewis could you please take a look again?

PR #1258, opened 2019-08-30
fix: Update base config to work with sass-loader 8.x

alencarlucas:
[Source](https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#800-2019-08-29)

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "parse-dashboard": "./bin/parse-dashboard"
   },
   "engines": {
-    "node": ">=4.3"
+    "node": ">=8.9"
   },
   "main": "Parse-Dashboard/app.js",
   "jest": {
--- a/webpack/base.config.js
+++ b/webpack/base.config.js
@@ -53,7 +53,14 @@ module.exports = {
               importLoaders: 2
             },
           },
-          "sass-loader?includePaths[]=" + encodeURIComponent(path.resolve(__dirname, '../src'))
+          {
+            loader: 'sass-loader',
+            options: {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -120,7 +120,7 @@
     "parse-dashboard": "./bin/parse-dashboard"
   },
   "engines": {
-    "node": ">=4.3"
+    "node": ">=8.9"
```
davimacedo: Is this required?

alencarlucas: We have some dependencies that require at least node 8.x, as [Parse-Server](https://github.com/parse-community/parse-server/blob/master/package.json#L109). Besides that, sass-loader announced a breaking change in the [changelog](https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#800-2019-08-29) about minimum required node version.

davimacedo: I think it is pretty reasonable to require at least node >= 8.9 as in Parse Server. @dplewis @acinader thoughts?

acinader: Yes, I'm good with it.  Node 8 is at 8.16 and it goes out of maintenance in a few months at which point I think we set the minimum at 10.

More broadly (and to invite criticism), I think for node version for all concerned repositories (and parse-dashboard in particular), we should not bend over backward to support obsolete versions like early node 8 versions.

Users don't have to upgrade to newer versions of parse if they are unwilling to keep up with dependencies.

davimacedo: I agree. Let's go to Parse Dashboard 2.0 then.

PR #1315, opened 2019-09-30
The objectId was missing from the filter list

### LINKED ISSUE #1297: Filters is missing "objectId" in 2.0.3
Hey, loving the changes in `v2`, very happy to see new work happening on this repo! Just a bug I've noticed in `2.0.3`, the filters dropdown is missing `"objectId"`. This also happens to be the filter we use most often when searching for a specific record - if others agree, is it also worth having a quick input field, perhaps next to the Class title, which expands an input box and let's us `quick find` a record by `"objectId"`?

W0lfw00d:
This should resolve #1297. The objectId was missing from the filter list.

It was not include because of fix #1283 about am issue with the userPointers. That should now still be in place since, but I don't 100% understand that issue.


mrmarcsmith:
@davimacedo Can we make a release for this bug fix? I'm facing this issue on 2.0.3 which is currently the latest release.

Aud123:
Hi.
Same here for me. I am still running version 2.0.3 of the parse-dashboard, and to this day, the objectId is still missing from the filter. Thanks a lot!

j4zz:
I would also appreciate a 2.0.4 for this.

acinader:
working on this now.  should be out shortly.

Issue #1309, opened 2019-09-27
Dashboard duplicates last entry


ramiro-ciocca:
When you have a large collection, say more than a screen, and you order it by ascending "created at", the last entry is shown twice.  It is not duplicated in the database, it is just the dashboard showing it twice.

PR #1334, opened 2019-10-07
Prevent the duplicate last entry when sorting by _createdAt

W0lfw00d:
Only do a complex query for `createdAt` instead of `-createdAt`

This should resolve this issue: https://github.com/parse-community/parse-dashboard/issues/1309


dplewis:
@W0lfw00d It looks good to me. Can you provide details on how to replicate the issue?

W0lfw00d:
Basically have more row than can fit in the screen (so 40 for example) and check the length of `this.state.data`. When you sort by "createdAt" in ascending order and scroll down in the list, it will return 41 items.

PR #1347, opened 2019-10-17
prepare for 2.0.4

acinader:
@dplewis I can also wait for you to fully review #1334 so we can add to this release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 ## Parse Dashboard Changelog
 
 ### master
-[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.3...master)
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.4...master)
 
 * _Contributing to this repo? Add info about your change here to be included in next release_
+### 2.0.4
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.3...2.0.4)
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -89,7 +89,7 @@
     "file-loader": "4.2.0",
     "http-server": "0.11.1",
     "jest": "24.8.0",
-    "marked": "0.6.2",
+    "marked": "^0.7.0",
```
dplewis: Can you pin the packages?

CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 ## Parse Dashboard Changelog
 
 ### master
-[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.3...master)
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.4...master)
 
 * _Contributing to this repo? Add info about your change here to be included in next release_
+### 2.0.4
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.3...2.0.4)
+* Fix: Restore `ObjectId` to filter ([#1315](https://github.com/parse-community/parse-dashboard/pull/1315)), thanks to [Wolfwood](https://github.com/W0lfw00d)
```
dplewis: Add #1334

acinader:
I tried to remove all '^' from package.json, but caused problems with babel.  Doesn't make sense to me, but I don't think we should hold up this release to solve that existing problem.

dplewis:
That’s weird, will have to look into it later. 

Go ahead with the release 👍

Issue #1352, opened 2019-10-22
NPM Install Issue


danielchangsoojones:
I'm getting this error whenever I try to run npm install --save:

`ERROR in ../node_modules/parse/lib/browser/LocalDatastore.js
Module not found: Error: Can't resolve '@babel/runtime-corejs3/core-js-stable/array/from' in '/Users/danieljones/Documents/Btown-Rides/btown-rides-parse-dashboard/node_modules/parse/lib/browser'
 @ ../node_modules/parse/lib/browser/LocalDatastore.js 7:35-94
 @ ../node_modules/parse/lib/browser/Parse.js
 @ ../node_modules/parse/index.js
 @ ./dashboard/Data/Browser/Browser.react.js
 @ ./dashboard/Dashboard.js
 @ `./dashboard/index.js`

My parse dashboard used to work, but now whenever I try to deploy to Heroku or install npm, then it just will keep throwing this error. I know that it's not deploying to Heroku because the npm packages can't install properly. I feel like this is probably a problem on my local machine or somehow my npm is messed up. But, I can't figure it out. I even forked a perfectly clean copy of the parse-dashboard and tried to run npm install --save. I still get the exact same error.

acinader:
Hi @danielchangsoojones three bits of info you need to provide in order to get some help:

1. version of parse-dashboard
2. version of node
3. version of npm

I am using Parse Dashboard 2.0.4 on node v10.14.1 with npm 6.10.0

My guess is that you are trying to use version 2.0.3.  If so, upgrade and try again.

If you still have an issue, go ahead and reopen with the requested details.

stepanic:
@acinader I made the [fork](https://github.com/stepanic/parse-dashboard) today and I have an idea to implement a vertical form for editing rows in the class.

I have followed this instruction:
https://github.com/parse-community/parse-dashboard/blob/master/CONTRIBUTING.md
but I got almost the same errors, like @danielchangsoojones 

1. I tried
```
$ nvm use
Now using node v8.11.4 (npm v5.6.0)

$ npm install
```

Also, I've tried to use `npm install` with the latest stable node
`Now using node v10.17.0 (npm v6.11.3)`
but the same issue exists

`npm install`
STDOUT and STDERR are attached here as `debug.out`
[debug.log](https://github.com/parse-community/parse-dashboard/files/3763854/debug.log)

stepanic:
I can confirm that right now on version 2.0.5 at commit:
[...]

PR #1354, opened 2019-10-23
add @babel/runtime-corejs3

acinader:
get the build working again.

fixes: #1352

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 
 * _Contributing to this repo? Add info about your change here to be included in next release_
 ### 2.0.4
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.4...2.0.5)
+* Fix: Add @babel/runtime-corejs3 to fix broken build ([#1354](https://github.com/parse-community/parse-dashboard/pull/1354)), thanks to [Arthur Cinader](https://github.com/acinader)
+
+### 2.0.4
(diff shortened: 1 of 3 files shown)


acinader:
oops.  need to move @babel/runtime-corejs3 to dev-dependencies.  one min.

acinader:
ok, now I think it's all ready :).

dplewis:
@acinader corejs3 already exists in the package-lock.json file but it is labeled as a dev dependency

https://github.com/parse-community/parse-dashboard/blob/master/package-lock.json#L1061

The JS SDK has it as a main dependency.

https://github.com/parse-community/Parse-SDK-JS/blob/master/package.json#L33

To fix this you should delete the existing package-lock.json file and `npm install` and upload the new one. You should see that it is no longer a dev-dependency in the package-lock.json.

Greenkeeper only updated parse to 2.8.0 but not its dependencies.

dplewis:
@acinader Here is the culprit (From Travis)

Screen Shot 2019-10-24 at 12 41 25 PM

Greenkeeper should have uploaded the correct package-lock file. Can you look into it?

acinader:
It worked for me.    I also updated the changelog to get rid of a duplicate 2.0.5 entry and to give you the well-deserved credit for this.

One thing to note is that when I ran npm install, it changed the package-lock file (which I committed),  I'm assuming that this is because we grab the latest parse-server which may change the dependencies.

If this all looks good to you, just approve and i'll merge and publish.

thanks!!!

PR #1357, opened 2019-10-24
⚡️ Release v2.0.5

dplewis:
Update Package Lock file to handle dependencies (corejs3).

CoreJS is no longer a dev dependency 👇
https://github.com/parse-community/parse-dashboard/compare/2.0.5?expand=1#diff-32607347f8126e6534ebc7ebaec4853dR917


dplewis:
@acinader Can you pull it down and see if it works?

dplewis:
I alway assumed that for parse-server for example although package.json points to the latest. package-lock should point to a specific commit (that commit will get update if you regenerate package-lock.json).

PR #1431, opened 2019-12-21
Context menu with quick filters and links to related records in other Parse Objects 📃

404-html:
Handy context menu which allows:
- set filter;
- add filter;
- get related records from other Object;
- open "Edit row" dialog.

See it in action:

filter5

TODO:
- [x] generate context menu items dynamically with taking data type into account;
- [x] sort objects listed in context menu alphabetically;
- [x] make an exceptions for cells with data that is non-filterable;
- [x] context menu to expand sub items on the other side if there's not enough space;
- [x] add third context menu option. Option should be available only when clicking on `Pointer` type of cell or cell under `objectId` column. Option should say something like: "Get records in relation from...". Selecting this option should expand sub menu with `Objects` that have data related with this specific `id`. Clicking on any object listed should navigate to it with proper filter applied;
- [x] context menu to close after choosing an item (good to be configurable via props);
- [x] context menu to close after clicking outside it;
- [x] move `ContextMenu` item in `pig` list to keep alphabetical order (currently is at the bottom of the list);
- [x] context menu to work when clicking on empty cell.


dplewis:
Thanks for the contribution! I'll pull this down and try it out.

Does this work on mobile?

404-html:
Hey @dplewis,
 > Does this work on mobile?
I doubt, it's still in early stage. I created this MR for someone to check whether I'm not doing something terribly wrong. I should push more in next few days.

404-html:
I've slightly changed the approach and created general `ContextMenu` component instead of keeping it on `Browser` level.
I've also started putting items into TODO in the description.

404-html:
Hey @dplewis,

I believe it's ready to be reviewed. I'm looking forward for any comments.

P.S. it works on mobile too :)

davimacedo:
@404-html would you mind to resolve the conflicts so we can merge it?

davimacedo:
I think it is not passing the lint. Can you please take a look?

404-html:
Sorry @davimacedo, I resolved conflicts using GitHub web UI as I don't know how to pull from master into forked repo, it's I shame.
Anyway I fixed lint errors and I'm open for any comments 👍

404-html:
Small addition: I've added "Edit row" context menu option which allows to quickly access recently introduced, lovely edit row dialog (cheers @NinoZX).

edit-row

davimacedo:
Nice job! Thanks for the PR!

PR #1432, opened 2019-12-24
fix(Browser): Infinite scroll not working

douglasmuraoka:
After #1334, the infinite scroll is no longer loading the next set of objects. This commit undoes #1334, and tries another approach in order to fix the original problem.

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -405,7 +405,7 @@ class Browser extends DashboardView {
     let className = this.props.params.className;
     let source = this.state.relation || className;
     let query = queryFromFilters(source, this.state.filters);
-    if (this.state.ordering !== 'createdAt') {
+    if (this.state.ordering !== '-createdAt') {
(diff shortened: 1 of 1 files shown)


douglasmuraoka:
@W0lfw00d could you please check if your problem from #1334 is still solved?

dplewis:
I think I can replicate #1334 again. I'll try again. Can you explain the logic behind this in the meantime?

douglasmuraoka:
The objects in the dashboard are ordered by `-createdAt` by default. So, in order to fetch the next page, we use the `createdAt` value of the last object fetched so far (which should be lesser than the first object's date) and use it as the starting point for the next query. The change from #1334 was not considering this logic to fetch the next page of objects, thus, the query built to fetch the next objects had incorrect constraints and always resulting in zero objects for the next page.

In addition to that, this logic was not considering that `createdAt` itself could have a different order than descending, because the descending order was always being applied (as you can see at line 435 in the original file) and it was always trying to fetch objects with lesser `createdAt` dates than the last object of current page.

PR #1438, opened 2020-01-04
Taking hidden columns into account when calculating Editor position

### LINKED ISSUE #1437: Cell editor position is not taking hidden columns into account
#### Steps to reproduce

1. Hide some columns using "Manage columns" option in Browser;
2. Double click on a cell in a column that supposed to be on the right side of some hidden column.

Cell editor seems to be taking hidden columns into account.
image

It's easy to run into state when editor is rendered off visible screen.

404-html:
Fixes #1437.


davimacedo:
Good catch! Thanks for the PR.

Issue #1323, opened 2019-10-02
no warning dialog before deleting config parameter


mtrezza:
#### Issue
When deleting a Parse Config parameter by clicking on the trash icon <img width="43" alt="image" src="https://user-images.githubusercontent.com/5673677/66041140-6d2b2d00-e519-11e9-9062-432b75aa30e4.png"> the parameter is deleted without showing a warning dialog.

As it can happen that the trash icon is clicked by mistake, one wouldn't easily see which parameter has been deleted in a long list of parameters. There is also no built-in way to easily restore a key and value that has been deleted by mistake.

#### Suggested solution
Add a warning dialog for the user to confirm the deletion. Similar to the warning dialog when deleting a row.

<img width="554" alt="image" src="https://user-images.githubusercontent.com/5673677/66041327-e591ee00-e519-11e9-9825-727241328e5c.png">

The warning dialog should at least include the parameter's key name.

#### Environment Setup
- Parse Dashboard 2.0.3

#### Steps to reproduce
--

#### Logs/Trace
--

davimacedo:
I like the idea. Would you be willed to add this one?

mtrezza:
At the moment not, but maybe someone wants to pick it up.

mtrezza:
reopen, PR in the works

mtrezza:
@davimacedo can you reopen please, PR in the works

davimacedo:
Sure!

mtrezza:
PR merged

mtrezza:
@davimacedo When can we expect the next release? I would like to see this feature added.

davimacedo:
Go for it @TomWFox . We have a lot of good stuff here ready to be released.

PR #1443, opened 2020-01-13
added config param delete confirmation dialog

mtrezza:
Issue https://github.com/parse-community/parse-dashboard/issues/1323

Still needs to be tested, getting only blank screen for all dialogs in dashboard for some reason.


mtrezza:
It's working now.

The config parameter confirmation dialog requires to enter the parameter name.

image

PR #1451, opened 2020-01-17
Fix audience conditions when Installation class does not exist

davimacedo:
If you don't have created the Installation class yet and try to add a condition to an Audience, the Parse Dashboard crashes. This PR properly handle the problem and show an error message instead of crashing.


src/components/PushAudienceDialog/PushAudienceDialog.react.js
```diff
@@ -258,9 +263,9 @@ export default class PushAudienceDialog extends React.Component {
         </div>
         {futureUseSegment}
         <FormNote
-          show={Boolean(this.props.errorMessage && this.props.errorMessage.length > 0)}
+          show={Boolean(this.props.errorMessage && this.props.errorMessage.length > 0 || this.state.errorMessage && this.state.errorMessage.length > 0)}
```
acinader: i'm not 100% sure it matters from operator precedence standpoint, but it would be easier to read if you put the two conditionals in parens:

```
(this.props.errorMessage && this.props.errorMessage.length > 0) || (this.state.errorMessage && this.state.errorMessage.length > 0)
```

src/components/PushAudienceDialog/PushAudienceDialog.react.js
```diff
@@ -91,6 +92,10 @@ export default class PushAudienceDialog extends React.Component {
   }
 
   handleAddCondition() {
+    if (!this.props.schema || !Object.keys(this.props.schema).length) {
```
acinader: NO NEED TO CHANGE, but whenever I see something like this, I change it to:

```!(this.props.schema && Object.keys(this.props.schema).length)```

it's easier for ME to read (ymmv)

PR #1453, opened 2020-01-19
Added support for reverse proxy

githubsaturn:
Currently the app.js assumes that all "non-127.0.0.1" requests are insecure if they are not HTTPS. 

However, this is an incorrect assumption as the app could be behind a reverse proxy. This is specially important when you use Docker to deploy an nginx that manages HTTPS and routes the request to Parse. In that case, the remote IP is in private range of 172.17.0.1 or 192.168.0.1.

### DIFF
--- a/Parse-Dashboard/app.js
+++ b/Parse-Dashboard/app.js
@@ -88,8 +88,10 @@ module.exports = function(config, options) {
         req.connection.remoteAddress === '127.0.0.1' ||
         req.connection.remoteAddress === '::ffff:127.0.0.1' ||
         req.connection.remoteAddress === '::1';
+      // To support reverse-proxy:
+      const isForwardedFromHttps = req.headers['x-forwarded-proto'] === 'https';
       if (!options.dev && !requestIsLocal) {
-        if (!req.secure && !options.allowInsecureHTTP) {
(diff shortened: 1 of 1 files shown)


davimacedo:
In this case, you just need to set `allowInsecureHTTP` option true, right?

githubsaturn:
Setting `allowInsecureHTTP` tapes over the problem as a temporary fix. But it's not a true fix.

If user sets  `allowInsecureHTTP`, the would allow real plain text HTTP connections to go through as well which is a security risk. 

We don't want to allow:
User =====HTTP====> nginx ========HTTP=====>Parse

But we want to allow:
User =====HTTPS====> nginx ========HTTP=====>Parse

davimacedo:
In this case, you just need to setup NGINX to not allow HTTP requests. Or, even better, to redirect all HTTP requests to HTTPS.

githubsaturn:
Then what's the point of `allowInsecureHTTP` at all? You might as well remove the `allowInsecureHTTP` flag and default to `allowInsecureHTTP=TRUE` for everything.

davimacedo:
Someone might not want to use it below a reverse proxy and require the HTTPS connection.

PR #1464, opened 2020-02-01
Playground

MutterPedro:
Added the playground feature, inspired on the DB workbenches console. This is a feature that I miss so much when I am using Parse Dashboard. Often I need to make custom queries that is not implemented in my project and the ordinary filters can't provide to me. I was planning to do it for a long time, but just now I had a little bit of time. Accepting new suggestions and improvements. Hope you like it!

Screenshot from 2020-02-01 16-38-45
Screenshot from 2020-02-01 16-38-34
Screenshot from 2020-02-01 16-38-21
Screenshot from 2020-02-01 16-37-40


davimacedo:
I like the idea! Maybe we should improve little bit the design. But it is actually a very good start as is. Can you please check the integration errors out? @dplewis @acinader thoughts?

acinader:
just looking at the pics here, but l love the idea.

dplewis:
I like it, will definitely pull this down.

MutterPedro:
Hello guys, I am really happy you like the idea! I am open to ideas and suggestions, mainly on design. I am trying to see whats the problem in the CI, but It seems that just stuck on the `npm ci` step. It happened just when I merged the latest changes from the master... :thinking:

MutterPedro:
I updated the dependencies with `npm audit fix` and fixed the Travis CI problem :tada: :raised_hands: @davimacedo

Parse-Dashboard/parse-dashboard-config.json
```diff
@@ -1,12 +0,0 @@
-{
```
davimacedo: Can you please revert this change?

src/components/BrowserFilter/BrowserFilter.react.js
```diff
@@ -34,7 +34,7 @@ export default class BrowserFilter extends React.Component {
     this.node = ReactDOM.findDOMNode(this);
   }
 
-  componentWillReceiveProps(props) {
+  UNSAFE_componentWillReceiveProps(props) {
```
davimacedo: I'd either not change this now or use a safe method. Same thing for the other files. What do you think?

src/components/CodeEditor/CodeEditor.example.js
```diff
@@ -0,0 +1,20 @@
+/*
```
davimacedo: Do we need this file?

davimacedo:
In addition to the code review, I pulled and did some tests. I actually didn't understand very well how the output works. With one-line find statements it works pretty well, but I didn't understand the behavior in the case we have multiple lines and I'd like to log something or even save and return an object. It is also very hard to understand the errors since there is no error output. Maybe would be better just output the the execution logs instead of the object tree? Maybe both of them?

In terms of design:
- Do you think there is any chance to have an auto-complete?
- What do you think of a dark mode?
- I've also thought about this feature in the past but the way I imagined it would be in a collapsible window fixed in the bottom of the page (like most of the IDEs) in such a way that you can navigate through the pages without closing the playground. It would be very nice being able to see the data and work with the SDK at the same time. I am not sure if my idea it is clear to you. What do you think about it?

Parse-Dashboard/parse-dashboard-config.json
MutterPedro: Sure!

src/components/CodeEditor/CodeEditor.example.js
MutterPedro: Do you mean why we need to separate the CodeEditor as a component?

src/components/BrowserFilter/BrowserFilter.react.js
MutterPedro: I agree it appears when I ran a command suggested by a warning in the console by React, saying that `componentWillReceiveProps` is deprecated. I will change to the safe way

MutterPedro:
Hey @davimacedo, thanks for your review and feedback. It is the first version, so yes we can improve it much more. I was inspired by the **consoles** on the DBs workbench, where you run the queries and see the results. So, to emulate it, I make the output be the last statement, I hardcoded I return in the last statement before evaluating the code. If you write a multiline script, it won't work unless you separate the statements with a **semicolon**. That is the workaround that I needed to do by now to make it work with multi statements and returning just the last one. But I like the idea to print the `console.something()` statements too, or doing something like the [runkit](https://npm.runkit.com/) does.

About the design bullets:
- I will try to find a component with an auto-complete feature for **Javascript**.
- I guess that the component I used has a theming property, so we can make a toggle to switch the mode, what you think?
- I understand what you mean, but it would be new to Parse Dashboard, there is nothing with this design there yet, do you think it would fit the current design? My first think about the output was to show the result in a table, like in the browser tab. Like it would be a custom filter.

src/components/BrowserFilter/BrowserFilter.react.js
MutterPedro:
```
backend.js:6 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move code with side effects to componentDidMount, and set initial state in the constructor.
* Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: SideEffect(NullComponent), SidebarHeader
```

MutterPedro:
I am intending to change the code editor component to one of these:

- https://www.npmjs.com/package/react-ace (have basic autocomplete and themes)
- https://www.npmjs.com/package/react-monaco-editor (supported by Microsoft)

src/components/CodeEditor/CodeEditor.example.js
davimacedo: Yes. I didn't understand really well the need of this additional component so I decided to ask you and confirm. But, if it is something necessary, nevermind.

davimacedo:
I think it is for sure a good start that can be improved later. I've just tried to add some ideas but feel free to only implement what you believe is necessary/easy now. In terms of the output, why don't we just execute the code and print out the console logs + the object that it was explicitly returned using `return` keyword (I'd not try to add it automatically in the last line).

src/components/CodeEditor/CodeEditor.example.js
MutterPedro: I did it just to make a separation of concerns, in case I needed to change the code editor component dependency, I wouldn't need to change the playground itself, also I thought that was cool to let other people use it for other cases...  But it is not necessary, If you think it is better to use it outside a component, I can remove it.

davimacedo: I think it is good. Thanks for the explanation.

MutterPedro:
I thought about it... But you don't feel strange to put a return in a thing that you don't know is a function? I agree with you that printing the console would be great. I will change the code editor component to support basic autocomplete and dark mode, together with it I implement the console.

davimacedo:
yes... it is kinda strange... maybe we should just print out the logs. Then each one can choose the best way to use. Maybe I'd do something like this:
```js
console.log(main());

function main {
  ...
  return 'Hello World!'
}
```

MutterPedro:
I will try this approach to see if we like it

MutterPedro:
Hello guys! Here I am again. I made some improvements on the component, moved the editor to [React Ace Editor](https://www.npmjs.com/package/react-ace), now it haves a simple autocomplete, dark mode, snippets, and the results are only shown when `console.log` or `console.error` is used. I really liked this version, I guess the UX is much more intuitive and the UI is much more pleasant. What do you think? @davimacedo @acinader @dplewis 

Here are some print screens:
Captura de tela de 2020-02-15 15-57-03
Captura de tela de 2020-02-15 15-57-19
Captura de tela de 2020-02-15 15-57-34
Captura de tela de 2020-02-15 15-58-21
Captura de tela de 2020-02-15 15-58-26

Parse-Dashboard/parse-dashboard-config.sample.json
```diff
@@ -0,0 +1,14 @@
+{
```
davimacedo: Se don't need this file, right?

src/components/BrowserFilter/BrowserFilter.react.js
```diff
@@ -5,20 +5,20 @@
...
-import Filter        from 'components/Filter/Filter.react';
-import FilterRow     from 'components/BrowserFilter/FilterRow.react';
-import Icon          from 'components/Icon/Icon.react';
-import Popover       from 'components/Popover/Popover.react';
-import Position      from 'lib/Position';
-import React         from 'react';
-import ReactDOM      from 'react-dom';
-import styles        from 'components/BrowserFilter/BrowserFilter.scss';
-import { List, Map } from 'immutable';
+import * as Filters from "lib/Filters";
```
davimacedo: Is it possible to avoid the styles changes like this one? I think it would be better to continue with the current pattern.

davimacedo:
Nice job! It is very good now. What do you think about moving it inside API Console menu? Maybe it could be called "JS Console"

Parse-Dashboard/parse-dashboard-config.sample.json
MutterPedro: Yeah, sorry about that... It was just to me not to lose the original reference

src/components/BrowserFilter/BrowserFilter.react.js
MutterPedro: It came from the merge... I didn't change this file.

MutterPedro:
Yeah, I guess that would be the right place. Also, I need to update the docs to explain how the playground works, can I do it in this PR?

MutterPedro:
Made the requested changes :+1:

davimacedo:
I see you are also Brazilian, so my feedback will go in Portuguese: "Está animal!!!" :)

Two things more and we can merge:

1 - What do you think about changing the default example to be something like this?

```js
const myObj = new Parse.Object('MyClass');
myObj.set('myField', 'Hello World!')
await myObj.save();
console.log(myObj);
```

2 - I think your IDE is set to use " instead of ' and it replaced all existing " to ' in the files that you changed. Could you revert this?

MutterPedro:
Sim @davimacedo , sou brasileiro também, fico feliz que gostou da minha contribuição hehe :brazil: :tada: 

1. I changed the **JS Console** placeholder to your suggested code.
2. I added a rule on **eslint config** to force the **single quote**. It was my IDE indeed that was doing that reformatting.
3. After adding that rule, I ran `npm run lint -- --fix`. That's why there are many files changed, but at least they are all using the single quote  :sweat_smile:

okoumea:
Hello,
Please how to remove the placeholder ?
I have to manually delete each line every time i want to use the JS Console.

PR #1466, opened 2020-02-02
feat(query): set read preference to SECONDARY by default

omairvaiyani:
*Use-case*

We think it makes sense to keep the load off the primary where possible, but here is one scenario that has stung us many times:

Given how flexible the data-browser is, it's very easy to accidentally set the `sort` on a large collection on an unindexed field. Whilst there are multiple lines of defences that we can add (beforeFind in parse-server, no unindexed queries in MongoDB, etc), we think this is a good starting point for other developers.

*Implementation*
Right now, I've just set the common find and count queries to be `SECONDARY`. None of the critical flows such as writes, or even reads on Roles, ACLs etc are set to use this preference. For now, I've not added a configuration option to disable this. Happy to look into this if it's helpful?

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -341,6 +341,7 @@ class Browser extends DashboardView {
       query.ascending(field)
     }
 
+    query.readPreference(this.getReadPreference());
     query.limit(MAX_ROWS_FETCHED);
 
     let promise = query.find({ useMasterKey: true });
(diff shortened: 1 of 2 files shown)


davimacedo:
It makes a lot of sense for me but I believe it should be an option that the developer should opt-in maybe per app. I'd also set it to SECONDARY_PREFERRED (and not SECONDARY) for the case the database is not running on a replica set or for any reason the secondary is out.

mtrezza:
This is a very delicate topic. The default read preference of Parse Server should be what the default of the database is, which is `PRIMARY`. I am not aware of a database cluster that would not make `PRIMARY` the default, because anything else is a special mode that requires many considerations.

Issue #1391, opened 2019-11-25
Feature Request for Granular Pointer Permissions


pitxels:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [ √] You're running version >=1.0.23 of Parse Dashboard.

- [ √] You're running version >=2.3.2 of Parse Server.

- [√ ] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

When creating a pointer permission in the dashboard it would be great to be able to select get, find, create, update, delete separately as the other permissions, that could avoid the need of lots of coding.

pitxels:
Also the "Require authentication" option would be nice yo have in this modal as well.

PR #1478, opened 2020-02-27
Add ProtectedFields dialog and enhance Permissions dialogs

BufferUnderflower:
- Adds Protected Fields dialog
image

clp-protected-fields

- Added an Authenticated row - represents `requiresAuthentication` from schema.
- Allowed adding Array field as PP.
- Fixes incorrect behavior when it was possible to add Relation field as PP (not supported server-side).
- Allows to set Pointer Permissions (PP) per-operation in advanced mode. Previously it was only possible to set write/read groups ( `readUserFields` / `writeUserFields` in schema).  Upon saving the field name will go under corresponding `operation` into `pointerFields` array.   If all (read or write) operations  are checked  - saves into corresponding `readUserFields`/`writeUserFields` array instead. 
parse-community/parse-server#6351 parse-community/parse-server#6352 #1391

image

- Now saving CLP from dashboard does not clear the `protectedFields`.


davimacedo:
Great job and thanks for the PR! It is a huge update for Parse Dashboard.

I've just had a try. The CLP worked perfectly but the Protected Fields didn't work for me. Nothing happens when I click on "All field allowed. Click to protect.".

Can you please take a look?

BufferUnderflower:
> Great job and thanks for the PR! It is a huge update for Parse Dashboard.
> 
> I've just had a try. The CLP worked perfectly but the Protected Fields didn't work for me. Nothing happens when I click on "All field allowed. Click to protect.".
> 
> Can you please take a look?

I can't reproduce it here unfortunately (safari/chrome/firefox) on macos.
Could it be so that you've tried a class without any custom fields and it is just empty? Default fields are not shown.
Meanwhile i'll handle such case and show a message saying there are no fields that can be protected.

BufferUnderflower:
Also added a scroll hint

scroll-hint

davimacedo:
That was the problem. I hadn't created any field. Nice job!

PR #1494, opened 2020-03-12
CSV Export of class data

Vortec4800:
Here's a first stab at adding a CSV Exporter function to the data browser. I've mirrored the basic functionality of the existing delete function, if you have rows selected you can export just those rows, or there is a separate menu item to export the entire class data set.

Currently, the "all" mode limits the query to 10,000 rows. I'm not sure if we want to change that, or if someone has a better solution, but I'm worried about trying to execute an export on a class with 10 million rows in it. At least now you'll get 10k as a result which is large but manageable.

Anyway I'm much better at general Node development than I am at React development, so apologies upfront for any stupid mistakes. Let me know if you have any feedback, or input on what we want to do with this. Speaking from personal experience, being able to quickly export data into a CSV file would save me a ton of time so I thought it would be a useful addition.

Closes https://github.com/parse-community/parse-dashboard/issues/1756.

TODO:
- [x] Add changelog entry
- [x] Add feature description to README
- [x] Fix export bug of empty pointers
- [x] Fix export empty fields as empty CSV fields
- [x] Fix export object field values
- [x] Fix export array field values
- [x] Fix export ACL values

Potential future improvements:
- Make export limitation of 10k records overridable with custom value via dashboard config


src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,62 @@ class Browser extends DashboardView {
...
+      rowsToExport: null,
+    });
+  }
+
+  async confirmExportSelectedRows(rows) {
+    this.setState({ rowsToExport: null });
+    const className = this.props.params.className;
+    const query = new Parse.Query(className);
+
+    console.log(rows);
```
404-html:
```suggestion
```

404-html:
Nice! It would be good to take hidden columns into account (skip them in CSV file). Or make that configurable with a checkbox in export dialog?

Vortec4800:
I can look into that, do you know how I can tell if a column is hidden? I also need to figure out how to add a checkbox to the react dialog - I basically just re-used the delete dialog which doesn't have one.

404-html:
>  you know how I can tell if a column is hidden?
I did that fix recently, maybe it will give you some clue:
https://github.com/parse-community/parse-dashboard/pull/1438/files

Vortec4800:
Hmm, I tried filtering on `column.visible` but that doesn't seem to work. That property doesn't appear to be set on columns that are fetched directly from the schema.

Vortec4800:
I haven't been able to figure out how to hide the hidden columns, how do you all feel about merging this as-is and adding the hidden column feature in the future?

Vortec4800:
@404-html I just noticed this hasn't been merged yet. Are we good to do that? I did merge your PR into mine, so this one has your changes as well here.

404-html:
Hey @Vortec4800, I'll try to look into that over the weekend. I can give an approval but I'm afraid I don't have write access to merge. Anyway I can CC someone who have.

P.S. is it worth to add somewhere in the UI an info about 10,000 rows limit?

JacobWithACapitalJ:
any update on the status of this PR? It would be an extremely useful feature to have

Vortec4800:
I'm hoping we can still get someone with write access to review and merge this.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,73 @@ class Browser extends DashboardView {
     });
   }
 
+  showExportSelectedRowsDialog(rows) {
+    this.setState({
+      rowsToExport: rows,
```
404-html, line 842:
```suggestion
      rowsToExport: rows
```

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,73 @@ class Browser extends DashboardView {
...
+    });
+  }
+
+  cancelExportSelectedRows() {
+    this.setState({
+      rowsToExport: null,
```
404-html, line 848:
```suggestion
      rowsToExport: null
```

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,73 @@ class Browser extends DashboardView {
...
+        objectIds.push(objectId);
+      }
+      query.containedIn('objectId', objectIds);
+    }
+
+    const classColumns = this.getClassColumns(className, false);
+    // create object with classColumns as property keys needed for ColumnPreferences.getOrder function
+    const columnsObject = {};
+    classColumns.forEach((column) => {
+      columnsObject[column.name] = column
```
404-html, line 873:
```suggestion
      columnsObject[column.name] = column;
```

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,73 @@ class Browser extends DashboardView {
...
+      csvString += row + '\n';
+    }
+
+    // Deliver to browser to download file
+    const element = document.createElement('a');
+    const file = new Blob([csvString], { type: 'text/csv' });
+    element.href = URL.createObjectURL(file);
+    element.download = `${className}.csv`;
+    document.body.appendChild(element); // Required for this to work in FireFox
+    element.click();
```
404-html, line 904: Should this `a` element be removed now?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -0,0 +1,41 @@
...
+        icon='warn-outline'
+        title={this.props.selection['*'] ? 'Export all rows?' : (selectionLength === 1 ? 'Export this row?' : `Export ${selectionLength} rows?`)}
+        subtitle={''}
+        disabled={!this.valid()}
+        confirmText={'Yes export'}
+        cancelText={'Never mind, don\u2019t.'}
+        onCancel={this.props.onCancel}
+        onConfirm={this.props.onConfirm}>
+        {}
+      </Modal>
```
404-html, line 38: Can we add to this modal an info about 10,000 rows limit?

404-html (commented):
Hey @Vortec4800, I've added few comments to the code.

src/dashboard/Data/Browser/Browser.react.js
Vortec4800: I guess I don't understand. Why would we remove it?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
Vortec4800: Sure, let me see how I can work that in.

Vortec4800:
Does this work, or did you have something else in mind?

Screen Shot 2020-09-12 at 7 34 47 PM

src/dashboard/Data/Browser/Browser.react.js
404-html: Just to clear things up. After couple of exports we might end up with few, unnecessary DOM elements.

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
404-html: The one you posted in the main thread looks perfect 👏👏👏👏

Vortec4800:
@404-html Any thoughts here on my dialog solution?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
Vortec4800: Sorry I didn't see your response until just now. Oops!

I'll commit this now.

src/dashboard/Data/Browser/Browser.react.js
Vortec4800: I see what you mean. Okay I'll remove after the click.

Vortec4800:
Okay we should be good now. Let me know if there is anything else.

404-html:
Happy Sunday @Vortec4800 and thanks for the changes. I played with the build a but and I noticed one problem. 
Export doesn't cope well with rows that have empty pointer. Here's an example:

image
While the one marked with green exports fine the one with red results in following exception thrown:

```
Uncaught (in promise) TypeError: Cannot read property 'id' of undefined
    at dashboard.bundle.js:95
    at Array.map (<anonymous>)
    at h (dashboard.bundle.js:95)
    at t.<anonymous> (dashboard.bundle.js:95)
    at u (dashboard.bundle.js:53)
    at Generator._invoke (dashboard.bundle.js:53)
    at Generator.forEach.e.<computed> [as next] (dashboard.bundle.js:53)
    at n (dashboard.bundle.js:1)
    at s (dashboard.bundle.js:1)
```

To reproduce simply create an object with a pointer reference and wipe it with delete key (while cell is selected in dashboard). Do you think you can get that fixed?

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,74 @@ class Browser extends DashboardView {
...
+  }
+
+  async confirmExportSelectedRows(rows) {
+    this.setState({ rowsToExport: null });
+    const className = this.props.params.className;
+    const query = new Parse.Query(className);
+
+    if (rows['*']) {
+      // Export all
+      query.limit(10000);
```
dplewis, line 859: I think it would be a good idea to make this configurable (config.json) and default it to 1000 rows. If the database isn't indexed properly this could lead to problems.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -831,6 +837,74 @@ class Browser extends DashboardView {
...
+    const objects = await query.find({ useMasterKey: true });
+    let csvString = columns.map(column => column.name).join(',') + '\n';
+    for (const object of objects) {
+      const row = columns.map(column => {
+        const type = columnsObject[column.name].type;
+        if (column.name === 'objectId') {
+          return object.id;
+        } else if (type === 'Relation' || type === 'Pointer') {
+          return object.get(column.name).id;
+        } else {
```
j-heinze, line 891:
```suggestion
        } else if (type === 'Relation' || type === 'Pointer') {
          if (object.get(column.name)) {
          return  object.get(column.name).id
          } else {
          return 'undefined'
          }; 
        } else {
```
This should fix the issue with the empty pointers.

404-html: Hey @Vortec4800, is this suggestion okay?

Vortec4800: Sorry I haven't had a lot of time to dedicate to this lately, but I do at the moment so let me run through these suggestions and changes and try to get them all merged together.

mtrezza:
@Vortec4800 Would you be available to get this PR ready for merge? There may be convenience functionality we can lave to future PRs, but it would be nice to have this export feature in the next release of the dashboard.

Vortec4800:
I agree - I'd love to get this merged. Sorry, I've been trying to balance bandwidth lately and this has been an open (but neglected) tab in my browser for the better part of 6 months.

The only remaining item is the 10,000 column limit in the config.json. How do we feel about merging it as-is and adding that in a future update?

I've resolved any merge conflicts and I think otherwise, it's ready to be merged after review.

src/dashboard/Data/Browser/Browser.react.js
mtrezza: Good idea, we'll make a note about this as potential future improvement.

mtrezza:
> The only remaining item is the 10,000 column limit in the config.json. How do we feel about merging it as-is and adding that in a future update?

Sure, we can leave this as potential future improvement.

I've opened https://github.com/parse-community/parse-dashboard/issues/1756 to have a corresponding issue for this PR, for future high level discussions.

I have also added a TODO list to your first comment here. If you can add a short feature description to the README (really short, just take a look at the existing one), then you can mention the warning / limitation, that export is limited to 10k records, and any other limitations you have in mind. You don't need to add possible feature extensions though (such as export format JSON in addition to CSV), because that is always a WIP.

Vortec4800:
Sure thing. Will work on that now.

mtrezza:
Another caveat we may need to mention, how does the export handle delimiter and quotes within the data?

I notice that you are not using another dependency to export the data, so this is hand made and we'd need to manually account for these special cases.

Vortec4800:
Okay, let me try testing with those and I'll see if I can come up with a solution for them.

mtrezza:
I am fine with also leaving this for future improvement, but this could limit the practical usability of this feature drastically.

I think the usual convention is:

- a) no delimiter or quote (easy case)

  | a   | b   | c     |
  |-----|-----|-------|
  | one | two | three |

  exports as:
  
  ```
  a, b, c
  one, two, three
  ```

- b) contains delimiter

  | a   | b   | c     |
  |-----|-----|-------|
  | one | two,two | three |
  
  exports as:
  
  ```
  a, b, c
  one, "two, two", three
  ```

- c) contains quotes

  | a   | b   | c     |
  |-----|-----|-------|
  | one | "two,two" | three |
  
  exports as:
  
  ```
  a, b, c
  one, \""two, two"\", three
  ```

Vortec4800:
Okay so I checked the code, it's been a while since I wrote it...

Currently everything exports with quotes around each field, so delimiters within the field shouldn't cause a problem. However, a quote in the field would because nothing is escaping.

I think fixing this wouldn't be too difficult, just check the data for all three of your posted scenarios. Alternatively, I can likely include the export package, but I didn't know how everyone felt about adding more dependencies. Let me know and I can implement either one.

Just need to get the build working again, currently complaining about `Module not found: Error: Can't resolve 'react-popper-tooltip'`

Vortec4800:
Okay, take a look at what I just committed, that should address all the different combinations you outlined above, including combining quote and delimiter in the same field.

mtrezza:
Thanks, looks good, and you got the double quote correct (I'll correct my comment), according to FC-4180:

> If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.

README.md
```diff
@@ -570,6 +570,12 @@ This feature allows you to use the data browser as another user, respecting that
 
 > ⚠️ Logging in as another user will trigger the same Cloud Triggers as if the user logged in themselves using any other login method. Logging in as another user requires to enter that user's password.
 
+## CSV Export
+
+This feature will take either selected rows or all rows of an individual class and saves them to a CSV file, which is then downloaded. CSV headers are added to the top of the file matching the property names.
```
mtrezza, line 575: > matching the column names.

To call them by the same name through-out the dashboard.

src/dashboard/Data/Browser/BrowserToolbar.react.js
```diff
@@ -138,6 +139,16 @@ let BrowserToolbar = ({
           onClick={onCloneSelectedRows}
         />
         <Separator />
+        <MenuItem
+          disabled={!selectionLength}
+          text={`Export ${selectionLength <= 1 ? 'this row' : 'these rows'} to CSV`}
```
mtrezza, line 144: I think this great feature merits its own menu item! Instead of placing it under `Edit`, how about creating a new `Export` item and moving the two sub-items there?

I would add it between `Browse` and `Refresh` and we should have an existing icon with an arrow pointing downwards.

mtrezza (changes requested):

mtrezza:
Not sure whether this is a bug or something to note as improvement for a future PR, but when I export a field that contains an object like `ACL` or `authData`, it is exported with value `[object Object]`.

Vortec4800:
Yeah I wasn't sure how to deal with JSON objects, seemed a bit much to have them in a CSV file, but just showing [object Object] isn't very useful either.

The undefined string came from @j-heinze in his fix for the crash and I left it. I don't feel strongly about whether it should say "undefined" or just be an empty string, so whichever everyone wants it would be easy to swap that in. Would definitely be smaller with empty strings, like you say.

src/dashboard/Data/Browser/BrowserToolbar.react.js
Vortec4800: Got it, will post this in the main conversation area for review.

Vortec4800:
Finished moving the export menu, this will be better if we do end up adding JSON export in the future. You'll definitely want to review my code here, it looks correct to me but I'm a bit out of my element on this part...

Screen Shot 2021-08-16 at 7 30 38 PM

mtrezza:
The menu looks great, is there a comment where the crash issue when exporting empty fields is described?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -0,0 +1,41 @@
...
+    return true;
+  }
+
+  render() {
+    let selectionLength = Object.keys(this.props.selection).length;
+    return (
+      <Modal
+        type={Modal.Types.INFO}
+        icon='warn-outline'
+        title={this.props.selection['*'] ? 'Export all rows?' : (selectionLength === 1 ? 'Export this row?' : `Export ${selectionLength} rows?`)}
```
mtrezza, line 30: Can you change this to `Export selected row` instead of `Export this row`? Looking at the menu, it is not clear what `this / these` is referring to.

mtrezza (changes requested):

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
mtrezza: done

mtrezza:
@Vortec4800 I've made a few changes:

- Fix export empty fields as empty CSV fields
- Fix export object field values
- Fix export array field values
- Fix export ACL value
- Minor rewording in menu and dialog

Could you take another quick look over it, to see if you'd agree with the changes and whether there is anything you'd like to add before merging?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
Vortec4800: I agree with the change, though I pulled the wording from the other options in the Edit menu. Perhaps this is a change that should be made (in another PR) for the other options in Edit?

Vortec4800:
Everything looks good to me - I'm happy if you guys are. Excited to see this merged!

mtrezza:
@404-html any thoughts before we merge?

404-html:
Thank you for asking @mtrezza, I'm happy for this to be merged.
Finally I can delete my custom script for CSV export. But first I need to find it, which is always a struggle!
Great job gents, thank you.

mtrezza (approved):
Looks good to me! Thanks for this essential feature, I'm sure you'll be causing a few tears of joy among developers with this.

tehsunnliu:
Wow merged just when I needed it. 
Thank you all!!!!!!

PR #1516, opened 2020-04-08
Release 2.1.0

TomWFox:
I don't think there are any breaking changes in this one...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,26 @@
 ## Parse Dashboard Changelog
 
 ### master
-[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.0.4...master)
+[Full Changelog](https://github.com/parse-community/parse-dashboard/compare/2.1.0...master)
 
-* Feature: added JS Console feature to run custom queries
-* Feature: added save button on JS Console feature to save the custom query on the local storage
+### 2.1.0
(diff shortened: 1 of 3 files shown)


TomWFox:
I love the additions in #1478 from @BufferUnderflower but I am a bit concerned about it being included in this release.

@BufferUnderflower did express in https://github.com/parse-community/docs/pull/700#issuecomment-593099908 some concerns about the production readiness of protected fields, hence we decided to delay documentation of this feature and discussed the need for warnings about it.

So I'm wondering whether heavily promoting the feature in the dashboard is responsible? (to be clear I am just going off what @BufferUnderflower has said - I don't know the details of the implementation of protected fields)

TomWFox:
@davimacedo @dplewis @acinader thoughts on this when you have a chance 🙂

davimacedo:
I believe we should release it. Protected fields are part of Parse Server master for a while now. It may have its limitations but it is something that we need to discuss and tackle one by one separately.

TomWFox:
@davimacedo thanks for your thoughts. Do you think we should also publish the documentation in the guides?

I suppose my thoughts are that if it is only in the API ref then perhaps the people who come across the feature may more fully evaluate whether it provides adequate security for their needs. Whereas perhaps if people find it in the guides and dashboard they may use the feature without knowledge of its limitations.

TomWFox:
Resolved package-lock conflict, hopefully I haven't broken anything 🙃

PR #1556, opened 2020-06-23
Fix pointer CLPs

mess-lelouch:
As discussed here https://community.parseplatform.org/t/permission-denied-for-action-find-on-class/365/9, it seems that the latest version of the dashboard is not setting pointer CLPs correctly. The details can be found in the thread.

This change fixes it. For instance, if I set the following CLPs

image

I get the following CLP object

```
  "classLevelPermissions": {
    "find": {
      "pointerFields": [
        "user",
        "users"
      ]
    },
    "count": {
      "pointerFields": [
        "user",
        "users"
      ]
    },
    "get": {
      "pointerFields": [
        "user",
        "users"
      ]
    },
    "create": {},
    "update": {},
    "delete": {},
    "addField": {},
    "protectedFields": {
      "*": []
    }
  }
```

I believe that the most important change in behavior that this PR fixes is that now if I `find` as an user (with session token) with exactly the same CLP settings shown in the screenshot above (with public and authenticated permissions turned off), it returns me 0 rows if there are none, 1 row if there are 1 associated row to the user, etc. As you can read from the forum thread referenced above, by not setting the pointer CLPs correctly I could not execute the `find` action as an user.

Something weird I found along the way is that when I added the minor change in line 982, it added the pointer information like this:

image

But it also  had the effect of grouping **Write and Add fields**... let me know if it should be like that or if it should be changed. This behavior is new to me.

image

Please feel free to share any questions or concerns.

### DIFF
--- a/src/components/PermissionsDialog/PermissionsDialog.react.js
+++ b/src/components/PermissionsDialog/PermissionsDialog.react.js
@@ -534,7 +534,7 @@ export default class PermissionsDialog extends React.Component {
       // hide suggestions to avoid ugly footer overlap
       this.refEntry.current.setHidden(hidden);
       // also show indicator when input is not visible
-      this.refScrollHint.current.toggleActive(hidden);
+      // this.refScrollHint.current.toggleActive(hidden);
     };
 
     this.observer = new IntersectionObserver(callback, {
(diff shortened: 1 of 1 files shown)


src/components/PermissionsDialog/PermissionsDialog.react.js
```diff
@@ -534,7 +534,7 @@ export default class PermissionsDialog extends React.Component {
       // hide suggestions to avoid ugly footer overlap
       this.refEntry.current.setHidden(hidden);
       // also show indicator when input is not visible
-      this.refScrollHint.current.toggleActive(hidden);
+      // this.refScrollHint.current.toggleActive(hidden);
```
mess-lelouch, line 537: This line was throwing an error... the variable does not even exists... so I commented it out since I don't have enough context to fix it.

mess-lelouch:
@davimacedo @BufferUnderflower  I apologize in advance for the tagging. I have this PR which I think is kind of important because it fixes the CLP pointer feature. Please feel free to let me know what next steps should I take to get this PR reviewed and possibly merged.

mess-lelouch:
@davimacedo no need to apologize. I will be around to monitor if there are any unexpected issues caused by my changes.

PR #1586, opened 2020-09-01
Fixes EditRowDialog

NinoZX:
- update fields automatically when they are updated on server
- don't update array, object and polygon when no changes are in the field
- GeoPointEditor component added componentWillReceiveProps for values change
- TextInput component textarea mode added rows props to be used


NinoZX:
@mtrezza here is the new PR request at this [point](https://github.com/parse-community/parse-dashboard/pull/1523/files/5377b432ba4c9072fcceef1ffae9a87bb4ef7945#r454745921)

dplewis:
I just tried this out and it looks good. Can you rebase from master?

> update fields automatically when they are updated on server

I didn't see this happen. Can you explain how to test this?

dplewis (approved):
I tried it out and works perfectly. Thanks for the PR.

PR #1592, opened 2020-09-07
fix:(README) GraphQL ServerURL env variable

TheTyrius:
Hi :)

I believe there is a wrong environment variable used in an example in the README that prevented the GraphQL API console to work for me. Please kindly have a look.

Best regards, Philipp


dplewis (approved):
Nice catch!

PR #1613, opened 2020-11-01
Update getPath on Cloud Code

dblythy:
Hi all,

This PR is in relation to a PR I have made on the [Parse Server repo](https://github.com/parse-community/parse-server/pull/6983) that allows cloud files to be viewed in the dashboard.

Cheers


pmmlo:
@dblythy What would be the behavior for a distributed deployments?  If you have multiple parse-server replicas, wouldn't you simply be fragmenting that single node, unless you have a shared mount volume or universal storage where you are storing Cloud Code?

I think the option to view Parse Cloud Functions is fine, but I think updating Parse Cloud Code should be handled by the end user.  Otherwise, you would be limiting the flexibility of how Parse Server can be deployed.  What do you think?

dblythy:
Good points again @pmmlo, thanks again for reviewing the code and providing insight. You are correct - the editing functionality wouldn't work for distributed deployments. But, I don't think that should mean that the feature shouldn't be available for single instances, where the feature would work as expected.

Maybe we could add a warning, explaining that editing the code only works on single deployments? Or, have an option in Parse Server such as `singleInstance: true`, and the dashboard is a code viewer if it's multi-instance, but editor if it's single instance.

If the consensus is to leave it as cloud code viewing only, I'm happy to revert it to that!

dblythy:
@Moumouls made similar points [here](https://github.com/parse-community/parse-server/pull/6983#pullrequestreview-552676095)

PR #1624, opened 2020-11-17
Add MFA to Dashboard

### LINKED ISSUE #856: Improving security using 2FA
Hi team,
Have any of you implemented Two factor authentication (2FA) for Parse dashboard, it's very crucial that we have better security implemented for the dashboard.

Would love to hear any security augmentations you've made for your parse-dashboard.

dblythy:
Hi all, 

I'm currently working on adding MFA to Parse Dashboard.

How to enable MFA:

1. Create an MFA token for the user using `parse-dashboard  --generateMFA username`
2. Set the MFA secret to the user json.
3. Open the MFA URL using Google Authenticator

Now, when you login, you can use the OTP from Google Auth to validate.

Example:

```
"users": [
    {
      "user":"user1",
      "pass":"pass",
      "mfa":"NBMCGFTWFBHUCE2G",
    }
],
```

Now, if you open [this URL](//otpauth://totp/user1:dashboard?secret=NBMCGFTWFBHUCE2G&period=30&digits=6&algorithm=SHA1) on your mobile with Google Authenticator, it will add the OTP code, specific to the secret `NBMCGFTWFBHUCE2G`.

This PR is functional, the only problem I'm having is I can't get `parse-dashboard  --generateMFA username` to work. Could anyone point me in the right direction?

Thanks 😊 

Closes #856


mtrezza:
Related: https://github.com/parse-community/parse-server/pull/7079

mtrezza:
I think this feature is simply too good to wait for the Auth adapter of Parse Server to be merged. I also assumed it would be merged faster, but unfortunately is has become stale for now.

@dblythy Would you rebase this on master and resolve any conflicts to we can add this great feature?

dblythy:
I have evolved this to `node ./Parse-Dashboard/index.js --createUser`, to allow users to generate bcrypt and MFA secrets for their dashboard. Here's what it looks like:

Screen Shot 2021-09-04 at 3 36 15 pm

dblythy:
Ready for review @mtrezza 😊

src/login/Login.js
```diff
@@ -41,6 +41,9 @@ export default class Login extends React.Component {
         <LoginRow
           label='Password'
           input={<input name='password' type='password' />} />
+        <LoginRow
```
mtrezza, line 44: Is this text field always displayed? That may be confusing for users who do not need to enter an OTP. Normally OTP is done in a multi-step process and only displayed if the user has OTP enabled. To keep this PR simple, maybe the OTP field could need to be unfolded or at least at a text above it "if you use OTP, enter it".

mtrezza (changes requested):

dblythy:
Resolved @mtrezza! The OTP field now hides unless a login fails with an OTP error.

mtrezza (changes requested):
Can you please remove the RAS syndrome through-out the PR?

"OTP code" is like "MFA authentication", I suggest to simply spell it out, because in addition, not everyone may know what OTP refers to in this case. For example, instead of:

>  Please enter your OTP code.

let's write:

>  Please enter your one-time password.

mtrezza:
I am slightly concerned about otplib, because it is not maintained since 2 years, however, according to snyk it has no vulnerabilities. Normally I would argue against adding an unmaintained package, but in this case I think we can just use it and switch to anther one if needed.

image

mtrezza:
@dblythy Could you try to delete package-lock.js and node_module and run `npm i` to see whether it builds for you? It doesn't for me, but that may be another issue.

dblythy:
> I am slightly concerned about otplib,

Do you have any recommendations for other libraries? Probably a good suggestion.

I also had no issues before but now getting:

```
While resolving: parse-dashboard@2.2.0
npm ERR! Found: graphql@14.6.0
npm ERR! node_modules/graphql
npm ERR!   graphql@"14.6.0" from the root project
```

mtrezza:
> Do you have any recommendations for other libraries? Probably a good suggestion.

The other ones I found have the same issue; I guess it's ok if we just go with this one.

> I also had no issues before but now getting:

Which node and npm version?

dblythy:
I am on Node v14.16.0. Same issue on master. I'll try on my other computer

mtrezza:
Interesting, I tried on this PR with node 12 and 14 with npm 6 and it installs fine, but fails when executing `webpack --config webpack/publish.config.js --progress`. On master all works fine.

dblythy:
Hmmm. I'm pretty sure my desktop is on `npm 7`, due to issues installing the Dashboard

mtrezza:
That's why. npm 7 is currently not supported by the graphQL dependency it seems, see https://github.com/parse-community/parse-dashboard/issues/1639

dblythy:
I'm trying to downgrade to npm 6 but can't get it going. I'll add you to my fork so you can make changes

mtrezza:
Sure, thanks, I hope you can wake up to good news and I can get this working :)

mtrezza:
I took package-lock.json from master, used it in the branch and installed the 3 new dependencies of this PR. This way the dashboard builds without errors for me locally. npm 6, Node 12, 14.

dblythy:
Thank you @mtrezza, appreciate it!!

mtrezza:
Oh, I just noticed that this is actually a double PR, because it also includes a CLI option to create a password hash - neat! 🙌

We now have a feature section in the readme, do you want to add some lines there? We're starting already to collect material for the future dashboard docs.

dblythy:
> Oh, I just noticed that this is actually a double PR, because it also includes a CLI option to create a password hash - neat! 🙌

Thank you! Yep, I also added auto-generated secure passwords. Inquire is actually an awesome package and helps making life so much easier.

Change to readme done!

Parse-Dashboard/Authentication.js
```diff
@@ -90,6 +101,14 @@ function authenticate(userToTest, usernameOnly) {
     //the provided auth matches one of the users
     this.validUsers.find(user => {
       let isAuthenticated = false;
+      if (user.mfa && !usernameOnly) {
+        if (!userToTest.otpCode) {
+          otpMissing = true;
+        }
+        if (!authenticator.verify({ token:userToTest.otpCode, secret: user.mfa })) {
+          otpValid = false;
```
pmmlo, line 109: @dblythy Really great you kept up with this!

Just reading through (not tested), I believe you made a common logical error.  If there are multi-users with multiple OTP-enabled, the first user will work, but I believe any subsequent user matches will fail.  3 quick fixes come to mind:

1) Put an `else` statement after line 110, with `otpValid = true;`.  (Least change option).

2) Default value otpValid as false on line 95: `let otpValid = false;` and make line 108 condition: `authenticator.verify({ token:userToTest.otpCode, secret: user.mfa })` (remove the `!`).

3) Put an additional condition that matches the username before running through the OTP validation (can be most efficient option).


Also, not super ideal using sh1 for the totp, but that's totally okay with good security practices.

Thanks again for contributing!

dblythy: Nice pickup! My bad, thank you!

mtrezza:
I will try the feature out and review the security aspects of it later today. I'm not familiar with what the current best practice for OTP is in terms of hashing algorithm, authenticator app compatibility and HOTP vs. TOTP. Since Parse Dashboard doesn't have a limit on password retries (unlike Parse Server), it is more susceptible to brute force attacks, so I'll take this opportunity to see what OTP configuration can compensate best for that.

mtrezza:
I notice that the algorithm `SHA256` is defined in two places. What do you think about letting the user set the algorithm via dashboard config, and we only default it to `SHA256`? That could make this feature more sustainable and if a user want to use a higher algorithm for better security, the can do so.

dblythy:
That sounds like a good idea! Should it be mfaOptions.algorithm?

mtrezza:
Yep, sounds good to me!

dblythy:
I have moved the CLI to it's own file, so that if any other helpers are made in the future, they can be made very simply without bloating `index.js`

mtrezza:
I looked into TOTP app compatibility.

Auth apps are expected to follow:

- [RFC 6238 - TOTP Algorithm](https://tools.ietf.org/html/rfc6238)
- [RFC 4226 HMAC based OTP Algo](https://tools.ietf.org/html/rfc4226)

The standard only seems to specify SHA-1 as algorithm.

We can see in [this blog](https://labanskoller.se/blog/2019/07/11/many-common-mobile-authenticator-apps-accept-qr-codes-for-modes-they-dont-support/) post that:

- an auth app's compatibility can even vary by platform
- many auth apps accept a OTP of an algorithm they don't support without showing an error, but producing unusable OTPs

### Conclusion

Even though some apps may support higher algorithms, the common denominator seems to be SHA-1 with a 30 seconds period.

I suggest to keep that as a default setting in the dashboard to not cause confusion, for people, especially since auth apps often don't show errors when scanning codes for unsupported algorithms.

pmmlo:
> ### TL;DR
> 
> Let's use SHA-1 algorithm with 30 seconds period as default.
> ### Why
> 
> I looked into TOTP app compatibility.
> 
> Auth apps are expected to follow:
> 
>     * [RFC 6238 - TOTP Algorithm](https://tools.ietf.org/html/rfc6238)
> 
>     * [RFC 4226 - HMAC based OTP Algo](https://tools.ietf.org/html/rfc4226)
> 
> 
> These standards only seems to specify SHA-1 as algorithm.
> 
> We can see in [this blog](https://labanskoller.se/blog/2019/07/11/many-common-mobile-authenticator-apps-accept-qr-codes-for-modes-they-dont-support/) post (from 2019) that:
> 
>     * all apps are compatible with SHA-1, 30 seconds period
> 
>     * some apps accept higher algorithms
> 
>     * an auth app's compatibility from the same vendor can even vary by platform
> 
>     * many auth apps accept to add a code of an algorithm they don't support, without showing an error but producing unusable OTPs
> 
> 
> ### Conclusion
> 
> Even though some apps may support higher algorithms, the common denominator seems to be:
> 
>     * SHA-1 algorithm
> 
>     * 30 seconds period
> 
> 
> I suggest to keep that as a default setting in the dashboard to not cause confusion for people who simply want to set up OTP without experimenting around with algorithms, especially since auth apps often don't show errors when adding codes for unsupported algorithms.
> 
> If someone wants to experiment, they can trial / error with their favorite auth app and set the algorithm in the dashboard options.

This is completely fine with me.  I really don't want to open up a can of worms on this other than to say for secure implementations, it is recommended to use a higher block size for preimage attacks.  There's a number of reasons why 2011's RFC6238 standard "defaults" to sha1 beyond better security/practice.  Conversation is commonly centered around other aspects.

Sha1 for TOTP is not broken, so I don't mind.  Worst case, one of us simply pushes a PR for review with a config in future.

Thanks for being thorough.

dblythy:
> There's a number of reasons why 2011's RFC6238 standard "defaults" to sha1 beyond better security/practice

We can always add a note to recommend higher algorithms, but that higher algorithms mightn't be supported in all auth apps. In this PR, developers can pick any algorithm that is supported in [nodejs](https://nodejs.org/api/crypto.html#crypto_crypto_createhmac_algorithm_key_options)

I've also added customisation here for the amount of digits and the period of the TOTP, this should give flexibility entirely to the developer.

mtrezza:
Sounds good. Adding OTP to the dashboard is already an improvement to the status quo, and we want to make it easy for developers who want an additional layer of security without having to investigate TOTP algorithms. Unless we can determine that most auth apps today support an algorithm higher than SHA-1, we stick with it as default.

This is really a consequence of a missing updated RFC standard and the lack of transparency to know which auth apps supports which algorithms. It's really an industry problem at the moment.

mtrezza:
I did a minor refactoring to remove duplicated code and make the log output more readable

mtrezza:
I think there is flaw in the flow. If I auto-generate a password, end encrypt it, I can't know it.

pmmlo:
> I think there is a tiny flaw in the flow. If I auto-generate a password, and encrypt it, I can't know it.

Try pasting to your config, password manager, or wherever you can securely read it.  I think @dblythy made a very smart choice to keep it off the command line history.

mtrezza:
It seems it is not copied to the clipboard; only the dashboard config is copied automatically. I'll change that.

I'll also change the instructions, because they say to share the QR code or URL with another user. Since they contain the OTP secret, the should never be shared beyond being displayed on the screen.

pmmlo:
> It seems it is not copied to the clipboard; only the dashboard config is copied automatically. I'll change that.
> 
> I'll also change the instructions, because they say to share the QR code or URL with another user. Since they contain the OTP secret, the should never be shared beyond being displayed on the screen.

I would double check before you change because it works exactly as your 1) describes for me.

mtrezza:
The (1) works, but it copies the dashboard config to the clipboard. The issue is that an auto-generated *and* encrypted password is never revealed to anyone and not copied to the clipboard.

pmmlo:
Oh, you erased the step in your refactor.  Look at line 144 of refactoring mfa flow.  1a17255d6f5b44e5b5988b065981b2a958cf08ab

mtrezza:
I don't think so, because `displayResult` is intended to contain the dashboard settings, not the password, that's why it is stringified in `copy(JSON.stringify(displayResult));`.

mtrezza:
Would it make more sense to copy only the password to the clipboard in stead of a JSON obj with the username? This way I can paste the pw directly into the pw manager, instead of pasting it somewhere else and copying on the pw.

dblythy:
> Would it make more sense to copy only the password to the clipboard

I think it makes more sense. You are correct.

pmmlo:
> Would it make more sense to copy only the password to the clipboard in stead of a JSON obj with the username? This way I can paste the pw directly into the pw manager, instead of pasting it somewhere else and copying only the pw.

I thought about that.  I think it's a good idea.

I tried a few flows and what was comfortable for me was pasting the object in a secure note, copying/pasting the password into a manager, and copying the OTP URI into the authenticator (no scanning because I always have to minimize the QR code in my command line).

pmmlo:
> Oh I understand now, you are right. I will fix that.

FWIW, I personally did think `displayResult` and `result` could be named better, but overall, I thought `createUser()` was very readable.

mtrezza:
Yes, I renamed it now, everything is collected in a `data` object and whatever is needed can be taken from there.

dblythy:
Thanks for your help on this @mtrezza. It’s looking pretty good. 

And @pmmlo thanks for your comments and feedback! You are correct, I was conscious to keep the raw password out of command line history.

mtrezza:
So this is the new flow:

image

pmmlo:
> Thanks for your help on this @mtrezza. It’s looking pretty good.
> 
> And @pmmlo thanks for your comments and feedback! You are correct, I was conscious to keep the raw password out of command line history.

Definitely my favorite contribution from you!  Really awesome you were willing to dive into the community.  I unfortunately have to be careful with code contributions because of legal conflicts.  Wish I could have helped.  Do you work for back4app or someone similar?

@mtrezza Thanks for keeping this alive and reviewing.

dblythy:
> Definitely my favorite contribution from you! 

Agree! MFA should welcome some extra security to the Dashboard which will line up nicely with V5 of Parse Server. 

> I unfortunately have to be careful with code contributions because of legal conflicts.

Fair enough, your comments and reviews are always valued so that’s good enough for me!!

> Do you work for back4app or someone similar?

Nope! I am just a keen member of the Parse Community (plus I like solving problems that have been stale for years).

mtrezza (approved):

mtrezza:
I edited the readme a bit, let me know if you want to add anything, otherwise I think we're good to go!

mtrezza:
So there is actually an issue, when I try to open the link on mobile safari, it says invalid link, it doesn't open any app.

Is that an issue or is the URI supposed to be used differently? Usually websites only display the secret to manually copy/paste to an auth app.

luxhealth:
> I have never seen any 2FA website providing a URL as alternative to the QR code. How would the user know what to do with the URL?

I think there's usually some prompt saying, if they can't scan, press "this button" (which copies the URI) and paste into your authenticator.

mtrezza:
Well, Microsoft Authenticator and Google Authenticator, don't accept the URL, but 1pw does. Maybe we just display both.

pmmlo:
> Well, Microsoft Authenticator and Google Authenticator, don't accept the URL, but 1pw does. Maybe we just display both.

Interesting...DUO and Authy accept the URI.  I think technically the standard is to provide a button to copy the URI.  I trust your judgment on this.  A quick fix if we get it wrong.

b38334d-b87e583-QR_for_2FA

mtrezza:
I think technically what you are saying makes sense, the URL provides more info for the auth app. But in practically, auth apps are designed to be easy, so they just take the secret.

How about this?

image

mtrezza (approved):
Looks fantastic to me!

Great work @dblythy, this is truly a milestone for Parse Dashboard security - one of the most important security features added in recent times. And not only that, it's also done elegantly with a guided CLI that makes configuring it a piece of 🍰! I'll prepare a release soon to make this feature available. Definitely a feature we should mention in our first newsletter!

dblythy:
Thanks again for your help @mtrezza!!

Klaitos:
Hello guys, this PR break the docker image : 

`SyntaxError: Unexpected token {
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/src/node_modules/inquirer/lib/utils/screen-manager.js:10:13)`

Klaitos:
Building the image docker with node 12 instead of 8 is fixing the issue. One the dependencies included in this PR need node more than 8

funkenstrahlen:
@dblythy A huge thank you for implementing this! ❤️ Works flawlessly and provides a much higher data security.

dblythy:
@funkenstrahlen i’m glad you like it! Let me know if you have any other ideas for heightened security for the dashboard 🎉

parseplatformorg:
🎉 This change has been released in version [3.3.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [3.3.0](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0)

PR #1677, opened 2021-04-06
Update components to the latest version.

romanbsd:
This updates the dependencies to the latest versions and fixes few bugs and warnings.
I switched from npm to yarn since npm was not able to resolve dependencies.
It needs testing in the graphql area.


.github/workflows/release.yml
```diff
@@ -7,18 +7,18 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
-          node-version: 12
+          node-version: '12'
```
davimacedo, line 12: Does it mean it will only work on node version 12 now on? If not, maybe it would be better to test both versions.

davimacedo: Should we update the package.json as well for the minimum supported version now? @dplewis @mtrezza thoughts?

davimacedo (commented):
Thanks for the PR! Great job!

.github/workflows/release.yml
mtrezza: What are the consequences of switching to yarn? I think we'd want to keep the CI of all repos as uniform as possible. Not only when thinking about mono-repo, but also release automation.

romanbsd: I just changed integer to string as per specification, and in the ci task I changed 10 to 12 so it's consistent in both files. There's nothing that I know of that requires node 12 to be honest. I think it will work fine with 10. So - for your consideration.
Regarding yarn - while it's faster and more capable, the reason I switched to yarn is that npm just failed to resolve all the dependencies for some reason. After struggling with it I gave up and switched to yarn.

.github/workflows/ci.yml
```diff
@@ -13,17 +13,17 @@ jobs:
...
     - name: Cache Node.js modules
       uses: actions/cache@v2
       with:
         path: ~/.npm
-        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
         restore-keys: |
             ${{ runner.os }}-node-
-    - run: npm ci
+    - run: yarn install
```
mtrezza, line 26: What are the consequences of switching to yarn? I think we'd want to keep the CI of all repos as uniform as possible and stick to npm. Not only when thinking about mono-repo, but also release automation. Switching the package manger because a dependency cannot be resolved sounds like buying a new car because a light bulb needs to be changed. This is something we should discuss further.

mtrezza (changes requested):

.github/workflows/ci.yml
mtrezza: > the reason I switched to yarn is that npm just failed to resolve all the dependencies for some reason. After struggling with it I gave up and switched to yarn.

@romanbsd Thanks for clarifying. We are moving towards release automation, which will be focused on npm, so I suggest we stick with npm for now and try to resolve the issue. Could you please give more details about what failed?

.github/workflows/release.yml
mtrezza: Thanks I actually opened another [thread](https://github.com/parse-community/parse-dashboard/pull/1677#discussion_r608163725) to discuss yarn, to keep this thread about Node 10.

.github/workflows/ci.yml
romanbsd: You can run `npm i` on this branch and see for yourself, the errors are pretty similar between different npm versions. I think it boils down to npm being unable to install version which has a conflicting dependency. I gave it some time, but couldn't fix it without resorting to creating and manually modifying npm-shrinkwrap.json manually.

.github/workflows/release.yml
romanbsd: In the current codebase, package.json, ci and release workflows all have different versions specified. Decide upon some version (I think 10 is ok) and I'll change it in all three files.

.github/workflows/ci.yml
mtrezza: Interesting, are you talking about the current master before this PR?

romanbsd: Actually, if I'm not mistaken, I even had problem on master. But here I refer specifically to my branch and my package.json

mtrezza: I assume you are talking about the graphql dependency error? Maybe you can revert this PR back to npm and we look at the errors here?

romanbsd: `npm i` fails, so it's impossible to generate the package.lock ...

romanbsd: interestingly enough, I just tried `npm i` and it worked. Switching back to npm.

.github/workflows/release.yml
romanbsd: I reverted all my changes to these files.

package.json
```diff
@@ -34,78 +34,83 @@
     "README.md",
     "LICENSE"
   ],
+  "resolutions": {
```
mtrezza, line 37: I this is not needed anymore

package.json
```diff
@@ -34,78 +34,83 @@
...
-    "react-dnd-html5-backend": "10.0.2",
-    "react-dom": "16.14.0",
-    "react-helmet": "6.0.0",
-    "react-json-view": "1.21.3",
-    "react-redux": "5.1.2",
-    "react-router": "5.1.2",
-    "react-router-dom": "5.1.2",
-    "regenerator-runtime": "0.13.5",
-    "semver": "7.3.4"
+    "@babel/runtime": "^7.13.10",
```
mtrezza, line 41: Can you please remove all the range operators such as `^` in package.json?

src/components/Loader/Loader.example.js
```diff
@@ -5,7 +5,8 @@
  * This source code is licensed under the license found in the LICENSE file in
  * the root directory of this source tree.
  */
-import { center } from 'stylesheets/base.scss';
```
mtrezza, line 8: Can you revert these imports throughout the project back into 1 line?

mtrezza (changes requested):
Good to hear it works with npm and we don't need yarn after all. When you do the changes, especially removing the range operators, you may want to do that in batches and run `npm i` in between, because that may be what has caused the install to fail previously.

package.json
romanbsd: Why? It's package-lock.json for recording the exact versions.

mtrezza: Because there are automated dependency checks that run on `package`, not `package-lock`.

src/components/Loader/Loader.example.js
romanbsd: No, I believe it's no longer supported in css-loader (or sass-loader?).

mtrezza: Oh, alright then.

mtrezza (commented):
It is somewhat surprising that despite the update of so many dependencies, sometimes by several major versions, virtually no code changes were necessary.

Parse Dashboard does not seem to be well tested, so the CI will easily pass, but this seems suspicious to me. I think it would be good to test this thoroughly by hand before merging.

mtrezza:
> I switched from npm to yarn since npm was not able to resolve dependencies

@romanbsd Did you use npm 7 when you had this issue, and now it is working because you use npm 6?

mtrezza:
What should we do with this PR? I'm concerned about the many major version upgrades in the dependencies without any significant code changes in dashboard. It's possible that we are breaking something and don't notice it, unless we go through the dashboard and test some functionality I guess.

romanbsd:
It's up to you what to do about this PR. For me it was the first step before trying to modernize the project and bring the components to the latest recommended practices. It was working for me (tm), but this definitely needs more testing.

mtrezza:
It this PR in response to https://github.com/parse-community/parse-dashboard/issues/1679?

romanbsd:
This PR is a preparation for https://github.com/parse-community/parse-dashboard/pull/1761

mtrezza:
Then how is this PR related to https://github.com/parse-community/parse-dashboard/issues/1679?

romanbsd:
Besides upgrading the components, it also has one bugfix:
https://github.com/parse-community/parse-dashboard/pull/1677/files#diff-f4ba6f461663e0326e0cd8e081894cc4b977bacd715975017d1c09a41c76fa02L11

romanbsd:
It's related, as it's a first step for 1679. But this PR can be reviewed in isolation. It's not related per se, and was opened before.

mtrezza:
> Besides upgrading the components, it also has one bugfix

Could you open a separate issue and PR for the bugfix?

mtrezza:
I suppose we could proceed this way:
1. merge master into this and resolve conflicts
2. test out the dashboard thoroughly, since there are many major version upgrades and we don't have proper tests

PR #1638, opened 2021-01-07
FIX: Date pickups not working on filters

davimacedo:

PR #1682, opened 2021-04-16
Fix date filters

sadakchap:
- applied changes of this [pr](https://github.com/parse-community/parse-dashboard/pull/1638)
- catching onClick event, to stop propagating to parent


davimacedo (approved):
Thanks for the PR! LGTM!

PR #1688, opened 2021-04-20
Added tooltip for readonly fields

sadakchap:
Added a tooltip for readOnly fields.
Showing tooltip on doubleClick event.
Screenshot from 2021-04-21 16-03-06

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -111,15 +111,6 @@
         "apollo-env": "^0.6.6"
       }
     },
-    "@apollographql/graphql-playground-html": {
-      "version": "1.6.26",
-      "resolved": "https://registry.npmjs.org/@apollographql/graphql-playground-html/-/graphql-playground-html-1.6.26.tgz",
-      "integrity": "sha512-XAwXOIab51QyhBxnxySdK3nuMEUohhDsHQ5Rbco/V1vjlP75zZ0ZLHD9dTpXTN8uxKxopb2lUvJTq+M4g2Q0HQ==",
-      "dev": true,
-      "requires": {
-        "xss": "^1.0.6"
-      }
-    },
     "@apollographql/graphql-upload-8-fork": {
       "version": "8.1.3",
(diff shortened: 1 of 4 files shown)


davimacedo (approved):
LGTM!

davimacedo (changes requested):
Actually, the tests are failing. Could you please check it out?

sadakchap:
Yes, `npm run ci` was failing as I did added a dependency but did not committed `package-lock.json` file. 
I just added it now, please have a look.

davimacedo (approved):
LGTM!

mtrezza:
This pop-up and selection restriction is a rather unintuitive UI behavior. I am not aware of any spreadsheet or table UI element that works that way.

If a cell is read only, then the text field can be set as disabled, which makes it selectable but not editable. It seems quite intrusive to prevent the user from selecting the date string in that field. It also prevents a user from double clicking into a date field and selecting only a part of the date. 

Should we revert this pop-up?

image

sadakchap:
> This pop-up and selection restriction is a rather unintuitive UI behavior. 

Agree.

> If a cell is read only, then the text field can be set as disabled, which makes it selectable but not editable. It seems quite intrusive to prevent the user from selecting the date string in that field because it also prevents from selecting and copying only a part of the date string or scrolling the text left / right if the column is narrow.

For copying date string, user can still do `ctrl+c` after selecting the field.


> Should we revert this pop-up and instead simply make the field non-editable?

Yes, we can do that. But, then how would the user know why Editor is disabled as we will not be explicitly telling the user that this field is actually read-only ? Do you believe that this will be intuitive?

mtrezza:
> Yes, we can do that. But, then how would the user know why Editor is disabled as we will not be explicitly telling the user that this field is actually read-only ? Do you believe that this will be intuitive?

Common UI design practice is to slightly gray out the field to indicate that it cannot be edited but is still selectable. Here is an example from the [Google Material Design](https://material.io/components/text-fields#filled-text-field) framework:
image

mtrezza:
I have opened https://github.com/parse-community/parse-dashboard/issues/1772 to revert this.

PR #1690, opened 2021-04-28
added cancel button for new row

sadakchap:
Added a cancel button for aborting add new row operation
Screenshot from 2021-04-28 18-37-56

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -122,6 +122,7 @@ class Browser extends DashboardView {
     this.handleShowAcl = this.handleShowAcl.bind(this);
     this.onDialogToggle = this.onDialogToggle.bind(this);
     this.abortAddRow = this.abortAddRow.bind(this);
+    this.saveNewRow = this.saveNewRow.bind(this);
   }
 
   componentWillMount() {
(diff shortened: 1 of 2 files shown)


davimacedo (approved):
I like the idea. There is currently no way to cancel. The user would need to add and then delete. And it becomes a bigger problem when you have required fields, since they need to type in all fields (for later delete). @mtrezza @dplewis thoughts?

mtrezza:
Editing a field currently uses key binding to cancel, there is no button, but pressing ESC cancels editing without saving.

Do we need a button for could we simply bind the ESC key to it? Not adding a button would keep the UI clean.

davimacedo:
I'd go with both.

mtrezza:
Can we then just remove the double empty line somehow, now that the button is creating an additional empty line?

image

sadakchap:
sure, we can remove that extra line. now it looks something like this.

Screenshot from 2021-04-29 16-32-55

mtrezza:
Look good!

mtrezza:
Sounds good!

sadakchap:
Hey, @mtrezza I updated the pr, please let me know if we need more changes.

>I just realized that (ESC) key binding is missing generally throughout the dashboard, even in places where it may be intuitively expected, such as to close a modal dialog. Just pointing this out for a potential separate "key binding" PR.

should the ESC key close any opened dialog ?

mtrezza:
> should the ESC key close any opened dialog ?

I just mentioned this as a general remark because I noticed it, but it's not really part of to this PR.

sadakchap:
I tried implenting suggested changes, now it looks something like this...
- Added a new "Add" button to save new row

Screenshot from 2021-05-05 18-31-28

@davimacedo @mtrezza thoughts ?

mtrezza:
Yes, according to UI principles the order should be `Add`, `Cancel` (left to right). Could you change that? Other than that its looks good to me.

sadakchap:
Sure, now it looks like this
Screenshot from 2021-05-05 20-43-07

mtrezza:
Looks good. Are there any responsive UI considerations for these buttons?I think there has been some effort to make the dashboard work on mobile.

sadakchap:
@mtrezza , I think the buttons will look little smaller but same, as these buttons are using existing Button component

mtrezza (approved):
Looks good! Thanks for this PR!

davimacedo:
@sadakchap could you please resolve the conflicts?

sadakchap:
> @sadakchap could you please resolve the conflicts?

Sure @davimacedo.

Klaitos:
This PR is so great ! Thanks for that. But we can't copy cell value once we entered in add new row mode :/
On MacOS Cmd+C doest not copy anymore the cell value

davimacedo:
@Klaitos would you mind to open a new issue?

Klaitos:
I've done the pr https://github.com/parse-community/parse-dashboard/pull/1714

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -303,6 +304,66 @@ class Browser extends DashboardView {
...
+                  [targetClassName]: this.state.counts[targetClassName] + 1
+                }
+              });
+            },
+            error => {
+              let msg = typeof error === "string" ? error : error.message;
+              if (msg) {
+                msg = msg[0].toUpperCase() + msg.substr(1);
+              }
+              obj.set(attr, prev);
```
RaschidJFR, line 342: Hey @sadakchap I've just run into this line while working on #2003 . I noticed that `attr` and `prev` are not defined so it's giving me an error. 

image

I'm hesitant to change this line to `obj.revert()` but I'd like to check first with you before proceeding.

sadakchap: Hey @RaschidJFR, I think we can safely remove this line. 

Earlier we use to save new row(making an api call) every time a user was updating any cell(Using [this](https://github.com/parse-community/parse-dashboard/blob/659c5445412aca3a59b1d7975ec3430b552d2676/src/dashboard/Data/Browser/Browser.react.js#L862)). In case of error, this line would revert current selected cell value. But, now I think we don't need it as this line, as this function gets call when user hits save btn not everytime user is making change in any cell.

PR #1691, opened 2021-04-28
Push Notification: Hiding setup locales message on adding all locales

sadakchap:
Hiding setup locales message on adding all locales messages when locales message is set up.


davimacedo (approved):
LGTM!

PR #1692, opened 2021-05-03
showing '(auto)' for auto-generated fields

sadakchap:
Showing 'auto' for auto-generated fields while creating new row or editing cloned rows.
Bascially for readOnly fields while adding new row.

Screenshot from 2021-04-28 23-45-52

### DIFF
--- a/src/components/BrowserCell/BrowserCell.react.js
+++ b/src/components/BrowserCell/BrowserCell.react.js
@@ -324,7 +324,7 @@ export default class BrowserCell extends Component {
             }
           }}
         >
-          {content}
+          {row < 0 ? '(auto)' : content}
         </span>
       </Tooltip>
     ) : (


davimacedo:
@dplewis @mtrezza what do you think? I think it is more clear for the users to understand which fields should be filled and which should not.

mtrezza:
Yes, makes sense, maybe gray it out to differentiate between that and a string field with value `(auto)` in it.

sadakchap:
Sure @mtrezza , now it looks like this - 

Screenshot from 2021-05-05 18-22-12

mtrezza:
Thank you, I think that looks good!

mtrezza (approved):

davimacedo (approved):
LGTM!

Issue #1693, opened 2021-05-03
Exclude fields from table. not just hide


itzharDev:
any way to exclude fields from the parse dashboard (not only hiding, actually not getting this field data from dashboard query)?

something like select\exclude as exists in SDK.

itzharDev:
in Browser.react.js

this lines just do the job:

    query.limit(MAX_ROWS_FETCHED);
    let columns = localStorage.getItem(`ParseDashboard:${VERSION}:${this.props.params.appId}:${source}`)
    if (columns) {
      columns = JSON.parse(columns)
      columns = columns.filter(clmn => !clmn.visible).map(clmn => clmn.name)
      for (let columnsKey in columns) {
        query.exclude(columns[columnsKey])
      }
    }

davimacedo:
I think that it is something that could be added, but why do you need that?

itzharDev:
I have a big collection in which I store base64 images in a field
I don't need to fetch this data on a browser
it couse my browser to load very slow and got stack in 100+ records

davimacedo:
That makes sense. Do you want to open a PR for this?

itzharDev:
ill send in few minutes

itzharDev:
https://github.com/parse-community/parse-dashboard/pull/1694

PR #1694, opened 2021-05-03
exculde hidden culomns

itzharDev:
https://github.com/parse-community/parse-dashboard/issues/1693

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -355,13 +355,7 @@ class Browser extends DashboardView {
     }
 
     query.limit(MAX_ROWS_FETCHED);
-    let columns = ColumnPreferences.getPreferences(this.props.params.appId, source)
-    if (columns) {
-      columns = columns.filter(clmn => !clmn.visible).map(clmn => clmn.name)
-      for (let columnsKey in columns) {
-        query.exclude(columns[columnsKey])
-      }
-    }
+    this.excludeFields(query, source)
 
     let promise = query.find({ useMasterKey: true });
     let isUnique = false;
@@ -378,6 +372,16 @@ class Browser extends DashboardView {
 
     const data = await promise;
     return data;
+  }
+
+  excludeFields(query, className) {
+    let columns = ColumnPreferences.getPreferences(this.props.params.appId, className)
+    if (columns) {
+      columns = columns.filter(clmn => !clmn.visible).map(clmn => clmn.name)
+      for (let columnsKey in columns) {
+        query.exclude(columns[columnsKey])
+      }
+    }
   }
 
   async fetchParseDataCount(source, filters) {
@@ -458,13 +462,7 @@ class Browser extends DashboardView {
       query.addDescending('createdAt');
     }
     query.limit(MAX_ROWS_FETCHED);
-    let columns = ColumnPreferences.getPreferences(this.props.params.appId, source)
-    if (columns) {
-      columns = columns.filter(clmn => !clmn.visible).map(clmn => clmn.name)
-      for (let columnsKey in columns) {
-        query.exclude(columns[columnsKey])
-      }
-    }
+    this.excludeFields(query, source)
 
     query.find({ useMasterKey: true }).then((nextPage) => {
       if (className === this.props.params.className) {
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -355,7 +355,7 @@ class Browser extends DashboardView {
     }
 
     query.limit(MAX_ROWS_FETCHED);
-    this.excludeFields(query, source)
+    this.excludeFields(query, source);
 
     let promise = query.find({ useMasterKey: true });
     let isUnique = false;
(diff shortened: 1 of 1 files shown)


src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -355,6 +356,14 @@ class Browser extends DashboardView {
     }
 
     query.limit(MAX_ROWS_FETCHED);
+    let columns = localStorage.getItem(`ParseDashboard:${VERSION}:${this.props.params.appId}:${source}`)
```
davimacedo, line 359: I believe we should get it from `ColumnPreferences.getOrder` like in here: https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Data/Browser/DataBrowser.react.js#L28

davimacedo (changes requested):

src/dashboard/Data/Browser/Browser.react.js
itzharDev: I already tried this.
there are missing params at the query point to use it

davimacedo: They should be all available at the DataBrowser component. Have you tried to send them up?

itzharDev: tried, yes. there is no data anywhere at query point

    console.log('this.refs.dataBrowser', this.refs.dataBrowser);
    console.log('this.state', this.state);
    console.log('this.props', this.props);


aaa

itzharDev: tried, yes. there is no data anywhere at query point

    console.log('this.refs.dataBrowser', this.refs.dataBrowser);
    console.log('this.state', this.state);
    console.log('this.props', this.props);

aaa

davimacedo: I beleive it should be added in here: https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Data/Browser/Browser.react.js#L166

In the same way it is currently doing for the ordering state.

itzharDev: if you talk about `getColumnSort` it gives only a single column which class sort by

davimacedo: In a similar way you could use the `ColumnPreferences.getOrder` function in that point in order to get the visible columns and set it to the component state.

itzharDev: tried and not do the work

    ColumnPreferences.getOrder(
          {},
          this.props.params.appId,
          source
        ))

result `[]`

davimacedo: Try with:
```js
   const columnPreferences = context.currentApp.columnPreference || {}
    let order = ColumnPreferences.getOrder(
      {},
      context.currentApp.applicationId,
      props.className,
      columnPreferences[className]
    );
```

itzharDev: same

davimacedo: it is probably the first parameter. Instead of `{}`, try with `this.getColumns(props.className)`

itzharDev: 1. I don't find `getColumns` in all project
2. I've try also this:

Screen Shot 2021-05-04 at 8 43 41 PM


Screen Shot 2021-05-04 at 8 43 51 PM

davimacedo: Sorry. It is `getClassColumns`

itzharDev: its also miss some data:

Screen Shot 2021-05-04 at 8 57 26 PM

Screen Shot 2021-05-04 at 8 57 09 PM

davimacedo: It is in the right direction but you will have to figure it out. Probably, when the schema is not fetched yet, we should pass nothing in the first parameter or only consider `columnPreferences[props.className]`. Btw did you try to log the content of this?

itzharDev: I'm working on it locally but connected to my production MongoDB
probably, I build the project from my own branch and use it. if you take my solution and fix it - will be community profit

itzharDev: ok, I found the right way get columns
code updated

davimacedo (commented):
Great job! Could you please only add the `;` at the end of the lines to follow the same pattern? Thanks!

davimacedo (approved):
LGTM!

itzharDev:
great & tnx
when is the next release btw?

davimacedo:
@mtrezza @dplewis I think we can release a new version right away. What do you guys think?

mtrezza:
Agree, and by the way this is a clever PR.

itzharDev:
hi, any new with the release date?

mtrezza:
There is an issue resulting from this. When showing previously hidden columns, the fields are `(undefined)`, because their values are not fetched. Only when manually reloading the list, the values appear.

PR #1695, opened 2021-05-04
Added ESC key binding

sadakchap:
when adding new row, added ESC key binding to cancel add new row

### DIFF
--- a/src/dashboard/Data/Browser/DataBrowser.react.js
+++ b/src/dashboard/Data/Browser/DataBrowser.react.js
@@ -138,15 +138,20 @@ export default class DataBrowser extends React.Component {
     if (this.props.disableKeyControls) {
       return;
     }
-    if(this.props.newObject && this.state.current && this.state.current.row === -1){
-      // if user is editing new row and want to cancel it
-      if(e.keyCode === 27){
+    if (
+      this.state.editing &&
+      this.state.current &&
+      this.state.current.row === -1 &&
+      this.props.newObject
+    ) {
+      // if user is editing new row and want to cancel editing cell
+      if (e.keyCode === 27) {
         this.setState({
           editing: false
         });
-        this.props.onAbortAddRow();
         e.preventDefault();
       }
+      return;
     }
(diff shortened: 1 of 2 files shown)


src/dashboard/Data/Browser/DataBrowser.react.js
```diff
@@ -138,6 +138,13 @@ export default class DataBrowser extends React.Component {
     if (this.props.disableKeyControls) {
       return;
     }
+    if(this.props.newObject){
```
davimacedo, line 141: I believe if I click ESC when editting a cell, it will cancel the entire row, while it should only cancel the row I am currently editting, right?

sadakchap: yes, It should not cancel the new row, when editing other rows. Just updated the pr. please let me know about the changes

sadakchap:
now, it looks something like this -
- hit ESC on without editing new row
- hit ESC while editing new row
- hit ESC while editing other row, new row will not get cancel

EscKey

davimacedo:
It was a good progress so far. But maybe we need to improve one more thing. Let's say:
1) Click Add Row
2) Start adding value to a cell
3) Click ESC

I believe it will currently cancel the whole new row, right? I believe we should cancel only the changes done in the current cell.

sadakchap:
Sure, just updated the conditions for that.

davimacedo (approved):
It looks good to me. @dplewis @TomWFox @mtrezza any comment?

mtrezza:
Is it possible to press ESC to cancel editing the cell and then press ESC again to cancel adding the row?

sadakchap:
yes it is @mtrezza. if user hits ESC while editing, it will cancel changes to the cell, and If user hits ESC again, & if there is new row, it will cancel row.

mtrezza (approved):
Look good! Thanks for this PR.

For reference, this has been discussed in https://github.com/parse-community/parse-dashboard/pull/1690#issuecomment-829150691.

TomWFox:
Nice improvement!

PR #1697, opened 2021-05-05
Cloning Rows

sadakchap:
- disabled clone option for standard classes like _User, _Session, _Installation, etc
- providing a way to edit clone row when this operation is getting failed due to error 137: Error code indicating that a unique field was given a value that is already taken.

cloneOS


davimacedo:
@sadakchap thanks for the PR! I think we don't need to disable it for standard class but use the same process of editing the unique values instead. What do you think?

sadakchap:
Sure @davimacedo , updated the pr for that.

davimacedo:
@sadakchap could you please resolve the conflicts?

davimacedo (approved):
It looks good to me. @dplewis @TomWFox @mtrezza thoughts?

mtrezza:
What is the UI flow in case 2 rows are cloned, one row fails; does that mean that 1 row is already saved and the other one is pending for editing?

In the video is not clear how you complete editing, as there are no buttons like "Clone", "Cancel". I assume you hit enter to save the pending row, or does it auto-try to save after every edit?

sadakchap:
> What is the flow in case 2 rows are cloned, one row fails; does that mean that 1 row is already saved to the Db and the other one is pending for editing?

I think, it will show both of the rows for editing, which not good. maybe we can check for each error in response array & then display

> In the video is not clear how you complete editing, as there are no buttons like "Clone", "Cancel". I assume you hit enter to save the pending row, or does it auto-try to save after every edit?

It is auto-saving, just like any other rows.

> Maybe you want to add the buttons like we have for adding a row, so the user can choose on a per row basis which pending row to save or to cancel.

ok, I like the idea. I will try that.

mtrezza:
> I think, it will show both of the rows for editing, which not good. maybe we can check for each error in response array & then display

In case some of multiple rows in a clone operation succeed and some fail, what would the expected behavior be? Should some rows already be saved to the DB while others can be cloned / canceled on a per row basis? Or should the whole clone op be pending, even for rows that could be cloned successfully?

sadakchap:
> > I think, it will show both of the rows for editing, which not good. maybe we can check for each error in response array & then display
> 
> In case some of multiple rows in a clone operation succeed and some fail, what would the expected behavior be? Should some rows already be saved to the DB while others can be cloned / canceled on a per row basis? Or should the whole clone op be pending, even for rows that could be cloned successfully?

@mtrezza , updated the pr. Now this is its behaviour "some rows already be saved to the DB while others can be cloned / canceled on a per row basis".
Some rows which are cloned successfully - they are saved to DB & added to data part of state(for UI updation part)
some rows which fail - they are NOT saved to DB, and are left for clone option.

sadakchap:
Now, the flow is something like this
finalCloneOS

mtrezza:
That is pretty sweet! What happens if a row does not have a conflict, as mentioned earlier? Will it be saved, with just the conflicting rows pending?

sadakchap:
> * What happens if a row does not have a conflict, as mentioned earlier? Will it be saved, with just the conflicting rows pending?

It will be saved, and only conflicting rows will be left pending for editing.

> * What happens if there are 100 rows conflicting? How can one cancel all pending rows?

right now, we don't have any way to cancel all pending rows at once. Maybe we can add Esc key binding for cancelling all pending rows. Please let me know what you think about "adding Esc key binding for cancelling all pending rows".

mtrezza:
> Please let me know what you think about "adding Esc key binding for cancelling all pending rows".

Let's see...

- It's a significant action for a simple keystroke, but if there was a confirmation dialog, that could compensate this imbalance.

- In UI design, a key is usually bound to a UI element. The UI is actually in a "cloning modal state" while there are pending rows, as a result of an ongoing but interrupted cloning operation.

Maybe an intuitive solution would be to add a new temporary menu item like "Cloning" that contains sub items:
- Cancel all pending rows

That would be extensible with other items in the future, like to explicitly save all non-conflicting rows, which currently happens automatically, but may not be desired in cases where a user wants cloning to fail for *every* row if there are conflicts in *any* row.

sadakchap:
> Maybe an intuitive solution would be to add a new temporary menu item like `Cloning` that contains a sub item `Cancel all pending rows`

Could you please share a screenshot showing exactly where to add `Cancel all pending rows` option?

mtrezza:
Maybe right of the existing `Edit` you could insert a new menu item `Clone` with a sub-item `Cancel all pending...`

sadakchap:
@mtrezza , do we want something like this ?
image

Can we also move "Clone this row" option to "Clone" menu item ?

mtrezza:
> @mtrezza , do we want something like this ?

Yes, that's why I was thinking of.

> Can we also move "Clone this row" option to "Clone" menu item ?

I would not. "Cloning" fits well under "Edit". The "Cloning" menu is a modal menu that is associated the modal UI state. Think a spreadsheet editor; when you edit a Pivot Table the menu changes, because the UI enters a modal state.

sadakchap:
Sure @mtrezza, now it looks like this...

clonefinalist

mtrezza:
Very nice! Exactly what I had in mind, and I assume the modal menu also disappears when all rows are individually cancelled.

@davimacedo What do you think?

sadakchap:
> I assume the modal menu also disappears when all rows are individually cancelled / cloned.

yes, it does!

davimacedo:
@mtrezza it looks even better for me now.

mtrezza:
@davimacedo Did you so a click-through to see whether there are any obvious bugs or should I do it?

davimacedo:
I checked it when I first approved but not after the latest changed you asked.

davimacedo:
I can do it later again.

mtrezza:
I tried to install this branch but I get this error with `npm i`:

```

ERROR in ../node_modules/graphql-language-service-parser/esm/onlineParser.js 81:44
Module parse failed: Unexpected token (81:44)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|             : state.rule[state.step];
|         if (state.needsSeperator) {
>             expected = expected && expected?.separator;
|         }
|         if (expected) {
 @ ../node_modules/graphql-language-service-parser/esm/index.js 4:0-57 4:0-57
 @ ../node_modules/codemirror-graphql/mode.js
 @ ../node_modules/graphql-playground-react/lib/components/Playground/SchemaExplorer/SDLEditor.js
 @ ../node_modules/graphql-playground-react/lib/components/Playground/SchemaExplorer/SDLView.js
 @ ../node_modules/graphql-playground-react/lib/components/Playground/GraphQLEditor.js
 @ ../node_modules/graphql-playground-react/lib/lib.js
 @ ./dashboard/Data/ApiConsole/GraphQLConsole.react.js
 @ ./dashboard/Dashboard.js
 @ ./dashboard/index.js
```

mtrezza:
Testing this branch I found the following:

- Cloning a row with a required field that is not filled out results in the cloning op to fail. This also means that the checkboxes are unchecked. If someone is selecting specific rows, then cloning fails, they loose all the effort they put into selecting these rows. Maybe the checkboxes should at least not be unchecked if the op fails.
- Cloning a row in the "User" class seems to succeed in the dashboard, displaying a user with same object ID and username; the cloned row disappears after refreshing the view

sadakchap:
Thanks @mtrezza for feedback, :slightly_smiling_face: 

Updated the pr!

- For required fields, I think we won't be able to add any row without filling required fields (cloning happens after that) for Custom Classes. But for `User` class, the value of password isn't present there and that's why cloning fails due to "required field was not given". Now, added a condition for that as well.
- For refresh button, I think this option will be greyed out, if clone op fails & user won't be able to hit that.

Note: clone rows will be shown only if op fails due to "duplicate error or password missing error".

clonemaybefinal

mtrezza:
In't there a conceptual issue if cloning in user failed, but there was no error dialog? Regardless if why it fails and what the class is, it should never appear to succeed, but actually fail in the background.

sadakchap:
> Isn't there (additionally) a conceptual issue if cloning in User class failed, but there was no error dialog? Regardless why it fails and what the class is, it should never appear to succeed and actually fail in the background.

True, could you please try it again with new commit.

> I also wonder how does this play with #1694, when columns are hidden, not fetched from the server and an object is cloned. I assume cloning happens dashboard side, not server side. If an object is missing field values due to hidden columns, will the cloned object miss these values in the DB? Or is cloning fetching the whole object and then duplicating it? I tried it out and it seems to work, the hidden fields are also cloned, but I haven't looked into the mechanism behind it, maybe they were still cached somewhere.

It will fetch selected rows & then will try to duplicate it.

mtrezza:
I think I found the issue in the user class, actually 2 issues:
1. Cloning a user where no authData is set, works are you described. But it does not work when the user has `authData` set, for example:
  ```
   {
    "anonymous": {
      "id": 1622296206
    }
  }
  ```
  In this case, the row appears to be cloned (incl. same objectID, email, etc) without any error message. On refresh, the row disappears, so it is never saved to the DB.
2. Not sure whether this is related to this PR, but if one column in the `User` class is hidden, the `authData` field shows `(undefined)`, even if a value is set. Only if all columns are shown and the view is reloaded, the `authData` fields display correctly with values.

sadakchap:
> 1. Cloning a user where no authData is set, works are you described. But it does not work when the user has `authData` set,      
>    In this case, the row appears to be cloned (incl. same objectID, email, etc). On refresh, the row disappears, so it is never saved to the DB. The concerning issue is that - again - this all happens without any error message, which indicates bad error handling logic.

Thanks a ton :star_struck: , I was finally able to reproduce this issue. But, It seems like request( when cloning we are making a batch request ) did went well, giving response 200. The request itself is not failing. :confused: 

> 2. Not sure whether this is related to this PR, but if one column in the `User` class is hidden, the `authData` field shows `(undefined)`, even if a value is set. Only if all columns are shown and the view is reloaded, the `authData` fields display correctly with values.

Yes, seems like an issue with https://github.com/parse-community/parse-dashboard/pull/1694 not sure. It works fine at my side.
excludefields

mtrezza:
> Thanks a ton 🤩 , I was finally able to reproduce this issue. But, It seems like request( when cloning we are making a batch request ) did went well, giving response 200. The request itself is not failing. 😕

Strange, could you try to reproduce this without dashboard, for example with a curl request or cloud code? We may need to open an issue in the parse server repo.

sadakchap:
> Strange, could you try to reproduce this without dashboard, for example with a curl request or cloud code? We may need to open an issue in the parse server repo. 
```js
let query = new Parse.Query('_User');

// Request seems to be successful with 200 status
query.equalTo('objectId', 'sd9jnXEG54');
const res = await query.find({ useMasterKey: true });
let toClone=[];
for (const object of res) {
  toClone.push(object.clone());
}
await Parse.Object.saveAll(toClone, { useMasterKey: true });

// We get an error message(Password required) this way
// const res = await query.get('sd9jnXEG54');
// let clonedObj = res.clone();
// clonedObj.save();


```


> From what I understand you are sending the request, the server responds with 200 but the object is not stored?
yes. And request returns new data as well.

mtrezza:
> And request returns new data as well.

But no object is saved to the DB?

sadakchap:
Couple of things regarding creating User object :
1. `username` & `password` is required to create new User object
2. when `authData` is set, then `username` & `password` are **NOT** required to create new User object(I think we are using this in case of 3rd party authentication like facebook, google, etc. These do not share password). When creating an object with `authData` only it automatically creates random username.


> But no object is saved to the DB?

yes, I think so. As after refreshing, that particular row, which was cloned with `authData` set is disappears.

> Are we talking about the same issue? In my case I see the same object ID as the cloned object in the dashboard, which obviously cannot be possible. And when I refresh, the row is gone, because nothing is saved to the DB.

yes, request is returns that object with same `objectId`

mtrezza:
Could you open an issue in the parse server repo for the case of cloning and same obj id is returned? Ideally with a failing test case. So we can fix the issue there before merging this dashboard PR.

sadakchap:
> Could you open an issue in the parse server repo for the case of cloning and same obj id is returned? Ideally with a failing test case. So we can fix the issue there before merging this dashboard PR.

Sure, https://github.com/parse-community/parse-server/issues/7422


> I haven't looked into this, but I think that is expected. Auth data is stored differently in the DB than how it is displayed in the dashboard, which is just a representation. So if you enter no auth data entry (`{}`), no auth data is stored in the DB (not even an empty object) and the representation for "no auth data" is `undefined`.

Interesting! If no `authData` is saved in DB & username & password is not provided. We get new row with random `username`. Sounds good.

mtrezza:
> We get new row with random username & unique objectId

Is that a suggestion or an observation?

sadakchap:
> Is that a suggestion or an observation?

Observation

mtrezza:
> If no authData is saved in DB & username & password is not provided. We get new row with random username & unique objectId.

Generating a random username of a predetermined format seems rather arbitrary and unexpected behavior. It should return an error in my view, because Parse Server is handling the error and execution an action that a developer may not want. If that is really the current behavior of Parse Server, we could rethink that, but that's independent of this dashboard feature.

Going back to the two issues relevant for this PR:

> authData is undefined on that particular row (strange)

I think that is clarified with my explanation above, if not, let's discuss.

> request is returns that new object with same objectId

That should be fixed before implementing this PR because this is a severe bug.

sadakchap:
After [this](https://github.com/parse-community/parse-server/issues/7422) discussion, now I'm thinking does it even make sense to let users clone a `_User` object. As 
- `username` needs to be unique (cloning duplicates it)
- `password` is required but it is never present in dashboard(for obvious reason)
- if `authData` is present, it does not make sense to duplicate(cloning duplicates it) [comment](https://github.com/parse-community/parse-server/issues/7422#issuecomment-865611150)
- other fields are `emailVerified`, `email`, can be duplicated.

What are your thoughts on not letting users clone `_User` object.

mtrezza:
I think it's valid to let one clone a user object. A user object may contain a number of custom fields one wants to clone. I think we just need to understand which built-in fields not to clone.

sadakchap:
> I think it's valid to let one clone a user object. A user object may contain a number of custom fields one wants to clone. I think we just need to understand which built-in fields not to clone.

ya, true. We can remove `username` & `authData` from cloned objects, as we don't want them to get duplicated.

sadakchap:
Updated the pr, now removing `username` & `authData` before saving the cloned User objects. And I think it's ready for review again.

mtrezza:
Tried cloning user and saw:

image

- The username field does not show `(required)` but is blank
- The password field does not show `(required)` but `(hidden)` which looks as if a password was already set, but it is not, so it should show `(required)`, see https://github.com/parse-community/parse-dashboard/pull/1720#issuecomment-865321628

sadakchap:
>see [Showing * for required Columns #1720 (comment)](https://github.com/parse-community/parse-dashboard/pull/1720#issuecomment-865321628)

This https://github.com/parse-community/parse-dashboard/pull/1720 has not been merged to master yet, so it's changes are not here.

mtrezza:
I see, so this issue should self-resolve once both PRs are merged. I will ignore this issue then in this review.

src/dashboard/Data/Browser/BrowserTable.react.js
```diff
@@ -115,6 +115,64 @@ export default class BrowserTable extends React.Component {
         (rowWidth, { visible, width }) => visible ? rowWidth + width : rowWidth,
         this.props.onAddRow ? 210 : 0
       );
+      let editCloneRows = null;
```
mtrezza, line 118: Why `null`?

sadakchap: Initially `null`, as we don't want to render any rows.

mtrezza: Wouldn't it be enough to declare it and check for `undefined`?

```
let editCloneRows;

if (editCloneRows) ...

```

sadakchap: Yes, we can declare it & check for `undefined`. but, I think this variable is going to hold html node element. So, is check necessary here? we can also directly put something like this
``` 
let edilCloneRows;
...
<div>
 {editCloneRows}
</div>
```

mtrezza: Sure, I just wanted to know whether the `= null` is necessary. If not then I suggest to remove it to keep the code simple. There is also another `= null` below.

sadakchap: >  There is also another `= null` below.

Yes, I followed this exact reference. :nerd_face:
I don't think that `=null` is necessary, should I remove it then ?

mtrezza: If it's unnecessary, please remove it. Otherwise one may asume there was a reason why this is set to `null`, which is a distinct value that is different from `undefined`.

sadakchap: Sure, please see https://github.com/parse-community/parse-dashboard/pull/1697/commits/be85c8e6349540ae458c77792b749a7280b98436

mtrezza: The `username` field still is blank instead of `(undefined)`:
image

Will that resolve with merge of #1720 as you [commented](https://github.com/parse-community/parse-dashboard/pull/1697#issuecomment-870164371), or is this a separate issue?

mtrezza:
The `username` field still is blank instead of `(undefined)` when cloning a user:
image

Will that resolve with merge of #1720 as you [commented](https://github.com/parse-community/parse-dashboard/pull/1697#issuecomment-870164371), or is this an issue that needs to be fixed in this PR?

sadakchap:
@mtrezza  thanks for spotting that out :eyes:. Updated the pr :)

src/icons/clone-icon.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mtrezza, line 1: Where did you get this clone icon from and what is the icon license?

mtrezza (changes requested):

src/icons/clone-icon.svg
mtrezza: @sadakchap Just pinging in case you missed this comment. We need to be mind the license of the icon, once this is resolved, this PR is ready for merge.

sadakchap: I got this icon from font awesome, [License](https://fontawesome.com/license/free) we are using solid version of their icon

sadakchap: Sure

Parse-Dashboard/parse-dashboard-config.json
```diff
@@ -1,14 +1,133 @@
...
   "apps": [
     {
-      "serverURL": "http://localhost:1338/parse",
-      "appId": "hello",
-      "masterKey": "world",
-      "appName": "",
-      "iconName": "",
-      "primaryBackgroundColor": "",
-      "secondaryBackgroundColor": ""
+      "serverURL": "https://parseapi-homolog.back4app.com",
```
mtrezza, line 4: Please remove this file and I recommend to invalidate all exposed keys and secure your apps.

mtrezza (changes requested):

Parse-Dashboard/parse-dashboard-config.json
sadakchap: Thanks @mtrezza

mtrezza:
I noticed that the `package-lock.json` has been regenerated, but package.json has not been modified.
Can you revert `package-lock.json`? We only regenerate it in a separate PR, to better control any changes.

sadakchap:
> I noticed that the `package-lock.json` has been regenerated, but package.json has not been modified.
> Can you revert `package-lock.json`? We only regenerate it in a separate PR, to better control any changes. See [parse-community/parse-server#7417](https://github.com/parse-community/parse-server/issues/7417).

Sure I can, but it was failing the ci.

mtrezza:
I see. CI should be passing when you merge master into this branch and then just use the package-lock file from master, instead of the merged file.

sadakchap:
@mtrezza I revert `package-lock.json` file, and synced it with `master` branch, but it is still failing the CI :roll_eyes: 

Apart from the failing ci, I think this pr is ready to be reviewed.

mtrezza:
Thanks for letting me know! I'll take a look why the package-lock is causing issues and do a review.

mtrezza:
@sadakchap Could you give me access to this branch? Then I can push the correct package-lock.

sadakchap:
> @sadakchap Could you give me access to this branch? Then I can push the correct package-lock.

Sure, Done! you should get an invite :)

mtrezza:
Thanks @sadakchap, I updated package-lock.json with the file from master and the CI passes now. Could you solve the remaining conflicts, so we can merge this?

sadakchap:
> Thanks @sadakchap, I updated package-lock.json with the file from master and the CI passes now. Could you solve the remaining conflicts, so we can merge this?

Sure, resolved merged conflicts!

sadakchap:
@mtrezza, I have resolved the remaining conflicts. Please let me know if we need more changes :)

mtrezza:
I am looking at it right now and I notice the row height issue. I am now checking whether the issue exists on master - if that is the case, I will merge the fixing PRs first.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -773,42 +859,85 @@ class Browser extends DashboardView {
     }
 
     if (isNewObject) {
-      // for dynamically changing required placeholder text for _User class new row object
```
mtrezza, line 776: Was this removed on purpose?

mtrezza (commented):
Just want to make sure these code blocks were removed on purpose.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -347,10 +351,6 @@ class Browser extends DashboardView {
   }
 
   addRow() {
-    if (this.props.params.className === '_User') {
```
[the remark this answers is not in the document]
sadakchap: Earlier we had a state variable in Browser component which had list of all required fields. And specifically for `_User` class we were changing it dynamically by mutating this state variable accordingly in `updateRow` function. So, before adding a new row we were reassigning this variable to initial values.

But, now I have moved dynamically changing part to child component  [`BrowserRow.js`](https://github.com/parse-community/parse-dashboard/pull/1697/commits/846bca5d31a0fbc826548d47aa167f0c8e3a6465#diff-7ddbc5db133a2fa635df33b04f527538c31fa2d1890bc6a3c7dbb0d8ba5f82a0R24). So, now we won't need to reassign to initial value.

src/dashboard/Data/Browser/Browser.react.js
sadakchap: Since, dynamically changing required placeholder logic has now moved to child component `BrowserRow`.

Why did we need to move that logic down?

In this pr, we are also giving an option to edit failed cloned rows, having a state variable for keeping list of required field for that class will affect all rows that are in editing mode, i.e, if I fill `username` in first cloned edit row, it will change the placeholder for all of the rows, which is not we wanted. 

Since, each row knew which fields are filled and which are empty and also had required columns. So, I moved that logic down.

src/dashboard/Data/Browser/Browser.react.js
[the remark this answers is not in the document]
mtrezza: Amazing!

src/dashboard/Data/Browser/Browser.react.js
mtrezza: Makes sense

mtrezza (approved):
Thanks for this amazing new feature and your patience. I would say, ready for merge!

PR #1703, opened 2021-05-10
Updating counter in sidebar

sadakchap:
On success clone rows operation, updating data counter of the class for sidebar.


davimacedo (approved):
LGTM!

Issue #1551, opened 2020-06-02
Data browser shows only first 200 records


kiranshendre:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.

- [x] You're running version >=2.3.2 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
node: v13.11.0
npm: 6.13.7
parse-server: "4.2.0"
parse-dashboard: "2.1.0"
Setup is on localhost.

#### Steps to reproduce
- Update the existing parse-dashboard (2.0.5) using below commands.
`npm cache clean --force`
`npm update`
- Verified the changes with parse-dashboard - master branch, but issue still present, records aren't scrolling.
- Tried above with deleting node_modules and re installing, but no luck

![Capture](https://user-images.githubusercontent.com/12776120/83541797-c075d500-a517-11ea-89d7-a85b40ee48e5.PNG)

Please help.

kiranshendre:
Any help on this please?

kiranshendre:
@douglasmuraoka Could you please suggest here, am I missing anything?

mtrezza:
Reopening this issue, because #1706 has been reverted by #1800 due to index issues described [here](https://github.com/parse-community/parse-dashboard/pull/1706#issuecomment-919064899).

mathieulb:
Any news about this ? Is this bug related to running an old version of MongoDB ? Because I've had this problem for a while, but I didn't have this problem in 2021, which is after issue 1551 was posted. No warning/error message in browser console nor in server log. (I can't just upgrade my MongoDB version, we'll have to replace the server.)

I'm running Parse 5.3.0, Parse Server 7.3.0, Parse Dashboard 6.0.0, Node 20, MongoDB 3.x (js mongodb@6.8).

mtrezza:
An easy interim solution would be https://github.com/parse-community/parse-dashboard/issues/2571, just no-one has picked this up yet.

mtrezza:
Closed via https://github.com/parse-community/parse-dashboard/pull/2659

PR #1706, opened 2021-05-12
Infinite Scroll query update

sadakchap:
issue - https://github.com/parse-community/parse-dashboard/issues/1551

- updated initial query to sort records according to 'objectId'
- updated fetch next page query to get new records according to 'objectId'

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -363,6 +363,14 @@ class Browser extends DashboardView {
       query.ascending(field)
     }
 
+    if (field !== 'objectId') {
+      if (sortDir === '-') {
+        query.addDescending('objectId');
+      } else {
+        query.addAscending('objectId');
+      }
+    }
+
     query.limit(MAX_ROWS_FETCHED);
     this.excludeFields(query, source);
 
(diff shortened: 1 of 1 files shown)


davimacedo (approved):
Thanks for the PR. LGTM!

Klaitos:
Hello guys, @sadakchap since this commit we have big performance issues because of the initial sort done on the objectId + createdAt.
We have collection with millions of document (GB of data) where sorting by objectId+createdAt is not possible even with database indexes. Do you have any idea of a workaround ?

davimacedo:
Would you be able to identify the query that is being sent to MongoDB? Could you also share the indexes that you have in place for this collection?

Klaitos:
By workaround I meant why do we have to order by createdAt AND objectId at first load ?  
Moreover objectId is alphanumeric characters string. Before this PR, the first load was only done by createdAt.

mtrezza:
It's unclear to me how this PR intends to solve #1551. @sadakchap can you give more details?

We are lacking the high-level discussion here in the issue that should describe the approach of this PR and how it solved the issue.

sadakchap:
Earlier we were using `createdAt` as default sort on first load & infinite scroll was working fine with this, as subsequent request will be fetching next records based on last record.
But, if we were to sort records other than `createdAt`, the infinite scroll was getting stuck.. not at 30 records but maybe somewhere around 100 or 200.

#### Reason - 
For infinite scroll we are not using [skip](https://docs.mongodb.com/manual/reference/method/cursor.skip/), we are using sorting attribute ( can be `createdAt`, `name` or other fields in class ) & the **last record**. If we give wrong last record, we will get wrong next set of records.
E.g. let's say we are sorting with `name`, the first request will get records sorted according to `name`, and next subsequent request will get records sorted according to `name`, `objectId` & last record from the first request. And these two queries are different & gives wrong last record for next query, resulting in infinite scroll getting stuck.

#### Approach -
Making initial query also sorted according to `objectId` (why only `objectId`? as this will work as a tiebreaker, when some records have same value in sorting attribute since its always unique & its natural feature which makes `objectId` incremental). Please refer [this](https://medium.com/swlh/mongodb-pagination-fast-consistent-ece2a97070f3).

mtrezza:
@Klaitos Did the suggested composed index of `createdAt` and `objectId` solve the perf issues you experienced? I would like to verify this, because both fields on their own are high selective, so I wonder whether that would improve any perf at all.

Klaitos:
> @Klaitos Did the suggested composed index of `createdAt` and `objectId` solve the perf issues you experienced? I would like to verify this, because both fields on their would create highly selective indices, so I wonder whether that would create a performant index at all.

@mtrezza Unfortunately it depends on the collection data volume, we have a very very huge collection where we can't explore anymore via the parse-dashboard because of that new default sort. Even with the compound index mongo can't make it. On other collections it works, we had to implement a solution to create the new index on every collection (100+), not a big deal but i think this should be noticed somewhere for future users.

However we have a custom build of the parse dashboard in which we override some default behavior (default count on every collection is also an expensive operation for us), removing the default compound sort is on the list.

mtrezza:
> Even with the compound index mongo can't make it.

It would be interesting to get more details about your case if you'd like to share. Specifically the query planner `explain` log, to understand what is going on. Whether there is a general issue here or there is something specific to your use case. In any case, ideally we can make a change to the dashboard so that you don't have to maintain your custom solution. If you'd like to explore that, you could open a new issue to deep-dive into that.

Klaitos:
I sincerely think the issue is on our side because of the data architecture on one of our collection. We do not use the parse dashboard as data explorer but mostly for useful quick crud operation with precise find criteria. So reverting this commit was effective for us. If we are the only one affected by this commit perf issue let it go as it is no worries.

mtrezza:
Sounds good; if you change your mind at any point in the future, please feel free to open a new issue.

mtrezza:
@Klaitos I came across the same issue you described. This is actually a breaking change, that can be solved with compound indices. I think the challenge with this approach is that it is a breaking change that requires special indices just for dashboard browsing.

For example, if an app is using a complex or resource intensive index, dashboard requires the same index plus the additional `objectId` in the index compound. I don't think this is a feasible solution, given the high cost of indices on a DB.

mtrezza:
I have opened https://github.com/parse-community/parse-dashboard/issues/1799 for this. This PR should probably be reverted or modified to not force-add the `objectId` to the data query.

PR #1708, opened 2021-05-15
Improve cell value copying of pointer cell

sadakchap:
For ObjectId column,  since onDooubleClick it opens a popup, so it not very clear how to copy the value. So now, Showing tooltip text on mouse enter & hiding it on mouse leave.

showTooltip


davimacedo (approved):
It looks good to me. @dplewis @mtrezza @TomWFox thoughts?

mtrezza:
- It doesn't seem very visually clear which cell the tooltip is referring to and thus which cell value is copied.
- A tooltip is an unusual way of displaying a cell selection. A usual way would be cell highlighting through background color or border color change.
- A tooltip has a level of visual intrusion in the UI that needs to be justified. That is why tooltips are usually displayed after a delay. This tooltip does not contain any new info per cell, so that intrusion get unnecessary pretty quickly.

My suggestion: highlight a cell in a more usual way as mentioned above and add a soft visual feedback after keystroke to indicate which cell has been copied, for example a fading in background color.

mtrezza (changes requested):
- It doesn't seem very visually clear which cell the tooltip is referring to and thus which cell value is copied.
- A tooltip is an unusual way of indicating a cell selection. A usual way would be cell highlighting through background color or border color change.
- A tooltip has a level of visual intrusion in the UI that needs to be justified. That is why tooltips are usually displayed after a delay. This tooltip does not seem to have a delay (or it is negligible), nor does it contain any cell specific info, so that intrusion get annoying pretty quickly.
- The text inside the tooltip becomes unnecessary after reading it for the first time and making the user aware that they can copy the cell by hovering + keystroke. It is basically a handbook excerpt. That indicates that a tooltip is an incorrect UI element to address this issue.

One way to address the points above would be to highlight a cell in a more usual way as mentioned above and add a soft visual feedback after keystroke to indicate which cell has been copied, for example a fading in background color.

However, this mechanism would introduce an additional way to select a cell, by hovering. And just for value copying. This looks like an XY problem. The underlying issue is that a click on a pointer cell follows the pointer instead of selecting the cell. This is an unintended shortcut.

In my opinion, the cleanest approach  would be to select the cell on first click, and follow the pointer on second click. Also, if the key binding should be to a UI element, it could be formally bound to `Copy` in the cell context menu.

mtrezza:
@sadakchap @davimacedo Do you think it makes sense to selecting the cell on first click and follow the pointer on second click? Seems to be the simplest approach, and we don't need the tooltip.

sadakchap:
Yes @mtrezza , it make sense. And I also really like the new suggested layout for pointer fields. 

But, I don't clearly understand how we will be telling user that this cell value can be copied using `ctrl+c` ?

mtrezza:
> But, I don't clearly understand how we will be telling user that this cell value can be copied using ctrl+c

It can be considered common knowledge and is therefore intuitive that selecting a cell and pressing Ctrl+C copies the cell value. So when clicking into the cell 1 time, it is selected and the border color indicated that by changing to dark blue, just like any other cell.

src/icons/arrow.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
```
mtrezza, line 1: Don't we already have a built-in set of icons to use?

sadakchap: we have couple of arrow icons, they look this way
image
or
image

If we already have something similar to [this](https://github.com/parse-community/parse-dashboard/pull/1708#issuecomment-850977855), could you please redirect me to that file.

Ah, about the license of icon, I really don't know much. I took it from [here](https://icons.getbootstrap.com/icons/arrow-right-short) & I think, its open source.

sadakchap:
Now, it looks something like this,

cellClick

Note: user will only able to follow pointer cell link only if that cell is selected else hovering on icon will show normal arrow cursor.

src/icons/arrow.svg
mtrezza: The [icons](https://github.com/twbs/icons) seem to be licensed under MIT, so we would be good to go, but would need to include a copyright notice.

mtrezza:
> Note: user will only able to follow pointer cell link only if that cell is selected else hovering on icon will show normal arrow cursor.

I think that should not be the case. A click on the icon should immediately follow the link, regardless of whether the cell is selected or not.

src/icons/arrow.svg
sadakchap: ya sure, we can use this.

mtrezza: Where did you find that icon? I can't seem to find it under /src/icons

sadakchap: [Here](https://github.com/parse-community/parse-dashboard/blob/master/src/icons/right-outline.svg)

mtrezza: Ah right, I though we had one that points into the upper right corner.

mtrezza:
I think this looks good.

- Maybe we can soften the icon colors a bit, because they stand out in the otherwise quiet dashboard. Can you use the `right-solid` icon use white as the arrow color and the original pill color as the icon color?
- The file cell now looks like a text cell, so we would need an icon there:
  image
  I would suggest 3 dots in a circle, but we don't seem to have that icon.

sadakchap:
> I think this looks good.

Thanks :) 

> * The original pill also had a hover effect which seems to be lost now, I just noticed. Maybe not very important, but most clickable UI elements in the dashboard have a hover effect.

Yes, removed the hover effect as now clickable item is only arrow icon. But, as you stated we can make icon's background little darker on hover. What do you think?

> * The file cell now looks like a text cell, so we would need an icon there:
>   image

Hm, I haven't added any icon for file cell as there was no onClick, that is no action. I also tried clicking on file cell value on master branch. But, there was no action at all.

https://github.com/parse-community/parse-dashboard/blob/0ee0ae6f20fd73e52209872cbbf824b86539fbac/src/components/BrowserCell/BrowserCell.react.js#L264

mtrezza:
> Hm, I haven't added any icon for file cell as there was no onClick, that is no action. I also tried clicking on file cell value on master branch. But, there was no action at all.

I think the pill design for files was added to make it obvious that it is not a text cell but a special file cell. So I think we need something that indicates that. I think now that we have this new link icon, there is actually no need for that unique context menu anymore and we can do the following:
- For an empty file cell, display `(undefined)` as with any other cell.
- To upload a file into an empty cell, double click into the cell.
- When a file is added, show the file name and the link icon. To download a file, click the link icon.
- To delete a file, click into the cell and hits the delete key.
- There is no need for a "Replace" button anymore, as replacing means simply deleting and then uploading.

sadakchap:
Sure @mtrezza, updated the pr. please have a look & do let me know if we need more changes.

mtrezza:
> and on double click will have either option to upload file or replace file, like this

We don't need this menu on double click anymore:
- double click on empty cell should directly open the upload dialog
- double click on non-empty cell should probably do nothing. Or it could also open the upload file dialog, then it would be like "replace".

sadakchap:
> We don't need this menu on double click anymore:
> 
> * double click on empty cell should directly open the upload dialog

**upload dialog**, do you mean browser file upload dialog ?

mtrezza:
> upload dialog, I think you mean browser file upload dialog?

Yes

mtrezza:
This looks good, I think we just have to find a new layout for the cells, to avoid this (2nd row):

image

I think since we are not using the Pill layout anymore, there is no reason to keep the text centered.
How about, we make the test left aligned (like any other cell), and make the link icon right aligned.

This way, the link icon can always be visible because the text clips before the icon:

image

mtrezza:
There seems to be an issue with the modal edit row dialog.

It works in some classes, but apparently not in classes where there is a file cell:

```
[Error] The above error occurred in the <FileEditor> component:
    in FileEditor (created by EditRowDialog)
    in div (created by EditRowDialog)
    in div (created by Field)
    in div (created by Field)
    in Field (created by EditRowDialog)
    in div (created by EditRowDialog)
    in div (created by Modal)
    in Popover (created by Modal)
    in Modal (created by EditRowDialog)
    in EditRowDialog (created by Browser)
    in div (created by Browser)
    in div (created by Browser)
    in div (created by Browser)
    in Browser (created by subscribeTo(Browser))
    in subscribeTo(Browser) (created by BrowserRoute)
    in BrowserRoute (created by Context.Consumer)
    in Route (created by AppRoute)
    in Switch (created by AppRoute)
    in div
    in Unknown (created by AppRoute)
    in AppRoute (created by Context.Consumer)
    in Route (created by Index)
    in Switch (created by Index)
    in div (created by Index)
    in Index (created by Context.Consumer)
    in Route (created by Dashboard)
    in Switch (created by Dashboard)
    in div (created by Dashboard)
    in Router (created by Dashboard)
    in Dashboard

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
	logCapturedError (react-dom.development.js:19527)
	logError (react-dom.development.js:19564)
	(anonymous function) (react-dom.development.js:20708)
	callCallback (react-dom.development.js:12490)
	commitUpdateQueue (react-dom.development.js:12511)
	commitLifeCycles (react-dom.development.js:19883)
	commitLayoutEffects (react-dom.development.js:22803)
	callCallback (react-dom.development.js:188)
	dispatchEvent
	invokeGuardedCallbackDev (react-dom.development.js:237)
	invokeGuardedCallback (react-dom.development.js:292)
	commitRootImpl (react-dom.development.js:22541)
	commitRootImpl
	unstable_runWithPriority (scheduler.development.js:653)
	commitRoot (react-dom.development.js:22381)
	finishSyncRender (react-dom.development.js:21807)
	performSyncWorkOnRoot (react-dom.development.js:21793)
	performSyncWorkOnRoot
	(anonymous function) (react-dom.development.js:11089)
	unstable_runWithPriority (scheduler.development.js:653)
	flushSyncCallbackQueueImpl (react-dom.development.js:11084)
	flushSyncCallbackQueue (react-dom.development.js:11072)
	discreteUpdates$1 (react-dom.development.js:21893)
	discreteUpdates (react-dom.development.js:806)
	dispatchDiscreteEvent (react-dom.development.js:4168)
	dispatchDiscreteEvent
[Error] TypeError: null is not an object (evaluating 'fileInputElement.click')
	flushSyncCallbackQueueImpl (react-dom.development.js:77)
	flushSyncCallbackQueue (react-dom.development.js:11072)
	discreteUpdates$1 (react-dom.development.js:21893)
	discreteUpdates (react-dom.development.js:806)
	dispatchDiscreteEvent (react-dom.development.js:4168)
	dispatchDiscreteEvent
```

mtrezza:
Kindly let me know when this is ready for review.

dplewis:
You can single click a cell and press ctrl-c and a popup will show up. 
Screen Shot 2021-06-04 at 3 02 53 PM

mtrezza:
The issue that this PR solves is that a single click on a pointer cell follows the pointer. There is no way to press crtl+c before that.

dplewis:
Its a similar to how you set a field to be `undefined`, you have to click the edge of the cell not the value itself. Thats how I did it for the pointer field in my screen shot. Try it for yourself.

mtrezza:
That is exactly what this PR now solves. :)

dplewis:
This just gave me a new idea for a feature for copying the objectID when the modal is open, I’ll do a PR for it

mtrezza:
This looks good, @sadakchap can we merge this, or is there anything you want to change?

sadakchap:
> This looks good, @sadakchap can we merge this, or is there anything you want to change?

I haven't checked it on Edit Row dialog. Could you please try that ?

mtrezza:
I have found a slight layout issue, the cell text is not left aligned like all other cells:

image

sadakchap:
Got it. Will update!

sadakchap:
> 4. The modal edit row dialog still has the old pill layout; I think the field should look and work the same was as the cell does:
>    image

about the 2nd option "Select userId", if we don't show it as a pill then I think it does not look clickable 
image

mtrezza:
Do we need the "select owner", or should this be identical to the pointer cell in a row?

sadakchap:
> Do we need the "select owner", or should this cell behave identically to the pointer cell in a row?

User might not know that double clicking in Edit Row dialog cell opens Editor as for other fields usually it's single click.

mtrezza:
> User might not know that double clicking in Edit Row dialog cell opens Editor as for other fields usually it's single click. That's why I think it's good to have "select owner"(maybe with pill design).

I agree that we should think this through a bit more. We currently have different designs, some with bugs in the "Edit row" dialog. It would we good if we could agree on a design that works for the pointer, then over time correct the other fields in follow-up PRs.

For example, the calendar shows a grayed out text "Pick..." if no value is set, instead of a simple `undefined`. Once set, it becomes impossible to delete the field value, which is a similar challenge we need to solve for the pointer fields.

https://user-images.githubusercontent.com/5673677/122827682-f46b9600-d2e4-11eb-9145-77a321c8c2e5.mov

On the other hand there is this inconsistency when displaying `undefined` fields. The password field shows `hidden` as it does in the row, but the user field shows nothing:
image

I would suggest the following design for the pointer field, as a model for all fields:

- If no value is set, show `undefined` in grayed out text, similar to the row design -> same as row
- Single click into the field with `undefined` value should open the upload file dialog -> same as row
- Single click into the field with value set should select the field to allow for copy/paste -> same as row
- Single click into the field with value set and pressing "Del" key should delete the value --> same as row
- Double click on field with a value set should up open the "select object" / "upload file" dialog; it is a shortcut for "delete + set = replace" -> same as row
- Single click on link icon (only visible in a field with a value set) should follow the pointer / open the file; possibly in a new browser tab, otherwise the modal dialog is lost because it is probably not persistent?

mtrezza:
@sadakchap What are you thoughts on my previous comment? I think this PR is almost ready for merge.

sadakchap:
> @sadakchap What are you thoughts on my previous comment? I think this PR is almost ready for merge.

Yes, I really like the idea. I was trying to implement the key binding for Edit row Modal part. I will let you know when its finished :)
 .

mtrezza:
Great, if there is any point in my comment that is difficult to implement, please let me know any we could find an alternative or save that for a future PR.

mtrezza:
Just for into, there is another [amazing PR](https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r655729128) that is apparently waiting for this PR to be merged first. But no pressure 😅

sadakchap:
Hey @mtrezza , can we make edit row modal changes in other pr ? As this pr was originally for cells in Browser table and implementing same functionality as Browser cell has, will take little longer time.

mtrezza:
@sadakchap Sure, no worries. We should just make sure that the modal dialog is still usable and this PR doesn't have any negative impact. Did you notice anything?

sadakchap:
> @sadakchap Sure, no worries. We should just make sure that the modal dialog is still usable and this PR doesn't have any negative impact. Did you notice anything?

No, I haven't till now. But I will double check, and let you know here. Thank you.

mtrezza:
Great, that is the last thing I can think about, then this should be ready for merge.

sadakchap:
@mtrezza , I made some commits in order to fix some UI issues in the modal, and
- [x] Always show the input area with a light blue background.
- [x] If no value is set, show undefined in grayed out text, similar to the row design -> same as row ( only for string, number type fields)

> Single click on link icon (only visible in a field with a value set) should follow the pointer / open the file; possibly in a new browser tab, otherwise the modal dialog is lost because it is probably not persistent(?)

We are actually closing the modal, and changing the state as well as url of the current tab.

sadakchap:
Could you please review it once more, please ?

mtrezza:
Amazing! Is this ready for review?

src/components/Field/Field.scss
```diff
@@ -42,10 +42,14 @@
...
   padding: 0;
+  background: #f6fafb;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .input {
-  margin: 25px 20px 0 0;
+  // margin: 25px 20px 0 0;
```
mtrezza, line 52: This can be removed I guess

src/components/Field/Field.scss
```diff
@@ -57,7 +61,7 @@
   }
 
   .input {
-    margin: 13px 20px 0 0;
+    // margin: 13px 20px 0 0;
```
mtrezza, line 64: This can be removed I guess

mtrezza (changes requested):
just 2 comments

mtrezza:
The UI looks all good, just the comments above and we can approve this PR.

src/components/Field/Field.scss
sadakchap: remove the line or un-comment it ?

mtrezza: Do you think we should keep it as a comment?

sadakchap: I think removing them would be good.

src/components/Field/Field.scss
sadakchap: Done https://github.com/parse-community/parse-dashboard/pull/1708/commits/16010c7db70c04f249edafd95bf313fe8a195461 !

src/components/Field/Field.scss
sadakchap: Done! https://github.com/parse-community/parse-dashboard/pull/1708/commits/16010c7db70c04f249edafd95bf313fe8a195461

sadakchap:
Should I now rebase it on master and resolve any merge conflicts ?

sadakchap:
@mtrezza what should be next steps with this pr ?

mtrezza:
The good news is that we have a couple of great PRs in the pipeline that are practically ready for merge. I'll take a look over the weekend to see what merging order may cause the least friction. There will be some rebasing necessary, because most of the PRs are dealing with the DataBrowser. I'll ping you here once this is ready.

mtrezza (approved):
Looks ready for merge!

This PR is a real relief. It's so much easier to select a pointer field now - well done!

PR #1714, opened 2021-05-25
fix: Allow other key binding while adding new row

Klaitos:
Hello,

This PR allow using CTRL C/V and other key biding while adding a new row using the new Adding row feature.
Before, only escape key was binded

Capture d’écran 2021-05-25 à 11 53 09


davimacedo:
@sadakchap would you mind to leave your opinion about this PR?

sadakchap:
Looks good to me!

Thanks for the pr @Klaitos .

davimacedo (approved):
LGTM!

PR #1717, opened 2021-05-28
update on file change

sadakchap:
When creating a column of type file and selecting a file as the default value, there isn't any kind of feedback for selected file & don't know if the file uploading failed.

- showing a error note, if uploading fails
- changing file input value
- showing a spinner while uploading file

fileuploadChange


mtrezza (approved):
LGTM

@davimacedo Are changelog entries not mandatory as part of a PR in the dashboard repo?

davimacedo:
We currently only mandate changelog entry for Parse Server, but we can start doing it here as well.

mtrezza:
I think we should, to minimize the release effort.

PR #1718, opened 2021-05-28
Fix: Column name starting with numbers

sadakchap:
Creating new column name starting with number gives internal server error. Now, validating column name to not accept names starting with numbers.

### DIFF
--- a/src/dashboard/Data/Browser/AddColumnDialog.react.js
+++ b/src/dashboard/Data/Browser/AddColumnDialog.react.js
@@ -27,7 +27,7 @@ import {
 }                         from 'lib/Constants';
 
 function validColumnName(name) {
-  return !!name.match(/^[a-zA-Z0-9][_a-zA-Z0-9]*$/);
+  return !!name.match(/^[a-zA-Z][_a-zA-Z0-9]*$/);
 }
 
 export default class AddColumnDialog extends React.Component {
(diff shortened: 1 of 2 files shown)


mtrezza:
The Parse JS docs say:

> Keys must be alphanumeric strings.

Can you please reference an issue for this PR so we can have a high-level discussion about the issue and see whether this PR is the right approach?

sadakchap:
Sure @mtrezza , here's it https://github.com/parse-community/parse-dashboard/issues/1719

mtrezza:
Could you look into why there is no error message displayed in the dashboard when adding the column fails? I think that is rather concerning, see my [comment](https://github.com/parse-community/parse-dashboard/issues/1719#issuecomment-851390157).

sadakchap:
> Could you look into why there is no error message displayed in the dashboard when adding the column fails (before this PR)? I think that is rather concerning and we may need to refactor error handling, see my [comment](https://github.com/parse-community/parse-dashboard/issues/1719#issuecomment-851390157).

Yes, after looking into [code](https://github.com/parse-community/parse-dashboard/blob/0ee0ae6f20fd73e52209872cbbf824b86539fbac/src/dashboard/Data/Browser/Browser.react.js#L283). I think we are not handling errors at all, when adding a new column fails.

mtrezza:
> Yes, after looking into code. I think we are not handling errors at all, when adding a new column fails.

that's a good improvement.

mtrezza (approved):
LGTM and thanks for adding the error handling.

davimacedo (approved):
LGTM!

Issue #1658, opened 2021-02-22
null values showing in dashboard


dplewis:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [ ] You're running version >=1.0.23 of Parse Dashboard.

- [ ] You're running version >=2.3.2 of Parse Server.

- [ ] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce

I've seen this issue posted before a few years ago so this may be a duplicate. I noticed that null is showing up in my database. 

<img width="425" alt="Screen Shot 2021-02-22 at 12 09 43 PM" src="https://user-images.githubusercontent.com/9830365/108750734-012ed780-7507-11eb-88b4-80babb43a2a1.png">

In the database it show null as well

```
_id: "eKg7Ib3Lsx"
startDate: 2021-02-14T20:52:01.761+00:00
endDate: null
_created_at: 2021-02-14T20:52:02.467+00:00
_updated_at: 2021-02-22T18:04:06.333+00:00
```

I believe null is setup from iOS. I can try to replicate it.

I think we should add an `isNull` filter to find these values. I can do a PR. Just wanted to know why this is happening.

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

mtrezza:
See discussion in https://github.com/parse-community/parse-dashboard/pull/1721

PR #1721, opened 2021-06-03
Add isNull filter

sadakchap:
Feature request https://github.com/parse-community/parse-dashboard/issues/1658

Added new filter to filter records if field contains `null` value.
nullFilter

### DIFF
--- a/src/lib/Filters.js
+++ b/src/lib/Filters.js
@@ -163,14 +163,19 @@ export const Constraints = {
     field: null,
     comparable: false
   },
+  isNull: {
+    name: 'isNull',
+    field: null,
+    comparable: false
+  }
 };
 
 export const FieldConstraints = {
-  'Pointer': [ 'exists', 'dne', 'eq', 'neq', 'unique' ],
-  'Boolean': [ 'exists', 'dne', 'eq', 'unique' ],
-  'Number': [ 'exists', 'dne', 'eq', 'neq', 'lt', 'lte', 'gt', 'gte', 'unique' ],
-  'String': [ 'exists', 'dne', 'eq', 'neq', 'starts', 'ends', 'stringContainsString', 'unique' ],
-  'Date': [ 'exists', 'dne', 'before', 'after', 'unique' ],
+  'Pointer': [ 'exists', 'dne', 'eq', 'neq', 'unique', 'isNull' ],
+  'Boolean': [ 'exists', 'dne', 'eq', 'unique', 'isNull' ],
+  'Number': [ 'exists', 'dne', 'eq', 'neq', 'lt', 'lte', 'gt', 'gte', 'unique', 'isNull' ],
+  'String': [ 'exists', 'dne', 'eq', 'neq', 'starts', 'ends', 'stringContainsString', 'unique', 'isNull' ],
+  'Date': [ 'exists', 'dne', 'before', 'after', 'unique', 'isNull' ],
   'Object': [
     'exists',
     'dne',
@@ -183,6 +188,7 @@ export const FieldConstraints = {
     'keyLt',
     'keyLte',
     'unique',
+    'isNull'
   ],
   'Array': [
     'exists',
@@ -193,6 +199,7 @@ export const FieldConstraints = {
     'doesNotContainNumber',
     'containsAny',
     'doesNotContainAny',
+    'isNull'
   ]
 };
 
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -16,7 +16,7 @@ export default function queryFromFilters(className, filters) {
     query = className.query();
   }
   filters.forEach((filter) => {
-    addConstraint(query, filter);
+    query = addConstraint(query, filter, className);
   });
   return query;
 }
@@ -28,7 +28,7 @@ function addQueryConstraintFromObject(query, filter, constraintType) {
   }
 }
 
-function addConstraint(query, filter) {
+function addConstraint(query, filter, className) {
   switch (filter.get('constraint')) {
     case 'exists':
       query.exists(filter.get('field'));
@@ -105,7 +105,10 @@ function addConstraint(query, filter) {
       addQueryConstraintFromObject(query, filter, 'lessThanOrEqualTo');
       break;
     case 'isNull':
-      query.equalTo(filter.get('field'), null);
+      query.exists(filter.get('field'));
+      let nullQuery = new Parse.Query(className);
+      nullQuery.equalTo(filter.get('field'), null);
+      query = Parse.Query.and(query, nullQuery);
       break;
   }
   return query;
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -104,6 +104,9 @@ function addConstraint(query, filter) {
     case 'keyLte':
       addQueryConstraintFromObject(query, filter, 'lessThanOrEqualTo');
       break;
+    case 'isNull':
+      query.equalTo(filter.get('field'), null);
+      break;
   }
   return query;
 }


mtrezza:
Thanks for this PR.

In MongoDB, the value `null` is *sometimes* treated the same as if the field does not exist at all. In other words, these two documents are sometimes treated as identical:
```
{
  a: 1,
  b: null
}

{
  a: 1
}
```

That means there may not be difference between a "does not exist" and "is null" from a MongoDB perspective.

To unambiguously query for objects with a field of content `null` we would have to query for the [BSON Type](https://docs.mongodb.com/manual/reference/bson-types/) `10`.

sadakchap:
> In MongoDB, the value `null` is _sometimes_ treated the same as if the field does not exist at all. In other words, these two documents are sometimes treated as identical:
 
True. That's why we are filtering for both exists & equal to null value. If filtering field is `age` then we are sending a query like this.
```
where: {
  $and: [
      age: { $exists: true },
      age: null
   ]
}
```

mtrezza:
> That's why we are filtering for both exists & equal to null value. 
Do you mean that is how you personally use the filter to distinguish to query for a)?

It may return the same results, but the correct - because more efficient - query for this is:

```
where: {
    age: { $type: 10 }
}
```

sadakchap:
> Do you mean that is how you personally use the filter to query for a)?

yes!

> It may return the same results, but the correct - because more efficient - query for this is:
> 
> ```
> where: {
>     age: { $type: 10 }
> }
> ```

Yes, you are correct! But, How do I send a query like this, I searched in the docs but couldn't find anything that would query on type of the field.

mtrezza:
Not sure whether Parse Server supports this yet, you would have to look into the code, it may exist but not be in the docs.

I think we can simplify this PR by making the filter option name more descriptive to indicate that it does c) `field: null`.

How should we name the filter options for:
a) `field : { $type: 10 }`
vs.
c) `field: null`

to make them distinguishable?

sadakchap:
> Not sure whether Parse Server supports this yet, you would have to look into the code, it may exist but not be in the docs.

I look into this [file](https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/GraphQL/transformers/query.js), but I couldn't find any thing that would query on type.

> Therefore I suggest this naming to make them distinguishable:
> 
> a) `field : { $type: 10 }` filter name: `is null`
> c) `field: null` filter name: `does not exist or is null`
> 
> If later someone adds filter (a), they can name the filter `is null` and it won't conflict with the filter name in this PR.

Sounds good :+1:, so should I change the current query as well for `does not exist or is null` option 
from 
```js
query.exists(field);
let nullQuery = new Parse.Query(className);
nullQuery.equalTo(field, null); 
query = Parse.Query.and(query, nullQuery);
```
to
```js 
query.equalTo(field, null)
``` 


as the current query fetches records only with `null` value.

mtrezza:
> should I change the current query as well for does not exist or is null option

Yes I think so. Can one still query only for null values by adding the field exists filter additionally for the same field in the dashboard?

sadakchap:
> Can one still query only for null values by adding the field exists filter additionally for the same field in the dashboard?

I think only filter is allowed on a field.

mtrezza:
> I think only one filter is allowed on a field at a time. No, I don't think it is possible.

You could additionally add the `is null` filter if we find out how to do `where {  age: { $type: 10 } }`. I think it should be already possible, maybe with a REST request, if the JS SDK doesn't support this yet.

Otherwise you could add this:
```
query.exists(field);
let nullQuery = new Parse.Query(className);
nullQuery.equalTo(field, null); 
query = Parse.Query.and(query, nullQuery);
```

with a comment that this is a workaround because `where {  age: { $type: 10 } }` is not possible with the JS SDK.

src/lib/Filters.js
```diff
@@ -163,14 +163,19 @@ export const Constraints = {
     field: null,
     comparable: false
   },
+  dneOrNull: {
+    name: 'does not exists or null',
+    field: null,
+    comparable: false
```
dplewis, line 169: If you want to add additional filters you can set comparable to `true` also we should document what `comparable` and `composeable` (I think it allows you to add a value to compareTo) are

src/lib/queryFromFilters.js
```diff
@@ -104,6 +104,9 @@ function addConstraint(query, filter) {
     case 'keyLte':
       addQueryConstraintFromObject(query, filter, 'lessThanOrEqualTo');
       break;
+    case 'dneOrNull':
+      query.equalTo(filter.get('field'), null);
```
dplewis, line 108: I don't think this query will return null values.

dplewis (commented):
@sadakchap Thanks for the PR! I was the one who requested it. Just a few comments from experience.

src/lib/queryFromFilters.js
sadakchap: True, it will also include records which does not have filter field even set, that's why the name of filter `does not exits or null`

dplewis: That sounds good

> I think only one filter is allowed on a field at a time.

You can set composable to true if you to do multiple filters on a field. Similar to how less than does it. (Perhaps we should document these)

mtrezza: The difference is the MongoDB performance. The current query is an explicit "does not exist or is null" conditional, without `$and`. This distinction is also relevant for indexing.

dplewis: I mean document `composable` and `comparable` it usually takes a while to figure out what they mean.

dplewis: `field` is used when `comparable` is true. We can document that too

sadakchap: Thanks @dplewis for `composable` idea. @mtrezza I tried that but it seems like `exists` is not composable. 

I was selecting first `does not exists or null` & then `exists` for a field, but it seems to send query request with only `where={age:{$exists: true}}.

And other thing is that, if `exists` is not composable then we should even not be able to select that.

mtrezza: > this is what I tried,
> 1. select does not exists or null --> query.equalTo(field, null)
> 2. select exists for a field, --> query.exists(field)
> but it seems to send query request with only `where={age:{$exists: true}}.

It sends `where={age:{$exists: true}}` if you set a combined filter of (1) and (2)?

sadakchap: yes. 

I think that's strange as, for let's say 2 filters `less than` & `exists`, we do something like
```js
query.lessThan(field, compareTo); // compareTo is value entered by user to which s(he) wants to compare
query.exists(field);
```
This sends a query like `where={age: {$lt: 4, $exists: true}}`

Then, why not this
```js
query.equalTo(field, null);
query.exists(field);
```
sends a query like `where={age: {$eq: null, $exists: true}}`

mtrezza: @dplewis when you mentioned `composable`, on what level did you mean that? On a MongoDB level?

sadakchap: I don't think it's a limitation with MongoDB rather it's with JS Sdk.(just an opinion) 

As when I explicitly(using cURL) make a request with `where={age: {$exists: true, $eq: null }}`, it does return records only with `null` field value.

But, when I'm trying to couple `$exists` & `$eq` query, it only sends request with `where={age: {$exists: true}}`.
I'm trying this way to construct the query
```js
query.equalTo(field, null);
query.exists(field);
```
Maybe I'm missing some point with js sdk. Would really appreciate any help with creating a query with `where={age: {$exists: true, $eq: null }}`.

sadakchap: I created a new issue for this https://github.com/parse-community/Parse-SDK-JS/issues/1372.

mtrezza:
@sadakchap Do you think we can resolve any open questions here? I think this PR is almost ready for merge, it would be great to have it in the upcoming release of Parse Dashboard.

sadakchap:
> @sadakchap Do you think we can resolve any open questions here? I think this PR is almost ready for merge, it would be great to have it in the upcoming release of Parse Dashboard.

This pr was originally to add a feature through which users can filter records with only `null` value. 

But, we decided to add a filter `dneOrNull`, on which user a add one more filter `exists` => then user should be getting records with only `null` value. But, that is not the case, `$exists` and `$eq` query seems to overriding each other. Please see https://github.com/parse-community/parse-dashboard/pull/1721#discussion_r646628396 .

I also raised an [issue](https://github.com/parse-community/Parse-SDK-JS/issues/1372) in Parse-SDK-JS. But, I think I am really not the right person to solve this issue.

For this feature, maybe we can try coupling `$exists` & `$eq` with `$and`. Something like this https://github.com/parse-community/parse-dashboard/pull/1721/commits/38d2f16da23ebbfa6ce3d13e554c2f31896e2310 , maybe ?

mtrezza:
@sadakchap  So maybe we can finalize this PR like this:

- Add a filter condition "does not exist or is null" and use [(c)](https://github.com/parse-community/parse-dashboard/pull/1721#issuecomment-853907228)
- Add a filter condition "is null" as in [(b)](https://github.com/parse-community/parse-dashboard/pull/1721#issuecomment-853907228) once the Parse JS SDK [issue](https://github.com/parse-community/Parse-SDK-JS/issues/1372) is fixed and supports this properly

What do you think?

sadakchap:
Thanks @mtrezza . 

I was working on this [issue's](https://github.com/parse-community/Parse-SDK-JS/issues/1372) [**PR**](https://github.com/parse-community/Parse-SDK-JS/pull/1373). I was able to to clear all unit test cases. Right now, I think only LiveQuery Integration test cases are failing.

I know that, this is open from long time, but If we can wait just 1 more week ( so that I can try to fix those failing integration test cases). It would be really nice. Or else we can do as you have suggested(incase it takes more than a week).

mtrezza:
Sure. Let's wait as you suggest.

cbaker6:
> Another aspect to consider is how that plays with the Postgres adapter, or whether that is something unique to MongoDB.

For Postgres, the server treats `doesNotExist` and `null` the same, so I believe the filters here will produce the same results, which should be fine. I discuss more about this here: https://community.parseplatform.org/t/query-where-field-nil/2317/16?u=cbaker6

Issue #151, opened 2016-03-16
Choose which column should represent a pointer (instead of objectId)


AmbroiseCollon:
It would be nice to choose which column should represent a certain class when another points to it. 

For instance, I have a `Game` class which contains a column `player` which is a pointer to `User` class. In the rows, a button represents this relationship like so :
<img width="151" alt="capture d ecran 2016-03-16 a 11 40 58" src="https://cloud.githubusercontent.com/assets/2622390/13809729/666104fe-eb6c-11e5-9f68-ef510c2a3a6f.png">

It would be more user friendly to have something like this (here I chose `username` column) :
<img width="151" alt="capture d ecran 2016-03-16 a 11 40 58 2" src="https://cloud.githubusercontent.com/assets/2622390/13809787/af7dd7ac-eb6c-11e5-98fa-f7cacc558e8e.png">

drew-gross:
Neat idea! We'd need to fetch a bunch more data to actually retrieve the username, so we'd probably want to do this asynchronously.

AmbroiseCollon:
Any more advice on how to work on this ?

Btw, I just gave an example, but it could be good to have this not only on User class but on every class even Custom and on any column. If possible ^^!

drew-gross:
Ah, having this work on any class would be cool too. You would need to store the configuration of what column represents the object somewhere. Probably most of the modifications to make this happen would be in `Browser.react.js`, spread across `fetchData`, `fetchNextPage`, and probably a new function or two as well to handle the asynchronous part.

AmbroiseCollon:
I had a look. I'm clearly not competent on this. Never done react before. And not a big js user. But I'm happy you liked the idea. I hope someone would tackle this (or me in some future... :) )

mtrezza:
Interesting that this PR still open.

I think this is a nice idea, but before attempting any PR on this, we should agree on some fundamentals.

Some issues that come to mind:
- How to decide which representation should be displayed
- What to display if the representation field is missing
- What is the data traffic / performance impact? Does this require new indices, how are these created?
- How to copy the object ID if it is not displayed - what value is copied on Crtl + C - "Bob" or the object ID?

mtrezza:
Closed via #1723

PR #1723, opened 2021-06-03
https://github.com/parse-community/parse-dashboard/issues/151

fn-faisal:
[Issue 151 link](https://github.com/parse-community/parse-dashboard/issues/151)

- The resolution implemented in this PR for this issue is to give users access to an option in the menu where they can change the default key for a given class. 
- Throughout the Dashboard, anywhere that there is a pointer to this class object, the specified key's value (chosen in the edit menu option -> change pointer key) is used instead of objectId.

If no key is specified, the pointer value falls back to objectId.
Screenshot 2021-06-03 at 6 31 22 PM
Screenshot 2021-06-03 at 6 31 31 PM
Screenshot 2021-06-03 at 6 31 41 PM
Screenshot 2021-06-03 at 6 32 00 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,18750 @@
 {
   "name": "parse-dashboard",
   "version": "2.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "parse-dashboard",
+      "version": "2.1.0",
+      "license": "SEE LICENSE IN LICENSE",
+      "dependencies": {
+        "@babel/runtime": "7.14.0",
+        "bcryptjs": "2.3.0",
+        "body-parser": "1.19.0",
+        "codemirror-graphql": "github:timsuchanek/codemirror-graphql#details-fix",
+        "commander": "6.2.1",
+        "connect-flash": "0.1.1",
+        "cookie-session": "2.0.0-beta.3",
+        "copy-to-clipboard": "3.2.0",
+        "create-react-class": "15.7.0",
+        "csurf": "1.11.0",
+        "express": "4.17.1",
+        "graphql": "14.6.0",
+        "graphql-playground-react": "1.7.20",
+        "history": "4.10.1",
+        "immutable": "4.0.0-rc.9",
+        "immutable-devtools": "0.1.5",
+        "js-beautify": "1.13.13",
+        "json-file-plus": "3.2.0",
+        "moment": "^2.29.1",
+        "package-json": "6.5.0",
+        "parse": "2.19.0",
+        "passport": "0.4.1",
+        "passport-local": "1.0.0",
+        "prismjs": "1.23.0",
+        "prop-types": "15.7.2",
+        "query-string": "6.14.1",
+        "react": "16.14.0",
+        "react-ace": "9.4.0",
+        "react-dnd": "10.0.2",
+        "react-dnd-html5-backend": "10.0.2",
+        "react-dom": "16.14.0",
+        "react-helmet": "6.0.0",
+        "react-json-view": "1.21.3",
+        "react-popper-tooltip": "4.2.0",
+        "react-redux": "5.1.2",
+        "react-router": "5.1.2",
+        "react-router-dom": "5.1.2",
+        "regenerator-runtime": "0.13.5",
+        "semver": "7.3.4",
+        "sweetalert2": "^11.0.16",
+        "sweetalert2-react-content": "^4.0.1"
+      },
+      "bin": {
+        "parse-dashboard": "bin/parse-dashboard"
+      },
+      "devDependencies": {
+        "@babel/core": "7.8.7",
+        "@babel/plugin-proposal-decorators": "7.8.3",
+        "@babel/plugin-transform-regenerator": "7.8.7",
+        "@babel/plugin-transform-runtime": "7.8.3",
+        "@babel/preset-env": "7.9.0",
+        "@babel/preset-react": "7.8.3",
+        "babel-eslint": "10.1.0",
+        "babel-loader": "8.1.0",
+        "babel-plugin-transform-object-rest-spread": "6.26.0",
+        "core-js": "3.6.5",
+        "cross-env": "7.0.2",
+        "css-loader": "3.5.2",
+        "eslint": "6.8.0",
+        "eslint-plugin-jest": "23.8.2",
+        "eslint-plugin-react": "7.19.0",
+        "file-loader": "6.0.0",
+        "http-server": "0.12.0",
+        "jest": "24.8.0",
+        "marked": "0.8.2",
+        "node-sass": "4.13.1",
+        "null-loader": "3.0.0",
+        "parse-server": "github:parseplatform/parse-server#master",
+        "path-to-regexp": "3.2.0",
+        "puppeteer": "3.0.0",
+        "react-test-renderer": "16.13.1",
+        "request": "2.88.2",
+        "request-promise": "4.2.5",
+        "sass-loader": "8.0.0",
+        "style-loader": "1.1.2",
+        "svg-prep": "1.0.4",
+        "webpack": "4.42.1",
+        "webpack-cli": "3.3.10"
+      },
+      "engines": {
+        "node": ">=8.9"
+      }
+    },
+    "node_modules/@apollo/client": {
+      "version": "3.3.19",
+      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.3.19.tgz",
+      "integrity": "sha512-vzljWLPP0GwocfBhUopzDCUwsiaNTtii1eu8qDybAXqwj4/ZhnIM46c6dNQmnVcJpAIFRIsNCOxM4OlMDySJug==",
+      "dev": true,
+      "dependencies": {
+        "@graphql-typed-document-node/core": "^3.0.0",
+        "@types/zen-observable": "^0.8.0",
+        "@wry/context": "^0.6.0",
+        "@wry/equality": "^0.4.0",
+        "fast-json-stable-stringify": "^2.0.0",
+        "graphql-tag": "^2.12.0",
+        "hoist-non-react-statics": "^3.3.2",
+        "optimism": "^0.16.0",
+        "prop-types": "^15.7.2",
+        "symbol-observable": "^2.0.0",
+        "ts-invariant": "^0.7.0",
+        "tslib": "^1.10.0",
+        "zen-observable": "^0.8.14"
+      }
+    },
+    "node_modules/@apollo/client/node_modules/@wry/equality": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@wry/equality/-/equality-0.4.0.tgz",
+      "integrity": "sha512-DxN/uawWfhRbgYE55zVCPOoe+jvsQ4m7PT1Wlxjyb/LCCLuU1UsucV2BbCxFAX8bjcSueFBbB5Qfj1Zfe8e7Fw==",
+      "dev": true,
+      "dependencies": {
+        "tslib": "^2.1.0"
+      },
+      "engines": {
+        "node": ">=8"
+      }
+    },
+    "node_modules/@apollo/client/node_modules/@wry/equality/node_modules/tslib": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.2.0.tgz",
+      "integrity": "sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==",
+      "dev": true
+    },
+    "node_modules/@apollo/client/node_modules/symbol-observable": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/symbol-observable/-/symbol-observable-2.0.3.tgz",
+      "integrity": "sha512-sQV7phh2WCYAn81oAkakC5qjq2Ml0g8ozqz03wOGnx9dDlG1de6yrF+0RAzSJD8fPUow3PTSMf2SAbOGxb93BA==",
+      "dev": true,
+      "engines": {
+        "node": ">=0.10"
+      }
+    },
+    "node_modules/@apollo/client/node_modules/ts-invariant": {
+      "version": "0.7.3",
+      "resolved": "https://registry.npmjs.org/ts-invariant/-/ts-invariant-0.7.3.tgz",
+      "integrity": "sha512-UWDDeovyUTIMWj+45g5nhnl+8oo+GhxL5leTaHn5c8FkQWfh8v66gccLd2/YzVmV5hoQUjCEjhrXnQqVDJdvKA==",
+      "dev": true,
+      "dependencies": {
+        "tslib": "^2.1.0"
+      },
+      "engines": {
+        "node": ">=8"
+      }
+    },
+    "node_modules/@apollo/client/node_modules/ts-invariant/node_modules/tslib": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.2.0.tgz",
+      "integrity": "sha512-gS9GVHRU+RGn5KQM2rllAlR3dU6m7AcpJKdtH8gFvQiC4Otgk98XnmMU+nZenHt/+VhnBPWwgrJsyrdcw6i23w==",
+      "dev": true
+    },
+    "node_modules/@apollo/client/node_modules/zen-observable": {
+      "version": "0.8.15",
+      "resolved": "https://registry.npmjs.org/zen-observable/-/zen-observable-0.8.15.tgz",
+      "integrity": "sha512-PQ2PC7R9rslx84ndNBZB/Dkv8V8fZEpk83RLgXtYd0fwUgEjseMn1Dgajh2x6S8QbZAFa9p2qVCEuYZNgve0dQ==",
+      "dev": true
+    },
+    "node_modules/@apollo/protobufjs": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/@apollo/protobufjs/-/protobufjs-1.2.2.tgz",
+      "integrity": "sha512-vF+zxhPiLtkwxONs6YanSt1EpwpGilThpneExUN5K3tCymuxNnVq2yojTvnpRjv2QfsEIt/n7ozPIIzBLwGIDQ==",
+      "dev": true,
+      "dependencies": {
+        "@protobufjs/aspromise": "^1.1.2",
+        "@protobufjs/base64": "^1.1.2",
+        "@protobufjs/codegen": "^2.0.4",
+        "@protobufjs/eventemitter": "^1.1.0",
+        "@protobufjs/fetch": "^1.1.0",
+        "@protobufjs/float": "^1.0.2",
+        "@protobufjs/inquire": "^1.1.0",
+        "@protobufjs/path": "^1.1.2",
+        "@protobufjs/pool": "^1.1.0",
+        "@protobufjs/utf8": "^1.1.0",
+        "@types/long": "^4.0.0",
+        "@types/node": "^10.1.0",
+        "long": "^4.0.0"
+      },
+      "bin": {
+        "apollo-pbjs": "bin/pbjs",
+        "apollo-pbts": "bin/pbts"
+      }
+    },
+    "node_modules/@apollo/protobufjs/node_modules/@types/node": {
+      "version": "10.17.60",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.17.60.tgz",
+      "integrity": "sha512-F0KIgDJfy2nA3zMLmWGKxcH2ZVEtCZXHHdOQs2gSaQ27+lNeEfGxzkIw90aXswATX7AZ33tahPbzy6KAfUreVw==",
+      "dev": true
+    },
+    "node_modules/@apollographql/apollo-tools": {
+      "version": "0.5.0",
+      "resolved": "https://registry.npmjs.org/@apollographql/apollo-tools/-/apollo-tools-0.5.0.tgz",
+      "integrity": "sha512-7IOZHVaKjBq44StXFJEITl4rxgZCsZFSWogAvIErKR9DYV20rt9bJ2mY5lCn+zghfGrweykjLb9g4TDxLg750w==",
+      "dev": true,
+      "dependencies": {
+        "apollo-env": "^0.10.0"
+      },
+      "engines": {
+        "node": ">=8",
+        "npm": ">=6"
+      }
+    },
+    "node_modules/@apollographql/graphql-playground-html": {
+      "version": "1.6.26",
+      "resolved": "https://registry.npmjs.org/@apollographql/graphql-playground-html/-/graphql-playground-html-1.6.26.tgz",
+      "integrity": "sha512-XAwXOIab51QyhBxnxySdK3nuMEUohhDsHQ5Rbco/V1vjlP75zZ0ZLHD9dTpXTN8uxKxopb2lUvJTq+M4g2Q0HQ==",
+      "dev": true,
+      "dependencies": {
+        "xss": "^1.0.6"
+      }
+    },
+    "node_modules/@apollographql/graphql-upload-8-fork": {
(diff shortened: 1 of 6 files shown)


fn-faisal:
I will revert the package-lock.json but I'd need to add sweet alert dependency.

mtrezza:
Can you please rename the title accordingly?

fn-faisal:
> Thanks for this PR.
> 
> Can you please rename the PR title accordingly?

I have updated the PR title.

mtrezza:
Thank you.

There are some issues that come to mind:
- Does each cell trigger a separate query to fetch the representation, or is that done together with the paging mechanism as a bulk-fetch?
- Where is the information about the pointer key stored? Is it sustainable enough?

fn-faisal:
> Thank you.
> 
> * Does each cell trigger a separate query to fetch the representation, or is that done together with the paging mechanism as a bulk-fetch?
> * Where is the information about the pointer key stored? Is it sustainable enough?
> * Do we need the SweetAlert component? We already have existing components to achieve the same.

- For each cell, if the type is pointer and the pointer key is not objectId, it fetches the object using the get method of the Parse Query object.
- To store pointer data, localStorage is used to store Class name as key and the key as value.
- A similar PR was made using b4a dash and it has sweet alert so I went with it and this PR is based off of that one, but I can revert this to Parse Dashboard specific alter dialogue.

fn-faisal:
I've pushed a commit that will optimize the query. The second query has been removed and instead I've added include and select on the first query that fetches the data. I'll update the dialog in the next commit.

fn-faisal:
Screenshot 2021-06-05 at 12 03 27 AM

The dialog has been updated and the sweet alert2 dependencies have been removed.

fn-faisal:
Screenshot 2021-06-05 at 12 08 52 AM

The dialogue to change pointer value has been updated. The sweet alert2 deps have been removed.

mtrezza:
Great, kindly let me know when this is ready for review.

fn-faisal:
I've cleaned up the code and the PR is ready to be reviewed.

src/dashboard/Data/Browser/PointerKeyDialog.react.js
```diff
@@ -0,0 +1,72 @@
...
+            </Dropdown>
+          } />
+      )
+    }
+    return (
+      <Modal
+        type={Modal.Types.INFO}
+        icon='warn-outline'
+        title={'Change pointer value for class '+this.props.className}
+        subtitle={hasColumns ? 'The column will be used throughout dashboard as pointer value for this class.' : 'There are no columns on this class that can be set a pointer value.'}
```
mtrezza, line 55: Fix layout issue where text overlays icon:

image

mtrezza, line 55: Rename `pointer value` to `pointer key`. We should use the same term throughout code and UI.

src/dashboard/Data/Browser/PointerKeyDialog.react.js
```diff
@@ -0,0 +1,72 @@
...
+        confirmText='Update pointer value'
```
mtrezza, line 56: Rename `pointer value` to `pointer key`. We should use the same term throughout code and UI.

mtrezza (changes requested):

src/dashboard/Data/Browser/PointerKeyDialog.react.js
```diff
@@ -0,0 +1,72 @@
...
+              {this.props.currentColumns.map((t) => <Option key={t} value={t}>{t}</Option>)}
+            </Dropdown>
+          } />
+      )
+    }
+    return (
+      <Modal
+        type={Modal.Types.INFO}
+        icon='warn-outline'
+        title={'Change pointer value for class '+this.props.className}
```
mtrezza, line 54: Remove class name from dialog, due to possible layout issues with large class names; in stead use:

> Change pointer key for this class

mtrezza (changes requested):

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -235,18 +234,38 @@ export default class BrowserCell extends Component {
       content = <span>&nbsp;</span>;
       classes.push(styles.empty);
     } else if (type === 'Pointer') {
+      const defaultPointerKey = await localStorage.getItem(value.className) || 'objectId';
+
+      let dataValue = value.id;
+      if ( defaultPointerKey !== 'objectId' ) {
+        dataValue = value.get(defaultPointerKey);
+        if ( dataValue && typeof dataValue === 'object' ){
```
mtrezza, line 242: We need something else than `Invalid Value` in case the pointer key is set to a field that is `(undefined)`.

image

I suggest to display `(undefined)` which is the true current value of the pointer key, or maybe you have a different idea.

mtrezza (changes requested):

src/components/BrowserCell/BrowserCell.react.js
fn-faisal: I can replace the square brace but from my understanding of the dashboard, the undefined is not clickable. The idea that I had while implementing this was that if there's an issue while loading the pointer key value, display something but still leave the field clickable to allow users to navigate the between the pointer relation.
While undefined indicates that there is no or the row doesn't point to any row in the respective table.

src/dashboard/Data/Browser/PointerKeyDialog.react.js
fn-faisal: Screenshot 2021-06-07 at 5 55 17 PM

fn-faisal:
I have made the changes and code is ready to be reviewed again.

src/components/BrowserCell/BrowserCell.react.js
mtrezza: The word "invalid" has a different meaning than "undefined". If a pointer key value is "undefined", we should convey that it is simply "undefined". But we also cannot display it the same as `(undefined)`, which means that no value is set for the field.

Since we will abandon the pill design with https://github.com/parse-community/parse-dashboard/pull/1708 I think we end up with these ambiguations:

- a) `(undefined)` = no value is set for field
- b) `undefined ↗️` = pointer field value is defined as string "undefined"
- c) `(undefined) ↗️` = pointer field value is not defined

This means that a pointer key value that is undefined can only be displayed as (c).

mtrezza:
Thanks, can you please take a look at https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r655729128

src/components/BrowserCell/BrowserCell.react.js
fn-faisal: Ok, I'll update it to match the PR branch #1708 and update this PR to match (c)

mtrezza: I think we should merge the pointer redesign PR first, then merge master into this PR?

mtrezza:
@fn-faisal It would be great if we could merge this amazing PR. Would you mind merging this with master to resolve any conflicts? And do you think we should merge the pointer redesign PR first, and merge this PR afterwards?

fn-faisal:
@mtrezza I think it would we should merge the pointer PR first before merging this PR. I haven't yet caught this PR with the pointer PR yet.

mtrezza:
Ok, I will ping you here once the other PR has been merged.

mtrezza:
@fn-faisal #1708 has been merged. Could you rebase this PR on master to merge it next?

fn-faisal:
@mtrezza I updated the PR but running it locally, I encountered bugs that I updated for Pill when displaying array of pointers etc. The PR is ready for review.

mtrezza:
> I encountered bugs that I updated for Pill when displaying array of pointers and cell section/edit value

Do you mean you fixed bugs that were not part of this PR? If so, can you point out which changes these are?

fn-faisal:
@mtrezza These were new issues.

1. The cell border that appears when clicking on a cell wasn't appearing but the value was copyable.
2. The data from the edit field that appears, when double clicking on a cell and entering a new value and then clicking away to save, wasn't appearing in the cell.
3. The array of pointers was not rendering properly with the new pill design.

src/components/BrowserCell/BrowserCell.react.js
fn-faisal: I have also updated the PR with the new pointer design.

mtrezza:
Normally, bugs that are not part of the PR would require a separate PR. But I think in this case we can make an exception and reduce the work because we have a couple of PRs to merge that make changes to the same files and already have undergone multiple rebasing.

fn-faisal:
@mtrezza I encountered the bugs while resolving merge conflicts. They were not present in the master branch.

mtrezza:
Got it, in that case they can be considered part of this PR. Thanks for clarifying, and for the attention in finding these bugs. I will take a look later if I find any other side effects.

mtrezza:
Could you just rebase on the main branch? Apologies for all the rebasing, we just have so many great PRs that are being merged.

fn-faisal:
> Could you just rebase on the main branch? Apologies for all the rebasing, we just have so many great PRs that are being merged. This PR will be merged next if all goes well.

I have resolved the merge conflicts for this PR.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -1432,7 +1489,17 @@ class Browser extends DashboardView {
       }
     }
     let extras = null;
-    if (this.state.showCreateClassDialog) {
```
mtrezza, line 1435: Is this change intended?

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -23,12 +23,149 @@ export default class BrowserCell extends Component {
...
+        }
+        content = isNewRow && this.props.isRequired && this.props.value === undefined ? '(required)' : content;
+      } else if (this.props.value === null) {
+        this.copyableValue = content = '(null)';
+        classes.push(styles.empty);
+      } else if (this.props.value === '') {
+        content = <span>&nbsp;</span>;
+        classes.push(styles.empty);
+      } else if (this.props.type === 'Pointer') {
+        const defaultPointerKey = await localStorage.getItem(this.props.value.className) || 'objectId';
```
mtrezza, line 56: I assume there is currently no caching mechanism, correct? So if the same pointer is used with the same objectId in multiple rows, there will be a DB request for each cell? I would not make this is a requirement for this PR, but I think that would be one of the possible improvements we could mention in the top post of this PR, so people are aware of this caveat.

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -329,16 +480,32 @@ export default class BrowserCell extends Component {
...
+    let classes = [...this.state.classes];
+
+    if ( current ) {
+      classes.push(styles.current);
+    }
     if (markRequiredFieldRow === row && isRequired && !value) {
       classes.push(styles.required);
     }
 
+    // console.log(value, this.state.value);
```
mtrezza, line 505: Can this be removed?

mtrezza (changes requested):

src/components/BrowserCell/BrowserCell.react.js
fn-faisal: I have removed the console.log

src/components/BrowserCell/BrowserCell.react.js
fn-faisal: Yes, it'll perform a DB operation for each row. I'll mention it on the PR.

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -237,20 +373,36 @@ export default class BrowserCell extends Component {
       content = <span>&nbsp;</span>;
       classes.push(styles.empty);
     } else if (type === 'Pointer') {
+      const defaultPointerKey = await localStorage.getItem(value.className) || 'objectId';
+
+      let dataValue = value.id;
+      if( defaultPointerKey !== 'objectId' ) {
+        dataValue = value.get(defaultPointerKey);
+        if ( dataValue && typeof dataValue === 'object' ){
+          if ( dataValue instanceof Date ) {
```
mtrezza, line 382: This looks like strange formatting with the space inside the brackets. Do we run lint over this usually?

mtrezza:
Can you please also add a shot feature documentation including any [limitations](https://github.com/parse-community/parse-dashboard/pull/1723/files#r688584766)? We have a new section in README.md:

https://github.com/parse-community/parse-dashboard/blob/master/README.md#features

mtrezza:
I assume your [comment](https://github.com/parse-community/parse-dashboard/pull/1754#issuecomment-899504959) was for this PR?

mtrezza:
@fn-faisal I slightly changed the feature description, can you please look over it whether I got it right?

fn-faisal:
@mtrezza looks good to me 👍

mtrezza:
Actually I just realized that the example in the description doesn't work, it's not possible to use the `email` field of user as pointer key. Why is that?

Another question, when changing the pointer key back to `objectId`, does that remove any custom pointer key setting or does it still trigger server requests to resolve the poiner key?

fn-faisal:
@mtrezza The default system keys aren't passed to the pointer keys dialog. I added the objectId manually. If the email is a key not created by the user, it may be getting excluded from the columns passed to the pointer keys dialog.

Changing the key back to pointer key just sets the pointer key in localStorage back to objectId and I think the browser cell does not fetch for objectId.

mtrezza:
> I think the browser cell does not fetch for objectId.

Can you change this to make sure there is no fetching? If it still fetches that would be a bug. Maybe just completely delete the pointer key entry from the browser storage.

fn-faisal:
@mtrezza from the below screen shots, it looks like there's not fetch for pointer key. The pointer key includes the Parse object on which the value is picked from using the `.get` method. In the image below, the pointer key is not objectId and there's no extra network call.
Screenshot 2021-08-16 at 9 55 22 PM

src/dashboard/Data/Browser/PointerKeyDialog.react.js
mtrezza: This is marked as resolved, but the change has not been made.

mtrezza:
Yes, I can see there is a condition in code that doesn't fetch if the the pointer key is set to `objectId`, so we should be good.

src/components/BrowserCell/BrowserCell.react.js
```diff
@@ -23,12 +23,149 @@ export default class BrowserCell extends Component {
...
-    }
-
+      showTooltip: false,
+      content: null,
+      classes: []
+    };
   }
 
-  componentDidUpdate(prevProps) {
+  async componentDidUpdate(prevProps) {
```
mtrezza, line 32: What are these changes? They seem not to be part of this PR?

fn-faisal: The value of pointer key column is fetched from local storage using await, that's why async is added here.
https://github.com/fn-faisal/parse-dashboard/blob/feat/151-pcd/src/components/BrowserCell/BrowserCell.react.js#L376

mtrezza: I actually mean the many changes in this method:

https://github.com/parse-community/parse-dashboard/pull/1723/files?file-filters%5B%5D=.js#diff-4e4d2ae32c4431bc66b52e869432797428508c6a3cb0811018e658525f0637c9R33-R168

fn-faisal: I moved the changes from the component's render method to the did update method. I was having issues with async loading of data ( like key from local storage ).

mtrezza: When you say "moved", I don't see nearly the same amount of lines removed from the `render` method. That looks strange to me.

fn-faisal: Oh, sorry, I copied cell rendering logic to re-render on component updates for async calls. But this is for component updates. The lines in the render are for the initial render.

mtrezza: Do you mean that you copied these lines from somewhere else? From where?

fn-faisal: The lines are what render the browsercell based on the type of data passed. When the key is fetched from localstorage, it fetches and renders the pointer key's value. These lines are also present in the render function but since we need async calls, the lines are also partially copied to the componentDidUpdated method.
Previously, the component wasn't using states but I store the cell data, once fetched from pointer key, to state upon the resolution of promise callbacks and the component's DidMount function updates the cell value accordingly.

mtrezza: > These lines are also present in the render function 

Can you post a link to that render function that contains these lines?

fn-faisal: Before it used to simply get the cell values from props and no async calls were place.
https://github.com/parse-community/parse-dashboard/blob/1ddf29e84404ed36db86fc496fab82145add8bb1/src/components/BrowserCell/BrowserCell.react.js#L217

I updated it to store cell values in state so when the pointer key column name is fetched from localstorage and the value ,for this column, is fetched using js SDK .get method, the component state is updated with the value.
https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r690610276

The lines in the render function that check the data type and show how the cell should be rendered are present in render function (I wanted to clean it up too by moving the lines to a separate function and call it in the didMount and didUpdate component lifecycle function function but didn't get around to it)

These lines check the cell data type and how it should be rendered (I needed to re-render the cell data once it's fetched from the localStorage+SDK so I had to copy them to didUpdate function for when the cell data in the component state updates).
This is where the logic of how the cell should renders starts.
https://github.com/parse-community/parse-dashboard/blob/1ddf29e84404ed36db86fc496fab82145add8bb1/src/components/BrowserCell/BrowserCell.react.js#L222

and ends around this line
https://github.com/parse-community/parse-dashboard/blob/1ddf29e84404ed36db86fc496fab82145add8bb1/src/components/BrowserCell/BrowserCell.react.js#L330

mtrezza: > I wanted to clean it up too by moving the lines to a separate function and call it in the didMount and didUpdate component lifecycle function function but didn't get around to it

Thanks for explaining this in detail, I understand now. I think it would be really great if you could do the clean-up before we merge this PR. Duplicate code is often difficult to maintain, if someone later makes a change only in 1 part of code, but not in the other.

mtrezza:
@fn-faisal Please feel free to request a review if you think this is ready

mtrezza:
@fn-faisal Could you please resolve the conflicts for review?

mtrezza:
This is almost ready, the only thing I found was an issue if the pointer key is set to another pointer column or a file column.


https://user-images.githubusercontent.com/5673677/132421198-99971e1f-abec-4040-8d95-7bc335a2a2a8.mov

mtrezza:
Does this work on a per-app basis?

I am thinking, if I have multiple tabs of different dashboards open in the same browser session, would the pointer column settings influence each other?

fn-faisal:
@mtrezza I will update it to work for per app basis and display ID if the shadow value is undefined.

fn-faisal:
@mtrezza I had a suggestion about the pointer cell. What if an icon is displayed for the column cell that uses objectId and icon isn't shown for shadow values. 

Screenshot 2021-09-13 at 7 11 21 PM

fn-faisal:
@mtrezza I have added appId along with className to save pointerKey + if the shadow value is null, now the cell displays pointer value instead.

Parse-Dashboard/public/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <script>
+        PARSE_DASHBOARD_PATH = "/";
+    </script>
+</head>
+<html>
+<title>Parse Dashboard</title>
+<body>
+<div id="browser_mount"></div>
+<script type="text/javascript">
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
mtrezza, line 15: Can you please remove this?

mtrezza (changes requested):

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
mtrezza, line 1: Can you please remove this and instead add this to `.gitignore`:

```
# visual studio code
.vscode
```

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "parse-dashboard",
-  "version": "3.0.0",
+  "version": "2.2.0",
```
mtrezza, line 3: Can you please merge master into this branch? This should be based on master.

fn-faisal:
@mtrezza I have removed `.vscode`, updated `.gitignore` and  reverted `package-lock.json`

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ Details:
 - Purpose: official release
 - Suitable environment: production
 
+## New Features
```
mtrezza, line 18: Can you remove any change to the changelog? changelog entry will be generated automatically from now on based on the pull request title

src/components/Button/Button.example.js
```diff
@@ -10,6 +10,7 @@ import Button from 'components/Button/Button.react';
 
 export const component = Button;
 
+
```
mtrezza, line 13: can you remove this newline?

CHANGELOG.md
fn-faisal: I have removed the entry from CHANGELOG.md file

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -657,6 +684,30 @@ class Browser extends DashboardView {
       query.ascending(field)
     }
 
+    if (field !== 'objectId') {
```
mtrezza, line 687: This seems to be a merge issue; this code does not exist in the master branch anymore, see https://github.com/parse-community/parse-dashboard/pull/1800.

I am worried that there is some code in this PR that is obsolete but is still here due to master merging. Can you please take a minute and look through tall the change in this PR, compare it to the current master, and see if all the changes are related to your PR? Sorry for the extra work, but we better take an extra look before we introduce a bug.

mtrezza (changes requested):

mtrezza:
It seems there is an issue in the following case, given that:

- the `User` class has field `installation` that is a pointer to the `Installation` class
- the `Installation` class has field `user` that is a pointer to the `User ` class

Steps to reproduce:
1. In `User` class set pointer key to `installation` field
2. In `Installation` class refresh data table
3. Note that In `Installation` class the field values in column `user` show `(undefined)`, but should show the installation's `objectId`.

fn-faisal:
@mtrezza For the first issue regarding the pointers, I tried in the following video but could not reproduce the issue.

Uploading Screen Recording 2021-09-29 at 8.22.15 PM.mov…

fn-faisal:
@mtrezza For the first issue regarding the pointers, I tried the steps above in the following video but could not reproduce the issue.

https://user-images.githubusercontent.com/23558802/135311902-e1e7249d-f1f3-4e0a-820c-e6d5e4295328.mov

mtrezza:
Thanks, I'll take another look. Maybe you can take a look [this](https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r715015482) meanwhile.

mtrezza:
To make sure it doesn't get overlooked, the issue review is on hold because of https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r715015482

fn-faisal:
@mtrezza I've reverted the changes for comment [#1723 (comment)](https://github.com/parse-community/parse-dashboard/pull/1723#discussion_r715015482)

mtrezza:
Great, did you look through the whole PR if there are any other modifications that are not part of this PR?

fn-faisal:
@mtrezza I went through the PR and it only seems to have changes related to pointer key issue.

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -644,6 +648,29 @@ class Browser extends DashboardView {
     }
   }
 
+  setRequiredColumnFields() {
```
mtrezza: Is this also part of this PR?

fn-faisal: I have removed this function. It was not being used anywhere and may have got in during a merge conflict resolution.

mtrezza:
🎉 This pull request has been released in version [3.3.0-alpha.2](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0-alpha.2)

Issue #1729, opened 2021-06-15
Copy-paste of date data does not work


fn-faisal:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.

- [x] You're running version >=2.3.2 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce
When you click on a date type column, the copy pasting causes issues due to the popup. It would be easier to if there was an option to paste clipboard date to column value.

mtrezza:
Thanks for opening this issue.

Can you describe in more detail what the issue is? It is currently possible to double click into a date field and paste a date string, without using the calendar popup.

fn-faisal:
It does work but it causes confusion with the popup. If there's an option within the popup to also paste the date from clipboard, the user would be able to manage that from the calendar too without having to double click + erase existing date and then paste new date.
And since the button would only appear if there's a date string on clipboard, it can make copy pasting quicker.

mtrezza:
> If there's an option within the popup to also paste the date from clipboard, the user would be able to manage that from the calendar too without having to double click + erase existing date and then paste new date.

Selecting a cell's content and pasting from the clipboard is a common gesture. Are you suggesting to add a "Paste" button to the calendar pop-up to paste the clipboard content?

mtrezza:
Closing due to inactivity.

PR #1730, opened 2021-06-15
Copy-paste of date data does not work #1729

fn-faisal:
Upon double clicking a date column to edit, the field's date text is selected to make copy pasting of date data easier.

TODO:
- [x] Add Changelog entry


mtrezza:
Thanks for this PR.

Can you explain the benefit of this button is? I don't think we have a paste button for any other field anywhere in the dashboard. Why is it beneficial that specifically the date field has one?

fn-faisal:
@mtrezza while trying to paste date, when you double click and go into edit mode, it's hard to paste the date. Just pressing `ctrl`+ `v` just appends to the date. You have to select the entire date with `ctrl` + `a` and delete the existing date and then paste the new date with `ctrl` + `v`.

This button gives the user ease to paste date from clipboard. It only appears if 

1.  The dashboard has access to the clipboard.
2.  There's a date string on the clipboard.

otherwise the button is hidden.

mtrezza:
> @mtrezza while trying to paste date, when you double click and go into edit mode, it's hard to paste the date. Just pressing ctrl+ v just appends to the date. You have to select the entire date with ctrl + a and delete the existing date and then paste the new date with ctrl + v.

Isn't that the case with any field? In fact, isn't that the expected default behavior for any spreadsheet application or table control?

fn-faisal:
@mtrezza I've reverted the old logic ( Copy+paste button ) and updated the PR

mtrezza:
Can you please revert this logic:

image

It prevents a user from double clicking into a date field and selecting only a part of the date.

fn-faisal:
> This pop-up and selection restriction is a rather unintuitive UI behavior. If a cell is read only, then the text field can be set as disabled, which makes it selectable but not editable. We should not prevent the user from selecting the date string in that field. It prevents a user from double clicking into a date field and selecting only a part of the date.
> 
> Can you please revert this restriction and remove this pop-up:
> 
> image
> 
> The PR was good as it was before, just instead of introducing a new UI button, simply pre-select the date string also in custom date fields. This way you only have to crtl+v to paste the date string, or ctrl+c to copy it. Then we can merge this.

I set it to track the master branch ( parse-community/parse-dashboard ) in my local branch. This reverts the copy+paste logic and the PR only contains 1 file change.

I pulled from master and updated the PR.

I set this PR up to track the master branch but changes were pushed since then. Was this something recently implemented?

mtrezza:
Oh, you are right, this is on master, sorry, let me look into this...

mtrezza:
I just tried it out and it looks like this to me:

https://user-images.githubusercontent.com/5673677/130305940-7e72bdf3-4a14-4bc4-9b30-815db6971867.mov

I cannot select the date string, because after double click: nothing is selected; after another double click, it selects and un-selects. I can only select everything by pressing crtl+a.

fn-faisal:
@mtrezza I have updated the PR with a fix for this.

mtrezza:
Thanks, can you add the changelog entry?

fn-faisal:
I have updated the change log.

mtrezza (approved):
Thanks for this PR!

Issue #1731, opened 2021-06-15
Message error for required fields


fn-faisal:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.

- [x] You're running version >=2.3.2 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce
If I have a field called "test", the message error shows "Test is required" with the letter T in uppercase (see screenshot)

![Captura de Tela 2021-06-01 às 19 50 39 (2) (1)](https://user-images.githubusercontent.com/23558802/122051283-dc37da80-cdfd-11eb-9f42-428c4261ae90.png)

mtrezza:
I am wondering whether
a) dashboard should re-format the error message, or
b) Parse Server should send more readable error messages

I tend to think (b), because we are lacking a style guide for error messages in Parse Server and now Dashboard tries to correct this, but it seems like addressing the symptom instead of the issue. Error messages appear in a variety of forms:
- beginning upper case or lower case
- with or without trailing dot (.)
- with or without field quotes (e.g. `'test' is required`)

And in general, the wording comes in all shapes and forms.

fn-faisal:
In the current implementation, I see in the dashboard 
`
let msg = typeof error === 'string' ? error : error.message;     
if (msg) {        
  msg = msg[0].toUpperCase() + msg.substr(1);      
}`

The first letter of the all the error messages is currently being capitalised. 
So, I think for (b), this would need to be changed in the dashboard too.
The implementation of (b) makes sense.

mtrezza:
Would you want to tackle (b)? It is surely quite an undertaking, but I think it would be a sustainable fix and improve the quality of error messages in Parse Server.

fn-faisal:
Implementing (b) would make sense as it'll delegate the responsibility to show an error dialog and the server to send reformatted error strings.

mtrezza:
So we can close this issue?

fn-faisal:
I think there's already existing formatting in the dashboard as well which would need to be removed.
I can update this PR or make a new one.

fn-faisal:
[...]

PR #1732, opened 2021-06-15
lowercase column error toast

fn-faisal:
This is a fix for the issue I created earlier [issue link](https://github.com/parse-community/parse-dashboard/issues/1731)

The fix implemented for this is to not convert the server error response to first letter uppercase if the error code is 142 (validation error).


mtrezza:
@fn-faisal Can we close this PR? I think https://github.com/parse-community/parse-server/pull/7446 addresses this on the server side?

Issue #1716, opened 2021-05-27
Un-hidden columns contain empty fields


mtrezza:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.
- [x] You're running version >=2.3.2 of Parse Server.
- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

## Issue
The PR https://github.com/parse-community/parse-dashboard/pull/1694 excluded hidden columns from being fetched from the server. When showing previously hidden columns, the values for the un-hidden fields are `(undefined)`, because their values have not been fetched. Clicking the "Reset" button loads the fields.

## Instructions

1. Hide a column with values
2. Click "Reload"
3. Un-hide column --> fields are empty
4. Click "Reload" --> fields contain values

sadakchap:
Thanks @mtrezza for detailed instruction to reproduce the issue. But, I still could not reproduce the issue. could you please check below gif & tell me which step I am missing ?

#### Steps I followed - 
1. Hide a column with values
2. Hit refresh button
3. Un-hide that column --> still can see the values

![unhideCols](https://user-images.githubusercontent.com/44117648/119995202-f8a8da00-bfea-11eb-8da2-1794d1d1db60.gif)

mtrezza:
Could you try to create a new column and it with that? Maybe it only happens for certain columns.

sadakchap:
Still was not able to reproduce the issue.

mtrezza:
This is what I see on Safari. Note that without refreshing, the column contains data after un-hiding it. But after refreshing, it does not contain data when un-hiding it. I am using branch https://github.com/parse-community/parse-dashboard/pull/1697/commits/2a4479a90d37cf930b01436f64a0b062d2331606.

I think the simple question is, if columns a not fetched when they are hidden, how would the data show in the column after un-hiding it? Or does un-hiding a column trigger a query?


https://user-images.githubusercontent.com/5673677/120074213-171ade00-c09c-11eb-99aa-c5926c8db222.mov

fn-faisal:
[...]

PR #1734, opened 2021-06-15
Exclude hidden fields 

sadakchap:
Get column preference key correctly from local storage to exclude hidden fields values


mtrezza:
Is this ready for review?

sadakchap:
@mtrezza, thanks for waiting. This pr solves issue https://github.com/parse-community/parse-dashboard/issues/1716.

https://user-images.githubusercontent.com/44117648/127798381-01a01b25-4911-4573-91a7-dff3e128f591.mp4

sadakchap:
@mtrezza, I believe that this pr this ready for a review.

mtrezza:
Thanks for your patience, yes, I am looking at it right now.

mtrezza:
Great, it fixes the issue. I notice that it also fetched data from server for
a) fields that still contains data: hide a field, then show the same field
b) fields that are always sent from server, e.g. `createdAt`, `updatedAt`, `objectId`, `ACL`
c) password field in user, which is always empty

Can we optimize this, at least (b)?

sadakchap:
> I notice that it also fetches data from server for:
> 
> a) fields that still contains data (hide a field, then show the same field)
> b) fields that are always sent from server, e.g. `createdAt`, `updatedAt`, `objectId`, `ACL`
> c) password field in `_User`, which is not transmitted
> 
> Can we optimize this, at least (b)?

Sure we can, could you please first clarify these points ?

a) do you mean, _when it is making request  everytime we un-hide any field without hitting refresh btn_. If yes then, I did thought of making request only if we can not find that field's value in all of the rows. But, it is also possible that all rows contain `undefined` for that field and we will make another request just to get back `undefined` from server.
b) I thought of using `include` query, but it seems like server always sends back these `createdAt`, `updatedAt`, `objectId` fields. 
c) do you mean, when it makes request for unhiding `password` field ?

mtrezza:
Sure, let me try to clarify...

a) the dashboard should not fetch data from server, if a field is hidden and then un-hidden. When a field is hidden, the data of the field is still cached, unless the user hits the refresh button between hiding and un-hiding.

b) the dashboard should not fetch data from server if any of the createdAt, updatedAt, objectId, ACL fields are un-hidden, because these fields are always cached, because they are always fetched from server and never excluded.

c) the dashboard should not fetch data from server if the password field is un-hidden, because the password field is never sent from server as a security precaution, so there is no point is fetching data.

mtrezza:
@sadakchap Do you want to address these issues in this PR, or should we merge and leave this as potential improvement in the future? The PR is already an improvement as it is.

sadakchap:
Thanks for your patience @mtrezza . I am trying to implement suggested changes. I think it would be good to merge this pr after this changes. :)

mtrezza:
Great! Thanks for looking into this.

sadakchap:
@mtrezz, Thanks for waiting. I think now it is ready for a review.

sadakchap:
@mtrezza , Thanks for waiting. I think now it is ready for a review.

mtrezza:
Looks good, just one minor thing:
1. Hide all columns
2. Click on "refersh"
3. Click on "Manage columns" > "Show all" --> doesn't load the empty columns

Could you fix that? That's about the only bug I found.

sadakchap:
> Could you fix that? That's the only bug I found, then we should be good for merge.

Sure, here it is https://github.com/parse-community/parse-dashboard/pull/1734/commits/95b02e7b8c362edbaec70b90f3d0155e75566656!

mtrezza (approved):
Works great, thanks for this feature!

PR #1741, opened 2021-07-03
updated issue and PR templates

mtrezza:
- aligned templates with parse-server repo templates
- added changelog entry as PR TODO to avoid effort of creating changelog on release


TomWFox:
Thanks for the PR Manuel!

I know we haven't actually included them yet but it's probably worth incorporating the [wording changes we discussed](https://github.com/parse-community/.github/issues/6)...

```
<!--
    Thanks for contributing to the Parse Platform!

    Please only submit your issue if ALL of the below are true. Check each box [x] to indicate this to our triage team:
    Click the "Preview" tab for better readability.
-->

- [ ] This is not a vulnerability disclosure. If it is, [report here](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
```

mtrezza:
I don't agree with the verbose wording but I added an emphasis to check `every` box, which hopefully helps.

TomWFox:
@mtrezza did you intend to merge this without review?

mtrezza:
I incorporated your comment.

TomWFox:
Right but that was a comment, not a review of the PR. You also didn't give me a chance to review the changes you made which I would expect especially given the changes were quite different to the suggestion.

TomWFox:
Having reviewed the PR I'm ok with it remaining as is.

My only concern is with the more forceful approach towards making a PR with failing test...
```
### Failing Test / Pull Request
<!--
    Please submit a PR with a failing test (or passing test if you already include the fix), so we can verify that the issue exists.
-->

- PR with [test](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#environment-setup): #`FILL_THIS_OUT`
```
It could be beneficial in reducing maintainer time spent looking into issues and potentially weed out more non-issues. The negative could be a higher number of people abandoning during issue creation although it seems people frequently just ignore parts they can't/don't want to do so hopefully not a significant issue.

More broadly I'm not sure whether we want to be sending the message that making a PR with a failing test is the minimum barrier to entry when we are aiming to make the project more beginner friendly.

We can see how it works in practice but it would be interesting to here any thoughts from @parse-community/core-maintainers.

TomWFox:
@mtrezza In future please wait for an approval before merging.

Issue #1753, opened 2021-08-05
the bold border is not covering the whole cell


fn-faisal:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases).
- [x] I can reproduce the issue with the [latest version of Parse Dashboard](https://github.com/parse-community/parse-dashboard/releases).

### Issue Description
The border around the cell is no longer covering the whole cell when selecting a cell.
<img width="720" alt="Screenshot 2021-08-05 at 6 48 30 PM" src="https://user-images.githubusercontent.com/23558802/128361495-9303c841-2803-426e-8c6b-d4b65eaaa2ae.png">

### Steps to reproduce
Open a table in Database browser and click a cell. The border should appear but it does not cover the entire cell.

### Actual Outcome
The border appears smaller than the actual cell height.

### Expected Outcome
Upon clicking on a cell, the border should match the height of the column and cover the entire cell.

### Pull Request
https://github.com/parse-community/parse-dashboard/pull/1754

Dashboard
- Parse Dashboard version: `2.1.0`
- Browser (Safari, Chrome, Firefox, Edge, etc.): `Chrome, Safari`
- Browser version: `92.0.4515.131`

mtrezza:
Closed via #1754

PR #1754, opened 2021-08-05
cell select height fix

fn-faisal:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues/1753).

### Issue Description
The border around the cell is no longer covering the whole cell when selecting a cell.

### Approach
The height of the row is different from the height of the cell which causes the small border. Increasing the height of cell and fixing the height of row to match the cell fixes the border issue.

Screenshot 2021-08-05 at 7 05 24 PM

### DIFF
--- a/src/components/BrowserCell/BrowserCell.scss
+++ b/src/components/BrowserCell/BrowserCell.scss
@@ -13,7 +13,7 @@
   color: #0E69A1;
   height: 30px;
   // line-height: 22px;
-  padding: 6px 4px 0;
+  padding: 5px 4px 0;
   border-right: 1px solid #e3e3ea;
 }
 
--- a/src/components/BrowserCell/BrowserCell.scss
+++ b/src/components/BrowserCell/BrowserCell.scss
@@ -11,7 +11,7 @@
   text-overflow: ellipsis;
   cursor: default;
   color: #0E69A1;
-  height: 32px;
+  height: 30px;
   line-height: 22px;
   padding: 5px 4px 0;
   border-right: 1px solid #e3e3ea;
--- a/src/components/BrowserCell/BrowserCell.scss
+++ b/src/components/BrowserCell/BrowserCell.scss
@@ -13,7 +13,7 @@
   color: #0E69A1;
   height: 30px;
   // line-height: 22px;
-  padding: 5px 4px 0;
+  padding: 6px 4px 0;
   border-right: 1px solid #e3e3ea;
 }
 
--- a/src/components/BrowserCell/BrowserCell.scss
+++ b/src/components/BrowserCell/BrowserCell.scss
@@ -12,7 +12,7 @@
   cursor: default;
   color: #0E69A1;
   height: 30px;
-  line-height: 22px;
+  // line-height: 22px;
   padding: 5px 4px 0;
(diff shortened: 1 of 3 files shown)


mtrezza:
It seems the issue is actually that the row is too high, because the text is not vertically centered. So the proper solution would be to reduce the row height. Otherwise the rows seem quite large, which would reduce the number of visible rows per screen.

Can you adapt the fix to reduce the row height to the same as the cell height?

fn-faisal:
@mtrezza I have updated the row height from auto to 31px (same as the cell height). The ROW_HEIGHT was set to 31 but it wasn't reflecting the height of the row. In css, it was set to auto.

mtrezza:
It looks better now, but could you take another look, the bottom margin is still 1px (or 2?) to large:

image

fn-faisal:
> It looks better now, but could you take another look, the bottom margin is still 1px to large. In other words, the cell is still too high:
> 
> image

@mtrezza I have updated the PR.

mtrezza:
Thanks, something is still wrong here, it seems that the content is not centered in all cells equally.

If you take a look at the difference between the first column (`objectId`) and the second column (a column with a pointer), the text is not on the same height:

image

fn-faisal:
> Thanks, something is still wrong here, it seems that the content is not centered in all cells equally.
> 
> If you take a look at the difference between the first column (`objectId`) and the second column (a column with a pointer), the text is not on the same height:
> 
> image
> 
> In the first, non-pointer column the text should be 1 px higher, without changing the cell height.

@mtrezza I have looked into this and it seems like there's a line-height for the cells but the pointer's new layout isn't effected by the line height. I can remove the line-height and increase padding in the cells to compensate the line height.

fn-faisal:
After removing the `line-height` this is what the layout looks like now.
Screenshot 2021-08-11 at 6 29 40 PM

mtrezza:
The text on the left is still 1px higher than the text on the right, could you take another look?

fn-faisal:
@mtrezza I have updated the Features list.

fn-faisal:
@mtrezza I have updated the PR. Here's what the cells looks like now.
Screenshot 2021-08-16 at 7 43 24 PM
Screenshot 2021-08-16 at 7 43 34 PM

mtrezza:
Looking at your screenshot, both texts are new 1 px too high and the pointer icon is cropped off. I suggest you revert the last commit and just try to lower the text of in left field, as I described in my previous [comment](https://github.com/parse-community/parse-dashboard/pull/1754#issuecomment-898721920).

fn-faisal:
@mtrezza I've updated the PR. Here's the new snapshots.
Screenshot 2021-08-16 at 9 36 54 PM

mtrezza:
We are almost there, here is a zoomed in screenshot:

image

The text needs to move up 1 px.
The icon on the left needs to move up by a few pixel.

fn-faisal:
@mtrezza I've updated the PR
Screenshot 2021-08-17 at 5 31 41 PM

mtrezza:
I have fixed the height. For future reference, the structure is simple:

row height is 30px
cell has padding 5px
cell height therefore is 20px

And all is centered.

PR #1757, opened 2021-08-17
Open pointer in new browser tab with [Cmd] click

### LINKED ISSUE #1711: Open pointer in new browser tab with [Cmd] click
## Feature Suggestion

Open a pointer in a new browser tab when clicking on it while holding the [Cmd] button.

The purpose is that while the dashboard already allows to follow a pointer and jump back with the "Back" button, the previous page is newly fetched from the server and scrolled to top. So this type of navigation cannot be practically used to follow multiple pointers.

fn-faisal:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues/1711).

### Issue Description
Open a pointer in a new browser tab when clicking on it while holding the [Cmd] button.

Closes #1711.

### Approach
Upon clicking on a pointer cell with CMD key held, the relation is opened in a new tab. In addition to this, for windows, an option to open the relation in new tab is added to the ContextMenu.
Screenshot 2021-08-18 at 2 25 24 AM

### TODO:
- [ ] Add changelog entry


mtrezza:
Thanks for opening this PR so quickly!

- The context menu work fine.
- When I Cmd+click on the pointer, a new tab opens with the pointer object, but the original tab also navigates to the pointer object. Could you look into that?

fn-faisal:
> Thanks for opening this PR so quickly!
> 
> * The context menu works fine.
> * When I Cmd+click on the pointer, a new tab opens with the pointer object, but the original tab also navigates to the pointer object. Could you look into that?

@mtrezza I have updated the PR.

Parse-Dashboard/parse-dashboard-config.json
```diff
@@ -2,13 +2,15 @@
   "apps": [
     {
       "serverURL": "http://localhost:1338/parse",
-      "appId": "hello",
```
mtrezza, line 5: Can you please revert? I already deleted the index.html file.

mtrezza (changes requested):

Parse-Dashboard/parse-dashboard-config.json
fn-faisal: I've reverted these changes.

mtrezza:
This looks good! Thanks for the fixes, could you just add a changelog entry and we should be good to merge this one.

> * Added `Cmd`-click on pointer to open link in new browser tab (fn-faisal) [#1757](https://github.com/parse-community/parse-dashboard/pull/1757)

Issue #1679, opened 2021-04-06
Deprecations and readiness for react 17+


romanbsd:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] You're running version >=1.0.23 of Parse Dashboard.

- [X] You're running version >=2.3.2 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Current parse-dashboard

#### Steps to reproduce
Check Chrome console.

#### Logs/Trace
Check Chrome console.

There're some parts of the code which use obsolete and legacy APIs, namely deprecated lifecycle, references and context APIs.
I started fixing all these (and other eslint) issues, made a pretty good progress so far (see https://github.com/parse-community/parse-dashboard/compare/master...romanbsd:fix_warnings for work in progress), and a pull request can be made, but I'll need someone to take over QA, since I don't even use all the functionality and less familiar with.
So, if someone is willing to help with testing, it can be done.

mtrezza:
Hi @romanbsd, thanks for opening this issue and apologies for the lack of activity here. We would be very interested in someone pickup up the modernization fo the dashboard. Would you still be interested in pursuing this and if yes, please let me how we can support you.

romanbsd:
My work (linked above) is in a pretty good shape, but needs a review. It needs some work, but it's mostly done.
So I can open a pull request and you can start reviewing the changes.

mtrezza:
Sure, let's do that! There have been many PR merged in the dashboard lately, so it may take a bit of effort to rebase your PRs on the current master branch.

damianstasik:
@romanbsd may I take this issue? I already started doing some work on this with string refs and `componentWillReceiveProps`/`componentWillMount` refactors and I could take the rest if you are okay with it.

romanbsd:
Yeah, sure. I don't have time for that at the moment. But take a look here: https://github.com/parse-community/parse-dashboard/pull/1677/files
and especially here:
https://github.com/parse-community/parse-dashboard/pull/1761
It may have all the changes you need.

mtrezza:
Closing in favor of https://github.com/parse-community/parse-dashboard/issues/1902

PR #1761, opened 2021-08-19
Fix warnings

romanbsd:

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description
Update code to the current versions of 3rd party dependencies and standards.
Related issue: #1679

### Approach
Following current React practices

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog


mtrezza:
I think we discussed this already somewhere, but switching to yarn is something we would need to look at, because we don't use it anywhere else to my knowledge.

romanbsd:
I can probably switch back to npm. There was some problem which I think was resolved since then.

mtrezza:
That would be good.

mtrezza:
@romanbsd There are a bunch of conflicts here, would you want to switch from yarn to npm and resolve these? Then I would take care of testing the dashboard.

PR #1769, opened 2021-08-27
ci: add issue bot for PRs

mtrezza:
- [x] Add entry to changelog

This apparently doesn't for with GH Actions, looking for a solution...


mtrezza:
This requires a GitHub App stead of Actions.

Issue #1770, opened 2021-08-30
Impossible to edit array of pointers in v2.2.0


johanarnor:
### New Issue Checklist
- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and [Parse Dashboard](https://github.com/parse-community/parse-dashboard/releases). 

### Issue Description
In v2.2.0 it's no longer possible to properly edit the value of an array containing pointers. https://github.com/parse-community/parse-dashboard/pull/1727 introduces a click action that overrides the standard in-place edit box.

I therefore tried to edit the data via the modal, but this would instead corrupt the data. The data is displayed like this (which if saved will corrupt the data)
```
[
  {"objectId": "3EAyCmr7BU"},
  {"objectId": "EAGB5dfkCe"}
]
```
instead of the actual raw format.
```
[
  {"__type":"Pointer","className":"Item","objectId":"3EAyCmr7BU"}, 
  {"__type":"Pointer","className":"Item","objectId":"EAGB5dfkCe"},
]
```
However this issue with the modal wasn't introduced in v2.2.0, since it's present in v2.1.0. Another interesting behaviour is that the whole object will be "inflated" (i.e. all fields will be populated) if you visit that object in the same browser tab prior to trying to edit the array of pointers.

Also, a CTRL+C puts "[object Object]" in the clipboard instead of the actual raw object.

### Expected Outcome
I'm not sure what a good solution would be here. Maybe just remove the click action to bring up the original edit box.

I've never used the modal, so I'm not really sure what's the expected outcome here.

mtrezza:
@sadakchap Could you take a look at this?

sadakchap:
Sure @mtrezza , Thanks!

sadakchap:
Here's the pr https://github.com/parse-community/parse-dashboard/pull/1771

mtrezza:
Closing via #1771

PR #1771, opened 2021-08-31
Edit Array of pointers

sadakchap:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues/1770).

### Issue Description

- Not able to edit array of pointer field, as single click on the cell redirects to pointer class.
- Ctrl+c copies `[object Object]` instead of field json value.
- Edit modal displays wrong data for array of pointers.

Related issue: #1770

### Approach

- Using Pill design for pointer fields. that will only redirect to pointer class on arrow icon click, otherwise on single should select the cell & on double click opens Editor for that cell.
- set copyableValue to json of field value.
- Using `Prase._encoding` to avoid converting object into parseObjects.

### TODOs before merging

- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 - fix: date cell value not selected on double clicks (fn-faisal) [#1730](https://github.com/parse-community/parse-dashboard/pull/1730)
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


mtrezza:
Thanks for this quick PR!

Can you just add a changelog entry, so we can merge this?

mtrezza (approved):
Thanks for fixing this so quickly!

Issue #1787, opened 2021-09-08
Docker image doesn't work anymore with the MFA feature


Klaitos:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and [Parse Dashboard](https://github.com/parse-community/parse-dashboard/releases). 

### Issue Description
The merge of the MFA feature https://github.com/parse-community/parse-dashboard/pull/1624 add some dependencies which need node 12 at leat. The docker image is built over node 8. The container can't be run anymore.

### Environment

Dashboard
- Parse Dashboard version: `FILL_THIS_OUT`
- Browser (Safari, Chrome, Firefox, Edge, etc.): `FILL_THIS_OUT`
- Browser version: `FILL_THIS_OUT`

Server
- Parse Server version: `FILL_THIS_OUT`
- Operating system: `FILL_THIS_OUT`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `FILL_THIS_OUT`

Database
- System (MongoDB or Postgres): `FILL_THIS_OUT`
- Database version: `FILL_THIS_OUT`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `FILL_THIS_OUT`

mtrezza:
We would have to make a 3.0 release as bumping the min node version is breaking.

a) How could we have prevented this issue?
  Our CI needs a multi-environment setup, like parse server. Currently: 
  - the PR CI runs only on node 10.14
  - the release CI only on node 12
  
b) How do we resolve this issue?
  Maybe an opportunity to start a LTS trial also for dashboard

Klaitos:
Hello @mtrezza,

I'm not surprise for a `latest` tag to be buggy, it was my mistake to use it in my docker-compose. I fixed my compose to stick the version to 2.0.5 to prevent future issues.

For your **b)** i think it's a good idea to release a last stable version of parse-dashboard 2.x without MFA and node 12.

For your **a)** i need to think more about it.

mtrezza:
[...]

PR #1781, opened 2021-09-04
ci: Fix docker image pushing to Docker Hub

cbaker6:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description

Related issue: #1782 

### Approach

Use GitHub Actions to push to Docker hub during merges of PRs and releases. Merges to the master branch push the latest image to `latest` on Docker Hub and releases push the latest release with the same tag name. You can see the tags this creates using my research repo: https://hub.docker.com/repository/docker/netreconlab/parse-dashboard. Version 2.2.1 of the dashboard is me testing the tagging.

This is setup to build images for multiple architectures, but currently only the `amd64` (Intel Linux) build passes.

A side-effect is it creates another tag on docker-hub called `master`, which I guess could be useful if something else is pushed after `latest`

### TODOs before merging

- [x] @TomWFox you will need to go to Docker Hub and create an [access token](https://docs.docker.com/docker-hub/access-tokens/) for the GitHub organization. The token should then be added to the GitHub organization and named `DOCKERHUB_TOKEN` with access to all of the repos. In addition you should add the Docker Hub user name to `DOCKERHUB_USERNAME` with access to all repos
- [x] Add entry to changelog

PR #1788, opened 2021-09-08
feat(docker): increase node version in docker to 12

Klaitos:

### Issue Description

Related issue: https://github.com/parse-community/parse-dashboard/issues/1787

### TODOs before merging

- [x] Add tests
- [x] Add entry to changelog

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 #
 # --- Base Node Image ---
-FROM node:8-alpine AS base
+FROM node:12-alpine AS base
 
 RUN apk update; \
   apk add git;
@@ -23,7 +23,7 @@ RUN npm run prepare && npm run build
 
 #
 # --- Production Image ---
-FROM node:8-alpine AS release
+FROM node:12-alpine AS release
 WORKDIR /src
 
 # Copy production node_modules
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "parse-dashboard": "./bin/parse-dashboard"
   },
   "engines": {
-    "node": ">=8.9"
+    "node": ">=12"
   },
   "main": "Parse-Dashboard/app.js",
   "jest": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add multi-factor authentication to dashboard login. To use one-time password, run `parse-dashboard --createMFA` or `parse-dashboard --createUser`. (Daniel Blyth) [#1624](https://github.com/parse-community/parse-dashboard/pull/1624)
 
 ## Improvements
+- Docker image use now node 12 version [#1788](https://github.com/parse-community/parse-dashboard/pull/1788)
 - CI now pushes docker images to Docker Hub (Corey Baker) [#1781](https://github.com/parse-community/parse-dashboard/pull/1781)
 - Add CI check to add changelog entry (Manuel Trezza) [#1764](https://github.com/parse-community/parse-dashboard/pull/1764)
 - Refactor: uniform issue templates across repos (Manuel Trezza) [#1767](https://github.com/parse-community/parse-dashboard/pull/1767)


dblythy:
Thank you @Klaitos for the investigation and quick PR!!

dblythy:
I think this line will need to be updated as well:

https://github.com/parse-community/parse-dashboard/blob/7686c704100c59921d90bcd8ffcd1d2db7cb4d66/package.json#L131

Dockerfile
```diff
@@ -1,6 +1,6 @@
 #
 # --- Base Node Image ---
-FROM node:8-alpine AS base
+FROM node:12-alpine AS base
```
cbaker6, line 3: This one and the one below should probably use the node tag, `lts-alpine` instead of `12` directly.

cbaker6:
@mtrezza did something change in the docker commit after I added #1781? I don't see the docker CI testing this PR as it would of when I made my changes and I think it's important to make sure PR's don't break the docker image before they go to the master branch.

Dockerfile
cbaker6: It seems, `12` builds, but `lts` doesn't. It may be something in the npm packages as this occurs with lts: https://github.com/netreconlab/parse-dashboard/runs/3545759058?check_suite_focus=true#step:7:378

mtrezza:
Yes, the flow "docker-publish" doesn't run on PR merges, because it is a publish flow that runs *after* merging a PR into master. If we want to check whether a PR breaks anything, we need to add a docker build check to the CI.

cbaker6:
> Yes, the flow "docker-publish" is a publish flow that runs _after_ merging a PR into master. If we want to check whether a PR breaks anything, we need to add a docker build check to the CI.

It can do that automatically, when you removed:

```bash
  pull_request:
     branches: [ master ]
```

It stopped this PR check

mtrezza:
Yes, but the reason for #1786 is that we need to stop updating the docker hub `latest` tag on every merge into master, if we want to transition the `latest` tag to be the "latest stable".

cbaker6:
> Yes, but the reason for #1786 is that we need to stop updating the docker hub `latest` tag on every merge into master, if we want to transition the `latest` tag to be the "latest stable".

It looks like it's updating latest because of the lines added 1786:

```bash
          flavor: |
             latest=${{ steps.branch.outputs.branch_name == 'master' }}
```

If those lines aren't there, it will update to latest stable on releases and tags

mtrezza:
True that, I thought `on push` only runs for direct push, not for PR merges. So I'll remove the `branches: [ master ]` too.

Dockerfile
mtrezza: ci tests don't pass for node 16 because node-sass (also in the logs you linked) needs to be bumped. maybe that's why? not sure which node version lts-alpine uses.

package.json
```diff
@@ -128,7 +128,7 @@
     "parse-dashboard": "./bin/parse-dashboard"
   },
   "engines": {
-    "node": ">=8.9"
+    "node": ">=12.22.6"
```
mtrezza, line 131: Can you change to `>=12`?

Dockerfile
cbaker6: > ci tests don't pass for node 16 because node-sass (also in the logs you linked) needs to be bumped it seems. maybe that's why?

Makes sense, should it be bumped in this PR?

cbaker6:
> The original PR pushes "latest" to docker hub on every commit, we don't want to do that for now. I thought on push only runs for direct push, not for PR merges, so I'd also remove the branches: [ master ] and the schedule.

My original PR does not do this. I showed you the example on my repo of it not updating `latest` on every commit, nor does it update `latest` on nightly. The docker action handles this correctly, there's no need to add anything for  `flavor` or removing the lines I put before. You can see this by looking at my tags. I've been adding updates, and `latest` hadn't been updated for over 1 day https://hub.docker.com/repository/registry-1.docker.io/netreconlab/parse-dashboard/tags?page=1&ordering=last_updated

package.json
Klaitos: As far as I investigated, MFA feature now depends on  `"inquirer": "8.1.2"` as seen on my screenshot from the pull request MFA.

Capture d’écran 2021-09-08 à 18 54 04

The new version of inquirer require ora package on version 5.3.0
The ora package need node 10 at least. (https://github.com/sindresorhus/ora/blob/476935f318868265303d148992fc268639a0d573/package.json)

So my i was wrong bumping to node 12. 10 is the new minimum (sorry for that).

Klaitos: I can confirm to you node 10 is ok, i've tried 8, 10 and 12 with docker build on my laptop.
8 = X fails due to ora package.
10 => ok
12 => ok

mtrezza:
Yes, I remember you removing the flavor, but when will it push to `latest` in your current setup?

Dockerfile
mtrezza: I tried, it seems to requires some code refactoring because this is a chain reaction where we need to bump webpack to 5 as well. Dashboard is currently not officially Node 16 tested, that will be an undertaking of its own.

package.json
mtrezza: `node-qrcode` also requires "node": ">=10.13.0", and even worse, `otpauth` does not specific any min. node version.

mtrezza (changes requested):
Everything below node 12 reached its EOL, so we should make Node 12 the minimum and also use node 12 for the docker image.

mtrezza (approved):
Looks fine!

Next steps are:
1. Merge, that triggers a docker push so we have both `master` and `latest` pointing to an  image that works.
2. Adapt the docker-publish CI to be triggered only when a tag is pushed

Klaitos:
Let me know if i can help somehow

mtrezza:
Maybe you could try out the new docker image to see whether it works fine.

Klaitos:
will do !

Klaitos:
```
➜  Desktop docker run  -p 8080:4040 -v "/parsedashboard.json:/src/Parse-Dashboard/parse-dashboard-config.json" parseplatform/parse-dashboard:latest       

The dashboard is now available at http://0.0.0.0:4040/

```

I can confirm the dashboard works on the `latest` tag👏 

Thanks @mtrezza

mtrezza:
Amazing! Thanks for investigating this issue to thoroughly!

cbaker6:
> I thought we were not sure what the auto behavior for flavor is and when it will push to the latest tag? 

I don't know all of the logistics of auto, but I do know some... It doesn't update `latest` when merging into the master branch.

> Because I'm not sure whether that will happen when we push a version tag 3.0.0 - will it create a docker image with tag 3.0.0 and another identical one with latest?

This will create a docker image for `3.0.0` as well as update `latest`

mtrezza:
Thanks for clarifying, that is good to know for when we remodel for the new branches. For now we'll only push release version tag to docker to have more control over the pushes while we work on the new branches; see https://github.com/parse-community/parse-dashboard/pull/1789.

PR #1797, opened 2021-09-10
fix: bump graphql-playground

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description

Related issue: https://github.com/parse-community/parse-dashboard/pull/1645

Issue #1639, opened 2021-01-13
Install error on npm 7


tristanyi:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.

- [x] You're running version >=2.3.2 of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
ubuntu18
node 14
npm 6

#### Steps to reproduce

npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! core-js@2.6.12 postinstall: `node -e "try{require('./postinstall')}catch(e){}"`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the core-js@2.6.12 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

#### Logs/Trace
Note: If you get a browser JS error please run `npm run dev`. This will provide source maps and a much more useful stack trace.

elrumo:
I'm also having issues installing parse-server from NPM. I get dependancy errors.

siempay:
I am having same problem right now !

dplewis:
Can you try ignore scripts flag on install?

npm install --ignore-scripts

siempay:
Just tried it and it failed too 
```code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! Found: graphql@14.6.0
npm ERR! node_modules/parse-dashboard/node_modules/graphql
npm ERR!   graphql@"14.6.0" from parse-dashboard@2.1.0
npm ERR!   node_modules/parse-dashboard
npm ERR!     parse-dashboard@"*" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer graphql@"^0.11.0 || ^0.12.0 || ^0.13.0" from graphql-import@0.4.5
npm ERR! node_modules/parse-dashboard/node_modules/codemirror-graphql/node_modules/graphql-language-service-interface/node_modules/graphql-import
npm ERR!   graphql-import@"^0.4.4" from graphql-config@2.0.1
[...]

Issue #1633, opened 2020-12-22
Dependency on codemirror-graphql branch


mtrezza:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [x] You're running version >=1.0.23 of Parse Dashboard.
- [x] You're running version >=2.3.2 of Parse Server.
- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Parse Dashboard has a dependency on a personal branch:
https://github.com/parse-community/parse-dashboard/blob/faf8d2dc739625f183c8b548b8a3d09b079276a7/package.json#L41

The branch is not maintained.

#### Solution
The dependency should point to the [original repository](https://github.com/graphql/graphiql/tree/main/packages/codemirror-graphql) or Parse should to maintain its own fork if necessary.

Nes-si:
Seems like this issue makes it impossible using **parse-dashboard** in project if you have new npm version.
`npm i` finishes with an error:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! Found: graphql@15.4.0
npm ERR! node_modules/graphql
npm ERR!   graphql@"15.4.0" from parse-server@4.5.0
npm ERR!   node_modules/parse-server
npm ERR!     parse-server@"^4.5.0" from the root project
npm ERR!   peer graphql@"^14.0.0 || ^15.0.0" from graphql-language-service-parser@1.8.0
npm ERR!   node_modules/graphql-language-service-parser
npm ERR!     graphql-language-service-parser@"^1.0.18" from codemirror-graphql@0.6.12
npm ERR!     node_modules/codemirror-graphql
npm ERR!       codemirror-graphql@"github:timsuchanek/codemirror-graphql#details-fix" from parse-dashboard@2.1.0
npm ERR!       node_modules/parse-dashboard
npm ERR!         parse-dashboard@"^2.1.0" from the root project
npm ERR!     graphql-language-service-parser@"^1.2.2" from graphql-language-service-interface@1.3.2
npm ERR!     node_modules/codemirror-graphql/node_modules/graphql-language-service-interface
npm ERR!       graphql-language-service-interface@"^1.0.18" from codemirror-graphql@0.6.12
npm ERR!       node_modules/codemirror-graphql
[...]

PR #1807, opened 2021-09-20
feat: add support for npm 7 and Node 16, migrate to graphiql

dblythy:

### Issue Description

Related issue: #1797

### Approach

With [GraphQL Playground](https://github.com/graphql/graphql-playground/issues/1143) soon to be archived and not receiving support, it's sub-dependancies are not compatible with GraphQL 15. The course of action here is to migrate to GraphQL playgrounds successor, [GraphIQL](https://github.com/graphql/graphiql)

This fixes install issues on npm 7.

Note: GraphIQL is only compatible with GraphQL=<15.5.0

Closes:

#1797
#1752
#1633
#1639

### TODOs before merging

- [x] Add entry to changelog


mtrezza:
Yay! 🙌 for this PR!

dblythy:
@mtrezza circular dependancies throws `Error: Cannot find module 'typescript'`. Do you have any ideas how to fix this?

mtrezza:
Without having looked into it, I'd assume the module is referenced somewhere but not imported or defined as dependency. Could also be an upstream issue.

mtrezza:
- This PR uses graphql 15.5.0
- Parse Server `master` uses graphql 15.5.2
- Parse Server `4.10.3` uses graphql 15.4.0

Could there be incompatibilities between these different versions? In other words, do we need a compatibility table dashboard <> server?

dblythy:
Good idea, I will downgrade the dependancy to 15.4.0

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ## New Features
 ## Improvements
 ## Fixes
+- Fix npm 7 installation issue, migrate GraphQL playground to GraphIQL  (Daniel Blyth) [#1807](https://github.com/parse-community/parse-dashboard/pull/1807)
```
mtrezza, line 9: Can you remove this change? the changelog entry will be created automatically using the pull request title

mtrezza:
Does this also add support for node 16? Currently dashboard requires `"node": ">=12.0.0 <16.0.0"`. If node 16 is supported, we can remove the `<16.0.0` and add node 16 to the CI workflow (I can do that for you if you prefer).

dblythy:
I’m honestly not sure

dblythy:
Bumped some dependancies and some minor refactors to support Node 16

.github/workflows/ci.yml
```diff
@@ -123,9 +123,8 @@ jobs:
             NODE_VERSION: 12.22.6
           - name: Node 14
             NODE_VERSION: 14.17.6
-          # Enable the following lines when Parse Dashboard reached Node 16 compatibility
-          # - name: Node 16
-          #   NODE_VERSION: 16.9.0
+          - name: Node 16
+            NODE_VERSION: 16.9.0
```
mtrezza, line 127: Can you change this to the latest `16.10.0`, and also at the top of this file.

mtrezza (changes requested):

mtrezza:
Looks good, I'll just try it out locally with some node versions to see whether it all works fine.

mtrezza:
For npm 7 it installs, but I get this:

```
WARNING in ../node_modules/crypto-js/core.js 43:22-39
Module not found: Error: Can't resolve 'crypto' in '.../parse-dashboard/node_modules/crypto-js'
```

dblythy:
Let me know which way you think is best to approach this. I have no idea about webpack, react, etc, I was just trying to get this package fixed for npm 7.

mtrezza:
Webpack 4 used to auto-polyfill dependencies that were not meant to run in a browser. `crypto-js` being one of them, which is a dependency of Parse JS SDK which is a dependency of Parse Dashboard. Webpack 5 doesn't do that anymore, we'd need to manually add poyfills. The issue with polyfills is that they are essentially hacks as a substitute for native libraries that don't support running in browsers.

It seems this has been solved in https://github.com/brix/crypto-js/pull/364. Parse JS SDK 3.3.0 depends on `crypto-js@4.0.0`, so we'd need to bump it to the latest version, which should fix this. Alternatively move to another library.

mtrezza:
I release Parse JS SDK 3.3.1 Can you bump parse to 3.3.1 in this PR? Then we should good to release.

package.json
```diff
@@ -141,7 +141,7 @@
     "parse-dashboard": "./bin/parse-dashboard"
   },
   "engines": {
-    "node": ">=12.0.0 <16.0.0"
+    "node": ">=12.13.0"
```
mtrezza, line 144: This is a breaking change that would require a dashboard release 4.0.

Is there any way to avoid the 12.13.0 bump?

dblythy: I’m pretty sure I bumped it because it was causing the tests to fail

mtrezza: Found it, `sass-loader requires at least node 12.13.0 (>= 12.13.0)`

https://github.com/parse-community/parse-dashboard/runs/3698734537?check_suite_focus=true

dblythy:
Should be good to go!

mtrezza:
Really? Does this build fine with npm 7 and 6?

dblythy:
Yep, worked fine for me!

mtrezza:
I can confirm, it works for me locally too with npm 6 and 7.

Should we keep the package-lock file as version 1 (npm 6)? If we add a lock-file of version 2, there may be backward compatibility issues with npm 6.

mtrezza (approved):
Amazing work! This fix will surely be a relieve for many developers. Given the Node 16 / npm 7 support, I bumped this up to a feature commit.

parseplatformorg:
🎉 This pull request has been released in version [3.1.0](https://github.com/parse-community/parse-dashboard/releases/tag/3.1.0)

mtrezza:
Congrats @dblythy, your PR was the first one across Parse to be released automatically 🙌

LMBernardo:
@dblythy 
This has been a constant pain for me, I can't thank you enough for this PR. Greatly appreciated! Throw a crypto address on your profile and I'll buy you a beer. :grin: :beers:

dblythy:
@LMBernardo your generosity is greatly appreciated, however all of the core team are constantly working at improving our packages, and I can’t accept your offer simply because I addressed a difficult to isolate problem. If you would like to support our work, you are welcome to contribute to our [OpenCollective](https://opencollective.com/parse-server) 😊

Enjoy the update and let us know if you run into any other problems, or have any suggestions for improvements!

stage88:
Big win guys. Thank you.

PR #1827, opened 2021-10-04
feat: add ctrl + arrow key to jump to first/last cell; fix left/right key navigation with hidden columns

404-html:

- [ ] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description
This merge request is about following:

1. Fix for cell selection going to different dimension when navigating horizontally and having some columns hidden;

parse-hidden-columns

2. Adding support for ctrl+left/right keys. With such combination navigation jumps to first/last visible column;

### TODOs before merging

- [ ] Add tests


mtrezza:
I reworded the PR title slightly so that others reading the changelog have a better idea of the fix

mtrezza:
Please feel free to request a review when this is ready.

404-html:
Hey @mtrezza, thank you for your changes. I believe it's ready to be reviewed, please proceed.

src/dashboard/Data/Browser/DataBrowser.react.js
```diff
@@ -191,11 +199,12 @@ export default class DataBrowser extends React.Component {
         }
         e.preventDefault();
         break;
-      case 37: // Left
+      case 37: // Left - standalone & with ctrl/meta
```
mtrezza, line 202: Could you clarify the comment for future developers?

mtrezza:
Feature tested and works great!

src/dashboard/Data/Browser/DataBrowser.react.js
404-html: Comment updated 👍

src/dashboard/Data/Browser/DataBrowser.react.js
```diff
@@ -209,11 +218,12 @@ export default class DataBrowser extends React.Component {
         });
         e.preventDefault();
         break;
-      case 39: // Right
+      case 39: // Right - standalone & with ctrl/meta
```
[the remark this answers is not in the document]
404-html: Comment updated 👍

404-html:
Thanks for testing @mtrezza. I've added support for ctrl+up/down but I'm afraid it still needs some work.
While it works fine with small data sets it's not fully reliable when lazy-loading comes into place - auto scroll is not working. The reason behind this is that with large data sets not all rows are rendered (probably for performance reasons) and code responsible for scrolling to the current row is in `BrowserCell` component (`componentDidUpdate` lifecycle method).
So when, let's say, we are trying to jump from first to last (not rendered) row scrolling logic will not execute.
Changing this performance optimization to render all rows just for this feature is not reasonable.

Another thing I tried is to watch for current row change in `BrowserTable -> componentWillReceiveProps` and when it's first or last - scroll up or down to the bottom. Best way here would be to implement some smarted logic which will perform scroll only when needed. Anyway this change was not working reliably and it was quite slow, feels like all rows between were rendered which is not efficient.

My preference would be to abstract all scrolling logic into single class as having it now split between `BrowserTable` and `BrowserCell` (I'm not sure about other places) makes it a bit hard to not overlap.

I should have some time to work on that next week, or we can merge that as it is and raise new issue for ctrl+up/down + lazy loading optimization. Do you have any preference?

mtrezza:
Thanks for the thorough analysis!

> I should have some time to work on that next week, or we can merge that as it is and raise new issue for ctrl+up/down + lazy loading optimization. Do you have any preference?

I think this already provides great value as it currently is. The scrolling refinement maybe can be fixed as a patch in a separate PR, so we can already make some developers happy with the added left/right jumping feature. Would that be fine with you?

404-html:
Yes, thats fine with me. Thank you for asking 🤝

mtrezza (approved):
Looks good! Thanks for this improvement to speed-up the workflow with Parse Dashboard!

parseplatformorg:
🎉 This pull request has been released in version [3.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/3.2.0)

Issue #1826, opened 2021-10-03
Explanation of the `--dev` parameter


cyb3rko:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation
I found the hint on the README which adresses the `--dev` parameter of the dashboard ([See HERE](https://github.com/parse-community/parse-dashboard#local-installation)), which only says that it `is disabling production-ready security features` and that it's `useful if you are running on docker`.
But what exactly does that mean and what does it change in comparison to production mode?

### Feature / Enhancement Description
I think it would be a helpful addition if there could be a small description added to the README to explain this parameter, otherwise we don't really know why to use it.

cyb3rko:
@mtrezza I think it's enough information, right?
Or should I add empty information fields to make the assistant happy?

mtrezza:
Yes, thanks for reporting, it's clear what you mean (we will simplify the templates soon). Would you want to create a PR for this?

cyb3rko:
I would like to but I don't know what the parameter does so I think I can not help here, sry.
That's why I opened this issue.

mtrezza:
Did you take a look into the code? That would be my first step, I think it should be pretty easy to discover.

cyb3rko:
I took a look and I found this:

`program.option('--dev', 'Enable development mode. This will disable authentication and allow non HTTPS connections. DO NOT ENABLE IN PRODUCTION SERVERS');`

This fits to the occurences I found in the code for the keyword `dev` where it allows non HTTPS connections.
Do you think that's it? Then I could add it to the README?

mtrezza:
If you follow the dev parameter in code, is insecure http the only effect?

cyb3rko:
No, there's another point.

The description of the allowed non HTTPS connections in localhost is:
`Disallow HTTP requests except on localhost, to prevent the master key from being transmitted in cleartext`

The other thing is that auth is not necessary if you access the Dashboard via localhost:
`Allow no-auth access on localhost only, if they have configured the dashboard to not need auth`

That's it

mtrezza:
Sounds good, do you want to add a paragraph to the README?

[...]

PR #1842, opened 2021-10-06
docs: add info about --dev parameter

cyb3rko:

### Issue Description
Added a brief explanation of what the `--dev` parameter exactly does to be able to better evaluate if you really need it.

Related issue: #1826

### Approach
n/a

### TODOs before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,13 @@ parse-dashboard --dev --appId yourAppId --masterKey yourMasterKey --serverURL "h
 
 You may set the host, port and mount path by supplying the `--host`, `--port` and `--mountPath` options to parse-dashboard. You can use anything you want as the app name, or leave it out in which case the app ID will be used.
 
-NB: The `--dev` parameter disables production-ready security features. This parameter is useful when running Parse Dashboard on Docker. Using this parameter will:
+The `--dev` parameter disables production-ready security features. This parameter is useful when running Parse Dashboard on Docker. Using this parameter will:
+
 - allow insecure http connections from anywhere, bypassing the option `allowInsecureHTTP`
 - allow the Parse Server `masterKey` to be transmitted in cleartext without encryption
 - allow dashboard access without user authentication
-> ⚠️ Do not use this parameter when deploying Parse Dashboard in a production environment.
 
----
+> ⚠️ Do not use this parameter when deploying Parse Dashboard in a production environment.
 
 After starting the dashboard, you can visit http://localhost:4040 in your browser:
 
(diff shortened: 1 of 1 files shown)


mtrezza:
Also see this: https://github.com/parse-community/parse-dashboard/pull/962

cyb3rko:
Fits to what I added I think. It just means that the security bypasses only work if having the same IP as the dashboard.

mtrezza:
Maybe we can restyle this paragraph as a whole, how about this:

---

The `--dev` parameter disables production-ready security features. This parameter is useful when running Parse Dashboard on Docker.  Using this parameter will:
- allow insecure http connections from anywhere, bypassing the option `allowInsecureHTTP`
- allow the Parse Server `masterKey` to be transmitted in cleartext without encryption
- allow dashboard access without user authentication

> ⚠️ Do not use this parameter when deploying Parse Dashboard in a production environment.

---

This is how I read the changes from #962:

https://github.com/parse-community/parse-dashboard/blob/7c56b757dfb6cd1dd48c75968137c16a31a9565e/Parse-Dashboard/app.js#L91-L101

https://github.com/parse-community/parse-dashboard/blob/7c56b757dfb6cd1dd48c75968137c16a31a9565e/Parse-Dashboard/app.js#L141-L146

That means the dev param bypasses `allowInsecureHTTP` and allows insecure http connections from anywhere, regardless whether from localhost or remote address. That seems to be exactly the intended fix for Docker, because in the PR, [they say:](https://github.com/parse-community/parse-dashboard/pull/962#issue-400949125)

> when running with docker, as the local IP address is different from the dashboard, all the security features would light up

cyb3rko:
I see, so insecure http connections and non-auth access is always allowed on localhost and the `--dev` parameter is only for docker deployment. Alright.

Added your paragraph to the README.

mtrezza (approved):
Looks good! Thanks for making the docs more understandable!

parseplatformorg:
🎉 This pull request has been released in version [3.2.1-alpha.1](https://github.com/parse-community/parse-dashboard/releases/tag/3.2.1-alpha.1)

parseplatformorg:
🎉 This pull request has been released in version [3.2.1-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/3.2.1-beta.1)

parseplatformorg:
🎉 This pull request has been released in version [3.2.1](https://github.com/parse-community/parse-dashboard/releases/tag/3.2.1)

Issue #1849, opened 2021-07-12
Distinguish external from internal classes


davimacedo:
### New Feature / Enhancement Checklist

- [ X ] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ X ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ X ] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

It is currently possible to create a class called User and a class called _User. Both show up in Parse Dashboard with the name User and it causes confusion for some users when using Parse Server for the first time.

In addition to that, it is very confusing when we have to use _User or User throughout the sdks, endpoints, etc.

### Feature / Enhancement Description

I believe we should change Parse Server in such a way that User and _User point out to the same thing. So no matter if we send User, or _User, via the api, they will always refer to the _User class we currently have.

### Alternatives / Workarounds

An alternative could be simply disallow the developers to create a User class.

damianstasik:
@mtrezza I tried to experiment with grouping classes visually and for now, it seems like showing the real name would be the simplest solution:

<img width="263" alt="Screen Shot 2021-10-20 at 15 56 04" src="https://user-images.githubusercontent.com/920747/138106972-e2701816-8e87-4a29-9f08-87ca0c1b3bb0.png">

Would you accept a PR with such change?

mtrezza:
Sure, would you add a line to separate the internal classes, like in the screenshots in https://github.com/parse-community/parse-dashboard/issues/1849#issuecomment-886039636? I think it would be a helpful visual guide to faster identify internal classes.

damianstasik:
@mtrezza yes, [the PR is ready](https://github.com/parse-community/parse-dashboard/pull/1878), and here is how it looks:

https://i.imgur.com/dzfhv7Z.png

The only thing that I am not sure about is the separator API, there are a few ways of doing it, I chose the one that could be reused in other lists if needed.

mtrezza:
Looks perfect! I'll review.

mtrezza:
Closing via #1878

PR #1878, opened 2021-10-20
feat: add visual distinction in data browser for internal classes and display their real names with underscore

damianstasik:

### Issue Description
This PR refactors parts of the app that shows class names so that it displays the real name (the one with an underscore). Additionally, there is now also a separator between internal and custom classes in the browser sidebar.

Related issue: #1849

### Approach
- Show real class names
- Add separator between internal and custom classes

### TODOs before merging


damianstasik:
There are a few ways of implementing the separator:
1. Using separate `<CategoryList />` in the `Browser.react.js`, but that would mean the separator would be a feature only available in the browser sidebar.
2. Adding a new type of category item with a new property called `type`, in this case, it would include only `separator` and would be handled for any `CategoryList` instance.
3. Combination of 1&2 would be allowing a `component` property in the category list which could be used in `Browser.react.js` to add a custom separator only for the browser sidebar.

I chose the second option as it does not produce a duplicated code if other sections of the dashboard would also want a separator between categories.

src/lib/Constants.js
```diff
@@ -45,14 +45,14 @@ export const Directions = {
...
 
-export const SpecialClasses = {
-  _User: 'User',
-  _Installation: 'Installation',
-  _Role: 'Role',
-  _Product: 'Product',
-  _Session: 'Session',
-  _PushStatus: 'PushStatus',
-};
+export const SpecialClasses = [
```
damianstasik, line 48: Changed from the map into an array as the object values are no longer used anywhere.

mtrezza: Looks fine

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This pull request has been released in version [3.3.0-alpha.9](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0-alpha.9)

parseplatformorg:
🎉 This change has been released in version [3.3.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [3.3.0](https://github.com/parse-community/parse-dashboard/releases/tag/3.3.0)

PR #1885, opened 2021-10-26
feat: New option to make modal resizable

### LINKED ISSUE #1879: Allow config parameter dialog to be resized in width

### Current Limitation
Currently, the modal dialog of a Config parameter can only be resized in height but is fixed in width. That makes it difficult when editing large values, such as an array with many elements.

image

Allow the dialog window to be resized in width too.

### Example Use Case
n/a

Edit value in external editor and copy/paste.

### 3rd Party References
n/a

damianstasik:

### Issue Description
This PR adds a "resize" handle to modals that have the `resizable` prop enabled.

Closes: #1879

### TODOs before merging


damianstasik:
Here is how the final resize handle looks:
Screen Shot 2021-10-27 at 20 22 07

damianstasik:
@mtrezza regarding labels being resized as well: I tried to force a `max-width` on them, but the `Field` component would require a slightly bigger refactor that I did not want to include in this PR.

mtrezza:
That's fine then, thanks for looking into it.

mtrezza (changes requested):
I tried it out and I wonder how we can combine the vertical and horizontal resizing by dragging the dialog corner. That the dialog can only be resized horizontally in the dialog corner and only vertically in the field corner is a rather unusual behavior.

There is also a bug that selects all text in the dialog when I resize horizontally:

https://user-images.githubusercontent.com/5673677/139144024-1725f5ea-4505-4820-ba92-97696648b230.mov

damianstasik:
@mtrezza Maybe instead of the drag handle in the corner, we could allow resizing the modal with both the left and right edge of the container.

mtrezza:
Not sure how the experience of that would be. I can only think of a corner drag handle, like the field had. That is the UI behavior a user would typically expect. If there is another way that is intuitive, we can also explore that.

damianstasik:
It could work as in a regular OS window manager, the whole side could be used to resize the modal:

https://user-images.githubusercontent.com/920747/139558468-ee3f470a-853f-4e58-9af2-32bb575d67d7.mov

mtrezza:
If it's separately resizable on the horizontal and vertical side it would certainly be an improvement over the status quo, but it would still be impractical when you cannot resize both simultaneously. I think that's an issue when want to resize because there is a large JSON object that you want to display and you want to fit the dialog just right to display it.

What is the difficulty to let resize both at the same time? Is it a component restriction?

mtrezza:
Superseded by https://github.com/parse-community/parse-dashboard/pull/2910

Issue #1902, opened 2021-10-31
Outdated dependencies


damianstasik:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and [Parse Dashboard](https://github.com/parse-community/parse-dashboard/releases). 

### Issue Description
Currently, most of the dependencies from `package.json` are outdated, some of them have a vulnerability reported by `npm audit`. I created this ticket to keep track of the progress, as I think the best way to get everything updated is to update everything one by one which helps with testing and debugging what went wrong.

### Steps to reproduce
Run any command that checks for dependency updates.

### Actual Outcome
- [x] [@babel/runtime](https://github.com/parse-community/parse-dashboard/pull/1938)                            
- [ ] bcryptjs                                  
- [ ] commander                                 
- [ ] [graphiql](https://github.com/parse-community/parse-dashboard/pull/1907)                                  
- [ ] graphql                                    
- [ ] history                                    
- [x] [immutable](https://github.com/parse-community/parse-dashboard/pull/1900)                                  
- [x] [package-json](https://github.com/parse-community/parse-dashboard/pull/1904)                              
- [ ] react                                       
- [ ] react-ace                                 
- [ ] react-dnd                                  
- [ ] react-dnd-html5-backend                    
- [ ] react-dom                                   
- [x] [react-helmet](https://github.com/parse-community/parse-dashboard/pull/1906)                              
- [x] [react-popper-tooltip](https://github.com/parse-community/parse-dashboard/pull/1901)                      
- [ ] react-redux                               
- [ ] [regenerator-runtime](https://github.com/parse-community/parse-dashboard/pull/1945)
- [ ] [semver](https://github.com/parse-community/parse-dashboard/pull/1947)
- [ ] [@actions/core](https://github.com/parse-community/parse-dashboard/pull/1947)
- [x] [@babel/core](https://github.com/parse-community/parse-dashboard/pull/1938)                               
- [x] [@babel/plugin-proposal-decorators](https://github.com/parse-community/parse-dashboard/pull/1938)         
[...]

PR #1905, opened 2021-10-31
refactor: remove unused dependencies

damianstasik:

### Issue Description
Both `typescript` and `babel-plugin-transform-object-rest-spread` are not used anywhere, there is no need to keep them.

Related issue: #1902

### TODOs before merging


package.json
```diff
@@ -71,8 +71,7 @@
     "react-router": "5.2.1",
     "react-router-dom": "5.3.0",
     "regenerator-runtime": "0.13.8",
-    "semver": "7.3.4",
-    "typescript": "4.4.3"
```
mtrezza, line 75: typescript is needed

mtrezza (changes requested):

package.json
damianstasik: @mtrezza could you specify what is using typescript? I checked the repo and there is no configuration for TS, there are no `.tsx?` files and no build set up for TS 🤔

mtrezza: madge uses it; even if you don't see any reference for a dependency in the code base, another dependency may require it based on its configuration.

mtrezza:
Could you please resolve the conflict?

package.json
damianstasik: @mtrezza I checked the report `madge` generates and everything works.

Screen Shot 2021-11-01 at 16 28 49

> even if you don't see any reference for a dependency in the code base, another dependency may require it based on its configuration.

Yeah, I get that, I wouldn't try to send the PR otherwise. I checked peer deps of other packages before to be sure if TS is really not required.

mtrezza: image

It's a dependency N levels deep, you'd have look at `package-lock.json` from madge to see that, you won't see that in its `package.json`. I'm assuming it's configuration-dependent in madge, so it's probably not used in all cases, that's why it may be not a fixed dependency of the madge package.

PR #1946, opened 2021-12-05
refactor: upgrade linting dependencies

damianstasik:

### Issue Description
- This PR replaces deprecated `babel-eslint` with `@babel/eslint-parser` and updates ESLint deps.
- I ran the `lint` script and fixed simple issues, but there is one issue not related to this PR in `src/dashboard/Data/Browser/Browser.react.js@464` which should be addressed in a separate issue.
- I also removed `experimentalObjectRestSpread` and `jsx` flags from ESLint config as they are not needed, `jsx: true` is set by the React plugin and `experimentalObjectRestSpread` is no longer experimental.
- `root: true` is needed, so that Babel ESLint parser knows how to locate the babel config...
- ...which I extracted to a separate file so that both webpack and ESLint share the Babel same config
- This PR also upgrades all Babel deps to avoid installing older version of `@babel/eslint-parser` that was compatible with older deps.

Related issue: #1902

### TODOs before merging


src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -402,7 +402,7 @@ class Browser extends DashboardView {
           if (name === 'objectId' || this.state.isUnique && name !== this.state.uniqueField) {
             return;
           }
-          if (!!required) {
```
mtrezza: I am always critical of this rule, because it's not a mere style change but a code change, specifically removing a cast to boolean. Maybe we should disable this rule? Regarding this change we also have to make sure that `required` is always of boolean type.

src/dashboard/Data/Browser/DataBrowser.react.js
```diff
@@ -259,6 +259,7 @@ export default class DataBrowser extends React.Component {
   getNextVisibleColumnIndex(distance = 1, min = 0, max = 0) {
     if (distance === 0) { return this.state.current.col; }
     let newIndex = this.state.current.col + distance;
+    // eslint-disable-next-line no-constant-condition
     while (true) {
```
mtrezza: I suggest to change the code instead of disabling the warning?

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -506,7 +506,7 @@ class Browser extends DashboardView {
...
```
mtrezza: Same here

package.json
```diff
@@ -35,13 +35,15 @@
     "LICENSE"
   ],
   "dependencies": {
-    "@babel/runtime": "7.16.3",
+    "@babel/runtime": "7.16.7",
+    "@babel/runtime-corejs3": "^7.16.8",
```
mtrezza: Nice! could you just remove the `^` from the version?

src/dashboard/Data/Browser/Browser.react.js
mtrezza: I guess it's fine.

src/dashboard/Data/Browser/Browser.react.js
mtrezza: I guess it's fine.

src/dashboard/Data/Browser/DataBrowser.react.js
mtrezza: I guess it's fine.

src/dashboard/Push/PushAudiencesIndex.react.js
```diff
@@ -245,7 +245,7 @@ class PushAudiencesIndex extends DashboardView {
       }}>
     </FormModal>
 
-    if (typeof(data) !== undefined) {
+    if (typeof data !== 'undefined') {
```
mtrezza, line 248: Why so complicated? What is the difference between `typeof data !== 'undefined'`and `data !== undefined`?

mtrezza:
Could you take a look at the failing test so can merge this?

src/dashboard/Push/PushAudiencesIndex.react.js
damianstasik: In this case there would not be any difference as the `data` variable exists.

ci/nodeEngineCheck.js
```diff
@@ -76,16 +76,20 @@ class NodeEngineCheck {
...
+        const contentJson = JSON.parse(contentString);
+        const version = ((contentJson || {}).engines || {}).node;
+
+        // Add response
+        response.push({
+          file: file,
+          nodeVersion: version
+        });
+      } catch {
+        console.log(`Skipping ${file} due to package.json parsing issues`);
```
mtrezza: Not sure about silent fail; if it fails then we may overlook something. For which files did you see it failing?

damianstasik: For `eslint-plugin-react/node_modules/resolve/test/resolver/malformed_package_json/package.json` which is malformed specifically for their tests.

mtrezza: Got it, I think the better option here is to change the script to exclude certain paths (defined by regex) from parsing. This way we can exclude `eslint-plugin-react.*/malformed.*`

damianstasik: To be honest, I do not like how currently `engine` is being tested. I would propose to use npm's config flag called [`engine-strict`](https://docs.npmjs.com/cli/v7/using-npm/config#engine-strict) to stop CI if any of dependencies does not meet requirements.

damianstasik: Unfortunately it seems like `engine-strict` option does not work for npm@6, tested it locally and it does not work at all.

mtrezza: engine-strinct doesn't do the same as our script. If engines.node is for example >=12 and a dependency is >=14, then engine-strict will pass, as far as I remember. Our script will reject, because the dependency requires a higher node version. I wrote the script because currently there is no npm option for that.

damianstasik: Based on what npm's docs describe if you set `engine-strict` to `true` and try to install a dep that requires Node 14 on OS that has Node 12 set as a default, the installation will fail. I tried it in one PR, but it seems broken on npm 6, on npm 7/8 it works for some, but that would require additional work.

mtrezza: I would be happy if we could remove our script and replace it with `engine-strict`.

> If set to true, then npm will stubbornly refuse to install (or even consider installing) any package that claims to not be compatible with the current Node.js version.

That means it won't install based on the Node.js version, not based on `engines.node` in `package.json`. The script currently considers the `engines.node` version, not the installed Node version. To achieve the same, it means we have to add an additional Node environment in the CI with the version set in `engines.node`. That means currently for example, we need two Node 12 CI environments; the absolute minimum Node 12 version in `engines.node` and the latest Node 12 (minor/patch) version.

I think that is an even better approach than using our custom script. Because we are actually testing it in the CI. So I'm for replacing that, but we should do that in a separate issue and PR. Do you want to open a PR? It's fairly easy to add a CI job for the min. Node 14 version (it will probably be 2 duplicate jobs for now), with a comment that the new job must stay in sync with the engines.node version in package-lock.

mtrezza (changes requested):
Not sure how to resolve the conflict, `tapable` appears 2 times in the file.

ci/nodeEngineCheck.js
```diff
@@ -75,21 +75,20 @@ class NodeEngineCheck {
...
-        const version = ((contentJson || {}).engines || {}).node;
-
-        // Add response
-        response.push({
-          file: file,
-          nodeVersion: version
-        });
-      } catch {
-        console.log(`Skipping ${file} due to package.json parsing issues`);
+      if (/eslint-plugin-react\/.*malformed.*/.test(file)) {
```
mtrezza: The try / catch statement should be kept; the excluded filenames should be passed as an argument from the highest level in `check`:

```
const ignoredFilePaths = [
  '/eslint-plugin-react\/.*malformed.*/',
];

const versions = await check.getNodeVersion({ files, ignoredFilePaths, clean: true });`
```

mtrezza (changes requested):

ci/nodeEngineCheck.js
damianstasik: @mtrezza what do you want to do in the `catch` block?

mtrezza: Same as before, that should not change.

mtrezza:
Could you take a look at at the conflict? I tried to resolve it manually, but the CI started to fail, so I reverted.

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.4](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0-alpha.4)

parseplatformorg:
🎉 This change has been released in version [5.0.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.0.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0)

PR #1947, opened 2021-12-05
refactor: upgrade CI deps, bump Node.js version to 14

damianstasik:

### Issue Description
This PR upgrades CI dependencies, but also bumps the minimum version of Node to 14 since `semantic-release` minimum Node version is 14.17.0. It would be also good for the dashboard to require Node 14 as Node 12 will reach end-of-life is on 2022-04-30.

Related issue: #1902

### TODOs before merging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,8 +111,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: Node 12
-            NODE_VERSION: 12.22.7
           - name: Node 14
             NODE_VERSION: 14.18.1
           - name: Node 16
--- a/.github/workflows/release-automated.yml
+++ b/.github/workflows/release-automated.yml
@@ -13,7 +13,7 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-node@v2
         with:
-          node-version: 12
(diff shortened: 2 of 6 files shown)


damianstasik:
@mtrezza `semantic-release` requires Node v14, should we ignore the update or bump our Node version to match it?

mtrezza:
> ... bumps the minimum version of Node to 14 since semantic-release minimum Node version is 14.17.0.

A dev dependency min node version should not determine the product's min node version. Dev dependencies are not used in the final product, unless maybe dev dependencies for testing. It's possible that the "Node engine check" also considers dev dependencies, which it should not. If there is no good reason to upgrade semantic release to a version that requires Node 14, we don't have to do it. In April next year we'll drop Node 12 support anyway, then we can also upgrade semantic release.

> It would be also good for the dashboard to require Node 14 as Node 12 will reach end-of-life is on 2022-04-30

The dashboard repo does not have a compatibility table, but if you look at the one for [Parse Server](https://github.com/parse-community/parse-server#nodejs), you can see that we support a Node version until the end of its LTS date. I opened https://github.com/parse-community/parse-dashboard/issues/1956 for that.

Issue #2007, opened 2022-01-21
Add UI tests to CI


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation
The lack of UI tests in the CI process makes it difficult to identify bugs before release. That leads to severe bugs likely to make it into release, e.g.:
- https://github.com/parse-community/parse-dashboard/issues/1991
- https://github.com/parse-community/parse-dashboard/issues/1994
- https://github.com/parse-community/parse-dashboard/issues/2006
- https://github.com/parse-community/parse-dashboard/issues/2194

The current manual testing before merging a PR is not reliable nor scalable.

### Feature / Enhancement Description
Add UI tests to CI, starting with just a few basic tests as guidance on how to write these tests, that can be extended over time.

### Example Use Case
n/a

### Alternatives / Workarounds
Manual testing which is not reliable or scalable.

### 3rd Party References
n/a

paulovitin:
@mtrezza, Are you talking about Cypress or Integration with [RTL](https://testing-library.com/docs/react-testing-library/intro/) ecosystem?

mtrezza:
See https://github.com/parse-community/parse-dashboard/pull/2030, something like that.

mehulmathur16:
Hi @mtrezza, is this issue still open ?

mtrezza:
Yep, it is, but there has been at least 1 PR: https://github.com/parse-community/parse-dashboard/pull/2030, not sure what the state of that is.

mehulmathur16:
Okay @mtrezza. I'll give this a try.

mtrezza:
@mehulmathur16 fantastic, please take a look at the discussion in #2030, it contains some feedback regarding how to implement this.

PR #2030, opened 2022-01-29
ci: add UI tests for improved release stability

damianstasik:

### Issue Description
This PR sets up a foundation for E2E testing along with two tests to have a starting point and verify whether CI jobs works correctly.

Related issue: #2007

### TODOs before merging

- [x] Remove old electron tests and deps
- [x] Add commands for testing during development
- [x] Include a few tests
- [x] Setup one server, but handle single app and multi-app configs


mtrezza:
Wow, I'm super excited about this! This would be a major milestone for dashboard tests and release stability.

damianstasik:
@mtrezza I think the core is ready, please take a look when you will have some time. I am not sure about developer guide for testing, is it clear enough?

.github/workflows/ci.yml
```diff
@@ -151,3 +151,26 @@ jobs:
...
+        working-directory: cypress
+      - name: Test dashboard
+        uses: cypress-io/github-action@v2
+        with:
+          start: npm run start
+          wait-on: 'http://localhost:4040'
+          wait-on-timeout: 120
+          browser: chrome
+      - name: Store video artifacts
+        if: always()
```
visualfanatic, line 172: Video artifacts are useful for debugging weird bugs, edge cases etc. and are always stored regardless if tests failed or not.

mtrezza:
This is amazing!

How could I run cypress using a locally running Parse Server deployment for debugging, instead of using docker?

damianstasik:
@mtrezza Make sure it is started on the default port, has app ID set to `hello` and master key to `world` (default values in the `Parse-Dashboard/parse-dashboard-config.json` config).

mtrezza:
Cypress is listening on port 4040, so I think there are some more steps necessary. Could you write a quick 1-2-3 instruction on how to run it with local Parse Server? I think we'd also need that in the contribution docs.

package.json
```diff
@@ -131,7 +132,9 @@
     "prepare": "webpack --config webpack/publish.config.js --progress",
     "start": "node ./Parse-Dashboard/index.js",
     "madge:circular": "node_modules/.bin/madge ./src --circular",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release",
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run"
```
Moumouls, line 138: Hi @visualfanatic, 
First of all thanks for this PR !  🙂 

I have set up Cypress in my company and we use it extensively to cover our apps , and our apps mainly use parse-server as the backend framework.

On my team monorepo, I made the choice to start with a completely native structure:
- Our backend is started by npm script natively, here we are also lucky we can start parse server by CLI, thanks to the `parse-server` command, if we install parse-server as a devDep
- Also parse-server plays very well with mongodb runner (here "pretest" command: https://github.com/parse-community/parse-server/blob/43cc97d99fa7346739f9e376a22afe3d95439b4c/package.json#L123 )

Here i can suggest may be remove docker need for local development and directly work in native env with parse-server CLI and mongodb runner.

I can also suggest to add 2 commands here:
- "test:watch": which start parse+ mongo and then open cypress in UI mode
- "test": which start parse + mongo and then run cypress (headless)

Then a developer will be able to just `clone`, `npm i`  and `npm test:watch` without docker or manually launching/waiting services to be up.

I can also suggest here this popular package to ensure parse server startup before lauching cypress: https://www.npmjs.com/package/start-server-and-test

Moumouls: With my team I am very sensitive to the developer experience, i think a native usage with a single command usage (all in one) could be more accessible for futur contributors and will facilitate new cypress tests contributions.

mtrezza:
@visualfanatic just a friendly ping, it would be great if you could pick up this PR, it would be a long-awaited feature.

package.json
damianstasik: I would not want to ditch the docker environment completely, it's just another option for people that prefer it. If we allow using a local backend for E2E tests we need to make sure all features required by tests are enabled and set up properly. It also changes mindset when you write a test, because you have to think about what other data there may be (or not) and how it can affect the test result. What if a flaky test overwrites some important data? I can think of a lot of similar concerns, that is why I would propose to allow both methods. Docker for those who prefer it and local server with a cautionary note about possible side effects. We could provide instructions for both methods with convenient all-in-one commands as you suggested.

@mtrezza @Moumouls what do you think?

mtrezza: I'd suggest to focus on the native usage at first. Docker would be an additional obstacle for contributors and could be a challenge especially for those who are just beginning to contribute. On GitHub it may slow down the CI significantly to the point where it's not feasible to run the CI for every PR commit. CI resources are not free, so there is also an economic aspect to it. Maybe docker env could be a complementary PR later on when we have more insight into the behavior of native testing and can tell more specifically whether/why docker is needed.

mtrezza: @damianstasik sorry for the long wait time, it slipped through my notifications that you were waiting for a response here. Do you think you could pick this PR up? We just made a release with an obvious bug today that could have been easily prevented with these UI tests. So I'm all eager to see this merged. I've also increased the bounty on the issue because I see you are working on this quite a bit.

Issue #2065, opened 2022-03-23
Add Parse Dashboard option `allowAnonymousUser`


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

Currently, Parse Dashboard relies on IP checking to check if a Dashboard is running remotely, and add additional security if so. This is problematic if you are using a proxy, which can result in protections becoming disabled.

### Feature / Enhancement Description

Rather than relying on IP checking for security, add Parse Dashboard option `allowAnonymousUser`. This would also change the behavior of being able to run Parse Dashboard locally without `allowInsecureHTTP` or `dev`.

### Example Use Case

-

### Alternatives / Workarounds

-

dblythy:
WIP 😃

PR #2066, opened 2022-03-23
feat: add dashboard option `allowAnonymousUser`

dblythy:

### Issue Description

Parse Dashboard assumes security if it detects a remote server, which is unreliable. This adds the option `allowAnonymousUser` which defaults to false.

Related issue: #2065

### Approach

- Adds a Parse Dashboard option `allowAnonymousUser` that defaults to false. Setting this to true allows Dashboard to run with no users.
- Adds ability to specify options via command line or via config file
- Adds console.error to frontend so solutions to errors can be identified by developers

This also add additional security for when dashboard is run on localhost with no options. Previously Dashboard was effectively assuming the `dev` parameter if localhost. This PR will only run Dashboard in `dev` mode if it's implicitly set. To get previous localhost behaviour, the `dev` parameter is required to be set.

### TODOs before merging


Parse-Dashboard/app.js
```diff
@@ -87,21 +87,29 @@ module.exports = function(config, options) {
...
-      //Based on advice from Doug Wilson here:
-      //https://github.com/expressjs/express/issues/2518
-      const requestIsLocal =
-        req.connection.remoteAddress === '127.0.0.1' ||
-        req.connection.remoteAddress === '::ffff:127.0.0.1' ||
-        req.connection.remoteAddress === '::1';
-      if (!options.dev && !requestIsLocal) {
-        if (!req.secure && !options.allowInsecureHTTP) {
+      for (const key in options) {
+        if (options[key] && !config[key]) {
```
mtrezza: Does this work if an option has value `false`?

Parse-Dashboard/app.js
```diff
@@ -87,21 +87,29 @@ module.exports = function(config, options) {
...
+      }
+      if (!config.dev) {
+        if (!req.secure && !config.allowInsecureHTTP) {
           //Disallow HTTP requests except on localhost, to prevent the master key from being transmitted in cleartext
-          return res.send({ success: false, error: 'Parse Dashboard can only be remotely accessed via HTTPS' });
+          return res.send({
+            success: false,
+            error:
+              "Parse Dashboard can only be remotely accessed via HTTPS.",
+            log: "Parse Dashboard can only be remotely accessed via HTTPS. If you are running locally, use the --dev parameter to bypass allowInsecureHTTP.",
```
mtrezza: I think "bypass" is ambiguous, it means "undefined state" in this case because it doesn't indicate whether the `allowInsecureHTTP` is true or false when it's "bypassed".

That wording is also used elsewhere in the docs. Maybe we can be more clear and say it "overrides the option to `false`" or something like that?

mtrezza (changes requested):

Parse-Dashboard/app.js
dblythy: Good thinking

mtrezza:
Still some checks failing. Please request another review once this is ready.

mtrezza:
Is this a breaking change? Does an existing deployment need a reconfiguration to maintain the same security?

dblythy:
I would say so, an existing localhost configuration would need to be changed but remote servers should function the same as before

mtrezza:
I guess our deprecation policy should be (or is in practice already) extending to Parse Dashboard. So breaking changes should probably happen accumulated once a year only as well. Otherwise we may have a breaking change every other month.

Is there a way to make this non-breaking? For example, if the new option is not set, fall back to the old behavior.

dblythy:
The only way I think is if we set `dev` to true by default, but that brings security implications

mtrezza:
Yes, I think we should avoid that

mtrezza:
Could you rebase this on alpha?

mtrezza (commented):
It seems the discussion got stuck at how to avoid making this a breaking change. Once this has been figured out please request another review.

PR #2257, opened 2022-09-08
feat: auto-submit one-time password (OTP) after entering

### LINKED ISSUE #2164: Auto submit OTP after entering

### Current Limitation
The OTP has to be submitted manually after entering it.

Since Parse Dashboard knows the used OTP algorithm from the settings, it knows the expected length of the OTP. After entering the last digit of the OTP, the form should auto-submit. Saves the developer a click. This should also work then copy/pasting the number from an auth app.

### Example Use Case
n/a

Manually submit form.

### 3rd Party References
n/a

dblythy:

### Issue Description

OTPs should autosubmit when they reach the required length

Related issue: #2164

### Approach

The MFA algorithm doesn't currently store `digits` or `period` - this PR fixes that and passes digits to fronttend.

### TODOs before merging

### DIFF
--- a/Parse-Dashboard/CLI/mfa.js
+++ b/Parse-Dashboard/CLI/mfa.js
@@ -189,10 +189,10 @@ module.exports = {
       if (algorithm !== 'SHA1') {
         data.mfaAlgorithm = algorithm;
       }
-      if (digits !== 6) {
+      if (digits != 6) {
         data.mfaDigits = digits;
       }
-      if (period !== 30) {
+      if (period != 30) {
         data.mfaPeriod = period;
       }
       showQR(data.url);
@@ -225,10 +225,10 @@ module.exports = {
     if (algorithm !== 'SHA1') {
       config.mfaAlgorithm = algorithm;
     }
-    if (digits !== 6) {
+    if (digits != 6) {
       config.mfaDigits = digits;
     }
-    if (period !== 30) {
+    if (period != 30) {
       config.mfaPeriod = period;
     }
     showInstructions({ app, username, secret, url, config });
--- a/Parse-Dashboard/CLI/mfa.js
+++ b/Parse-Dashboard/CLI/mfa.js
@@ -101,14 +101,14 @@ const showInstructions = ({ app, username, passwordCopied, secret, url, encrypt,
 
   if (secret) {
     console.log(
-      `\n${getOrder()}. Open the authenticator app to scan the QR code above or enter this secret code:` + 
-      `\n\n   ${secret}` + 
+      `\n${getOrder()}. Open the authenticator app to scan the QR code above or enter this secret code:` +
+      `\n\n   ${secret}` +
       '\n\n   If the secret code generates incorrect one-time passwords, try this alternative:' +
-      `\n\n   ${url}` + 
+      `\n\n   ${url}` +
       `\n\n${getOrder()}. Destroy any records of the QR code and the secret code to secure the account.`
     );
   }
-  
+
   if (encrypt) {
     console.log(
       `\n${getOrder()}. Make sure that "useEncryptedPasswords" is set to "true" in your dashboard configuration.` +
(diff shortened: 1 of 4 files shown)


Parse-Dashboard/CLI/mfa.js
```diff
@@ -189,6 +189,12 @@ module.exports = {
       if (algorithm !== 'SHA1') {
         data.mfaAlgorithm = algorithm;
       }
+      if (digits !== 6) {
```
mtrezza: Not sure if this PR is ready for review yet, but I think it should be `!=`? In case a future code change uses a number object for example.

dblythy: `inquirier.prompt` specifies a `type` of number, so !== is valid here. See:

https://github.com/parse-community/parse-dashboard/blob/7c56b757dfb6cd1dd48c75968137c16a31a9565e/Parse-Dashboard/CLI/mfa.js#L41-L54

mtrezza: But is a strict equality comparison necessary here? Isn't this just comparing with the value `6`?

mtrezza: Or maybe that strict equality comparison makes sense, because it prevents to interpret a `1` as a boolean `true` but instead always compares it with the number `1`, right?

dblythy: the variable `number` will always be a `Number` as inquirer will convert CLI output to a number before passing through (`type: 'number'`)

src/login/Login.js
```diff
@@ -16,8 +16,16 @@ export default class Login extends React.Component {
...
     let errorDiv = document.getElementById('login_errors');
+    let otpLength = 6;
     if (errorDiv) {
       this.errors = errorDiv.innerHTML;
+      try {
+        const json = JSON.parse(this.errors)
+        this.errors = json.text
+        otpLength = json.otpLength;
+      } catch (e) {
+        /* */
```
mtrezza: Empty catch block looks strange, we should actually add a lint rule to prevent that.

mtrezza (changes requested):

Parse-Dashboard/CLI/mfa.js
```diff
@@ -214,12 +220,17 @@ module.exports = {
...
     const config = { mfa: secret };
     if (algorithm !== 'SHA1') {
       config.mfaAlgorithm = algorithm;
     }
+    if (digits != 6) {
+      config.mfaDigits = digits;
+    }
+    if (period != 30) {
+      config.mfaPeriod = period;
+    }
```
mtrezza: This block seems to be identical to another block in the same file. Can this block be refactored into an external method?

mtrezza (changes requested):

mtrezza:
Note for testing: This should work when entering digit by digit and when copy/pasting the whole code at once like from an auth app.

src/login/Login.js
```diff
@@ -24,7 +24,7 @@ export default class Login extends React.Component {
         this.errors = json.text
         otpLength = json.otpLength;
       } catch (e) {
-        /* */
+        console.log(`could not pass error json: ${e}`);
```
mtrezza: Not sure what this code block does, but if this fails to parse, how would the user know there was an error unless they looked at the console? Or isn't that relevant to the user?

mtrezza:
Is this ready for review?

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [4.2.0-alpha.12](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0-alpha.12)

parseplatformorg:
🎉 This change has been released in version [4.2.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [4.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0)

PR #2264, opened 2022-09-14
feat: add column name to related records

### LINKED ISSUE #2191: Related records in context menu ambiguous
### New Issue Checklist

### Issue Description

Right clicking on a pointer in the data browser shows the menu "Get related records" where it lists all classes with a pointer of the same class. If a class has multiple pointers of the same class, the class shows up multiple times in the list, without indication which fields is referenced with each entry.

### Steps to reproduce
1. Create class `Test` with 2 pointer fields for class `_User`: `userA`, `userB`.
2. Open class `_User`.
3. Right click on `objectId` > `Get related records`.

### Actual Outcome
There are two entries in list for `Test` class without indication which entry corresponds to `userA` and which to `userB`.

### Expected Outcome
There are two entries in list for `Test` class with the field name next to it, for example:
- `Test - userA`
- `Test - userB`

### Environment

Dashboard
- Parse Dashboard version: `4.1.2`

dblythy:

### Issue Description

Related records can be hard to tell what field it is relating to. Especially difficult if a class has multiple pointers of the same column type

Related issue: #2191
Closes: #2191

### Approach

Adds field name in related records

### TODOs before merging

- [x] Add tests

### DIFF
--- a/src/components/BrowserCell/BrowserCell.react.js
+++ b/src/components/BrowserCell/BrowserCell.react.js
@@ -335,7 +335,7 @@ export default class BrowserCell extends Component {
         cl.forEach((column, field) => {
           if (column.targetClass !== pointerClassName) { return; }
           relatedRecordsMenuItem.items.push({
-            text: className, callback: () => {
+            text: `${className} - ${field}`, callback: () => {
               let relatedObject = value;
               if (this.props.field === 'objectId') {
                 relatedObject = new Parse.Object(pointerClassName);
--- a/src/components/BrowserCell/BrowserCell.react.js
+++ b/src/components/BrowserCell/BrowserCell.react.js
@@ -335,7 +335,7 @@ export default class BrowserCell extends Component {
         cl.forEach((column, field) => {
           if (column.targetClass !== pointerClassName) { return; }
           relatedRecordsMenuItem.items.push({
-            text: `${className} - ${field}`, callback: () => {
+            text: `${className}`, subtext: `${field}`, callback: () => {
               let relatedObject = value;
               if (this.props.field === 'objectId') {
                 relatedObject = new Parse.Object(pointerClassName);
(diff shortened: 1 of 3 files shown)


mtrezza:
Could you add a screenshot how this looks like?

dblythy:
Screen Shot 2022-09-14 at 7 28 44 pm

mtrezza:
I just tried it out with a long list and it becomes more difficult to quickly find a class in the list because it's now a lot of text.

Is it possible to visually differentiate the class name from the field name?
For example, by making the field name text color less prominent:

image

I assume most classes usually have only 1 entry in the list anyway, so the field name is in most cases unnecessary information.

The same approach is used in the column header (although I don't think this low of a contrast is accessibility compliant):

image

dblythy:
I will see what I can do

mtrezza:
Maybe a `<span style="...">` tag? I tried to add it, but it's written as is to the list item. So there's probably some encoding logic that needs to be changed. If it's not possible we can also merge as it and improve that at a later time.

dblythy:
Screen Shot 2022-09-17 at 5 53 19 pm

mtrezza (approved):
Much better! Nicely done with the color style

parseplatformorg:
🎉 This change has been released in version [4.2.0-alpha.13](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0-alpha.13)

parseplatformorg:
🎉 This change has been released in version [4.2.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [4.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/4.2.0)

Issue #970, opened 2019-02-03
Node Clustering results in prompt 'form tampered with' even with cookieSessionSecret set


MobileVet:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Dashboard!

- [X] Running version = 1.2.0 of Dashboard.

- [X] Running version = 2.8.2 of Parse Server.

- [X] You've searched through [existing issues](https://github.com/ParsePlatform/parse-dashboard/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before. **Lots on CSRF but NONE related to Clustering w/ Throng when `cookieSessionSecret` is set or similar**

#### Environment Setup

Heroku or any other load balanced system

Node clustering via the **Throng** module
Dashboard Option `cookieSessionSecret` set

#### Setup

Setup a Parse Server w/ the Dashboard grafted on to the Express app and pass it to **Throng**

```
const dashboardConfig = ({
  apps: [{
    appId: config.APP_ID,
    appName: config.APPLICATION_NAME,
    masterKey: config.MASTER_KEY,
    production,
    serverURL: config.SERVER_URL,
    supportedPushLocales: ["en"]
  }],
  port: config.PORT,
  trustProxy: 1,
  users: [{
    user: config.DASHBOARD_USERNAME,
    pass: config.DASHBOARD_PASSWORD
  }]
})

const dashboardOptions = ({
  options: {
    allowInsecureHTTP: false,
    cookieSessionSecret: config.DASHBOARD_SECRET
  }
})

throng({
  workers: config.WORKERS,
  lifetime: Infinity
}, start)

function start() {

  // Create our Express app and force HTTPS
  const app = express()
  app.use(requireHTTPS)

  app.use('/dashboard', dashboard(dashboardConfig, dashboardOptions))
  app.use(config.MOUNT_PATH, parseServer.create())
  app.listen(config.PORT, displayEnvironment)
}
```

#### Steps to reproduce

I have found that `cookieSessionSecret` will resolve the CSRF issue when running >1 dyno on Heroku or other distributed systems... UNLESS you are Clustering.

If you are clustering, via Throng for example, even if you only use a single 'worker' you will NOT be able to use >1 dyno.  Additionally, if you cluster with >1 worker, you can't access the Dashboard.

So, if Clustering, the only way the dashboard will be accessible is if you have a **single** dyno with a **single** worker.  Otherwise you get **'form tampered with'**

[...]

Issue #806, opened 2017-11-04
Prompt form tampered with, what is the problem?


lxq20081128:
Prompt form tampered with, what is the problem?

lxq20081128:
v1.1.2 version of the problem, with the https prompt tampered with form, with http and added to allow set, suggesting that must be https

fxchou123:
same issue here, I'm using Heroku as server

stage88:
In my case I fixed it by changing from 

```
let dashboard = new ParseDashboard({
  ...
}, true);
```

to 

```
let dashboard = new ParseDashboard({
  ...
}, {
  allowInsecureHTTP: true
});
```

mmarshak:
Same issue, I get that error. And it does not allow me to access the dashboard.

lxq20081128:
Hi, I used what you said, still does not work in v1.1.2 version.






------------------
              
         公司网址：http://www.mengsens.com/
         
         
                      广东梦森信息科技有限公司 -              廖晓清
             公司地址：广州市海珠区新港西路135号中山大学国家大学科技园B座1308室
         
         手机号码：139 2609 5308
         QQ号码 ：187 2117 737
         主营业务：快递鸽 - 快递物流综合解决方案
     
     
     
     祝：生活愉快！一切顺利！
 

 
 
 
------------------ Original ------------------

stage88:
@lxq20081128 Sorry that was my mistake, it actually doesn't work even if you what I suggested, which is why I deleted my comment :)

Hopefully we can get a response from one of the contributors as this is currently preventing me from upgrading.

montymxb:
@lxq20081128 I'm assuming you're referrig to #810? There were some pretty good details provided in that issue, I'm following up there currently. Let me know if this is something else though.

stage88:
Hi @montymxb, I think this issue is a different one, here is my config:
```
// configure Parse dashboard
let dashboard = new ParseDashboard({
  'apps': [{
    'serverURL': env.parseServerUrl,
    'appId': env.appId,
    'masterKey': env.masterKey,
    'appName': 'APP NAME'
  }],
  'users': [{
    'user': 'whatever',
    'pass': env.dashboardAdminPassword
  }],
  'useEncryptedPasswords': false,
  'trustProxy': true
}, {
  allowInsecureHTTP: true // HTTPS terminated by NGINX
});
```

Here is what displays after I successfully log in:
[...]

PR #2314, opened 2022-10-07
fix: cookieSessionSecret cannot be set in Dashboard options

dblythy:

### Issue Description

`cookieSessionSecret` is not infered from `config`, from options only (which is only set from CLI). This means dashboard + users breaks when using clusters.

Related issue: #970

### Approach

Adds cookieSessionSecret

### TODOs before merging


dblythy:
Nevermind, this is set via:

https://github.com/parse-community/parse-dashboard/issues/806#issuecomment-357577517

PR #2275, opened 2022-09-19
fix: view relation dialog requires browser refresh when navigating

dblythy:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description

Relation requires page reload when navigating backwards and forwards

Related issue: #2171
Closes: #2171

### Approach

Loads pointers on setting relation, and loads data when going back

### TODOs before merging

- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

PR #2317, opened 2022-10-12
fix: using browser navigation backward / forward button clears data browser

### LINKED ISSUE #2310: Using browser navigation clears data browser
### New Issue Checklist

### Issue Description

Navigating using the web browser's "back" and "forward" buttons doesn't shows empty table in the data browser. The data browser briefly shows a loading screen, then the rows in the data browser, then clears the data browser.

It also seems that the dashboard is at times completely reloading itself. This causes the whole dashboard to be reloaded from the server, which seems unnecessary.

It seems, this bug has been introduced with https://github.com/parse-community/parse-dashboard/pull/2273.

### Steps to reproduce

Navigate by clicking on a pointer in a row in the data browser, then again on a pointer, thus drilling down in the data hierarchy. Then navigate back and forth using the web browser buttons.

### Actual Outcome

The data browser is empty.

### Expected Outcome

The data browser should show the correct data.

### Environment

Dashboard
- Parse Dashboard version: `5.0.0-alpha.3`

damianstasik:

### Issue Description
This should fix the bug from the attached issue.

Related issue: #2310

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -150,6 +150,12 @@ class Browser extends DashboardView {
     this.redirectToFirstClass = this.redirectToFirstClass.bind(this);
 
     this.dataBrowserRef = React.createRef();
+
+    window.addEventListener('popstate', () => {
+      this.setState({
+        relation: null
+      })
+    });
   }
 
   componentWillMount() {
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -150,13 +150,6 @@ class Browser extends DashboardView {
     this.redirectToFirstClass = this.redirectToFirstClass.bind(this);
 
     this.dataBrowserRef = React.createRef();
-    window.addEventListener('popstate', () => {
-      this.setState({
-        relation: null,
-        data: null,
-      })
-      this.refresh();
-    });
   }
 
   componentWillMount() {


src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -150,13 +150,6 @@ class Browser extends DashboardView {
     this.redirectToFirstClass = this.redirectToFirstClass.bind(this);
 
     this.dataBrowserRef = React.createRef();
-    window.addEventListener('popstate', () => {
```
mtrezza, line 153: This was added to fix another bug, see https://github.com/parse-community/parse-dashboard/pull/2275. @dblythy It seems to cause navigation issues?

mtrezza (commented):
We're waiting for a comment from @dblythy since he recently added the lines of code this PR would remove.

src/dashboard/Data/Browser/Browser.react.js
dblythy: The reason this code was added is that the data doesn't reload when going from `http://localhost:4040/apps/abcd/browser/TestObject/qiivFSIcz2/relation` to `http://localhost:4040/apps/abcd/browser/TestObject`. 

It was also added to make the back button when clicking "view relation" work.

I can't seem to replicate the original bug, do you have any suggestions?

mtrezza: Maybe to replicate the bug, some latency is required for the data to load; if you run dashboard and server on the same machine, you may not experience it. To increase latency you could use a remote server or artificially add latency to your network. Sometimes navigating seems to work, then 2-5 seconds later the data browser is cleared, hence I think it's latency related. Try to add a filter, then in the results click on a pointer to jump to another class, then navigate back.

mtrezza: This bug is currently blocking the beta release

dblythy: Investigating today

dblythy: changing to:

```
window.addEventListener('popstate', () => {
      this.setState({
        relation: null
      })
    });
```

Seems to fix the issue whilst maintaining back arrow functionality

dblythy (changes requested):

src/dashboard/Data/Browser/Browser.react.js
mtrezza: Were you able to replicate the issue?

dblythy: Yes I was on a Parse Server instance running the dashboard in the UK which added some latency, and I observed the bug

mtrezza: Nice! Let me see whether I can still observe the issue...

mtrezza: Nice! Let me see whether I can still observe the issue with this fix. @damianstasik could you adapt the PR to the suggested changes we can try this out?

mtrezza: I've committed the change

mtrezza (approved):
Looks good; I wasn't able to replicate the issue anymore

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.7](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [5.0.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.0.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.0.0)

PR #2353, opened 2022-12-26
fix: Long column names are cut off in the "Manage Columns" menu

### LINKED ISSUE #1892: "Manage Columns" tab cuts off long column names
### New Issue Checklist

### Issue Description
The tap that opens when clicking the "Manage Columns" menu has a fixed size and therefore cuts off long column names.

### Steps to reproduce
1. Create a column in a class that is longer than the width of the "Manage Columns" tab
2. Open the "Manage Columns" tab

### Actual Outcome
Column name is cut off

### Expected Outcome
Tab should adapt its width to fit the longest column name, maybe up to a max width, and only then truncate the name.

### Environment

Dashboard
- Parse Dashboard version: 3.2.1

### Logs
n/a

Magdaraa:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues/1892)
### Issue Description

The tap that opens when clicking the "Manage Columns" menu has a fixed size and therefore cuts off long column names

Closes: https://github.com/parse-community/parse-dashboard/issues/1892

### Approach

Column name isn't cut off, tab is adapt to the width 

I did it with help from @404-html


mtrezza:
@Magdaraa, thanks for the PR and apologies for the long wait for review. It looks good, just a few observations:

image

1. There is still space at the right side of the menu for the tab to expand, but it doesn't. If that is too difficult to enable, then no worries.
2. When a name is truncated `abcdef...` , it would be to truncate the middle of the string, like `abc...xyz`. This allows to distinguish two strings that are both truncated and are only different in the end.

mtrezza (commented):
@Magdaraa, thanks for the PR and apologies for the long wait for review. It looks good, just a few observations:

image

1. There is still space at the right side of the menu for the tab to expand, but it doesn't. If that is too difficult to enable, then no worries.
2. When a name is truncated `abcdef...` , it would be to truncate the middle of the string, like `abc...xyz`. This allows to distinguish two strings that are both truncated and are only different in the end.

Whatever you could improve of the above would be fantastic, otherwise we can also merge as is, as it's already an improvement from before. And the lint check fails, should be easy to fix.

PR #2361, opened 2023-01-19
feat: Add export all rows of a class and export in JSON format

### LINKED ISSUE #1961: Can't export more than 10k rows as csv

### Current Limitation

**export all rows** only exports first 10k objects or rows in csv.

Should be able to export all data in a class.

### Example Use Case

1. user has a class with more than 10k data and wants to export all of the rows in an csv format.

First export all data in a class as json format selecting **export this data** option from **Edit** menu. Then user have to convert that data to csv format.

dblythy:

### Issue Description

It's currently only possible to export 10,000 rows, and only as a CSV

Closes: #1961

### Approach

Uses `Parse.Query.eachBatch` to paginate over full class and exports as files when:

- Array of objects is over 1gb in memory
- New objects length is less than batch size (`eachBatch` has finished)

Screenshot 2023-01-19 at 3 45 34 pm

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -1262,102 +1262,116 @@ class Browser extends DashboardView {
     const className = this.props.params.className;
     const query = new Parse.Query(className);
 
-    if (!rows['*']) {
+    if (!rows["*"]) {
       // Export selected
       const objectIds = [];
       for (const objectId in this.state.rowsToExport) {
         objectIds.push(objectId);
       }
-      query.containedIn('objectId', objectIds);
+      query.containedIn("objectId", objectIds);
       query.limit(objectIds.length);
     }
 
-
     const processObjects = (objects) => {
-    const classColumns = this.getClassColumns(className, false);
-    // create object with classColumns as property keys needed for ColumnPreferences.getOrder function
-    const columnsObject = {};
-    classColumns.forEach((column) => {
-      columnsObject[column.name] = column;
-    });
-    // get ordered list of class columns
-    const columns = ColumnPreferences.getOrder(
-      columnsObject,
-      this.context.applicationId,
-      className
-    ).filter(column => column.visible);
+      const classColumns = this.getClassColumns(className, false);
+      // create object with classColumns as property keys needed for ColumnPreferences.getOrder function
+      const columnsObject = {};
+      classColumns.forEach((column) => {
+        columnsObject[column.name] = column;
+      });
+      // get ordered list of class columns
+      const columns = ColumnPreferences.getOrder(
+        columnsObject,
+        this.context.applicationId,
+        className
+      ).filter((column) => column.visible);
 
-    if (type === '.json') {
-      const element = document.createElement('a');
-      const file = new Blob([JSON.stringify(objects.map(obj => {
-        const json = obj._toFullJSON();
-        delete json.__type;
-        return json;
-      }), null, 2)], { type: 'application/json' });
+      if (type === ".json") {
+        const element = document.createElement("a");
+        const file = new Blob(
+          [
+            JSON.stringify(
+              objects.map((obj) => {
+                const json = obj._toFullJSON();
+                delete json.__type;
+                return json;
+              }),
+              null,
+              2
+            ),
+          ],
+          { type: "application/json" }
+        );
+        element.href = URL.createObjectURL(file);
+        element.download = `${className}.json`;
+        document.body.appendChild(element); // Required for this to work in FireFox
+        element.click();
+        document.body.removeChild(element);
+        return;
+      }
+
+      let csvString = columns.map((column) => column.name).join(",") + "\n";
+      for (const object of objects) {
+        const row = columns
+          .map((column) => {
+            const type = columnsObject[column.name].type;
+            if (column.name === "objectId") {
+              return object.id;
+            } else if (type === "Relation" || type === "Pointer") {
+              if (object.get(column.name)) {
+                return object.get(column.name).id;
+              } else {
+                return "";
+              }
+            } else {
+              let colValue;
+              if (column.name === "ACL") {
+                colValue = object.getACL();
+              } else {
+                colValue = object.get(column.name);
+              }
+              // Stringify objects and arrays
+              if (
+                Object.prototype.toString.call(colValue) ===
+                  "[object Object]" ||
+                Object.prototype.toString.call(colValue) === "[object Array]"
+              ) {
+                colValue = JSON.stringify(colValue);
+              }
+              if (typeof colValue === "string") {
+                if (colValue.includes('"')) {
+                  // Has quote in data, escape and quote
+                  // If the value contains both a quote and delimiter, adding quotes and escaping will take care of both scenarios
+                  colValue = colValue.split('"').join('""');
+                  return `"${colValue}"`;
+                } else if (colValue.includes(",")) {
+                  // Has delimiter in data, surround with quote (which the value doesn't already contain)
+                  return `"${colValue}"`;
+                } else {
+                  // No quote or delimiter, just include plainly
+                  return `${colValue}`;
+                }
+              } else if (colValue === undefined) {
+                // Export as empty CSV field
+                return "";
+              } else {
+                return `${colValue}`;
+              }
+            }
+          })
+          .join(",");
+        csvString += row + "\n";
+      }
+
+      // Deliver to browser to download file
+      const element = document.createElement("a");
+      const file = new Blob([csvString], { type: "text/csv" });
       element.href = URL.createObjectURL(file);
-      element.download = `${className}.json`;
+      element.download = `${className}.csv`;
       document.body.appendChild(element); // Required for this to work in FireFox
       element.click();
       document.body.removeChild(element);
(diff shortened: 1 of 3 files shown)


src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -28,13 +35,26 @@ export default class ExportSelectedRowsDialog extends React.Component {
         type={Modal.Types.INFO}
         icon='warn-outline'
         title={this.props.selection['*'] ? 'Export all rows?' : (selectionLength === 1 ? 'Export 1 selected row?' : `Export ${selectionLength} selected rows?`)}
-        subtitle={this.props.selection['*'] ? 'Note: Exporting is limited to the first 10,000 rows.' : ''}
+        subtitle={this.props.selection['*'] ? 'Note: This will export mutliple files with maximum 1gb each. This might take a while.' : ''}
```
mtrezza:
```suggestion
        subtitle={this.props.selection['*'] ? 'Large datasets are exported as multiple files of up to 1 GB each.' : ''}
```

mtrezza (changes requested):

mtrezza (dismissed):
Looks good!

src/dashboard/Data/Browser/Browser.react.js
```diff
@@ -1255,15 +1257,12 @@ class Browser extends DashboardView {
...
+  async confirmExportSelectedRows(rows, type, indentation) {
+    this.setState({ rowsToExport: null, exporting: true, exportingCount: 0 });
     const className = this.props.params.className;
     const query = new Parse.Query(className);
 
-    if (rows['*']) {
-      // Export all
-      query.limit(10000);
-    } else {
+    if (!rows['*']) {
```
mtrezza, line 1306: What limit if any is applied if `rows['*'] == true`?

dblythy: `query.limit(objectIds.length);` (line 1272)

mtrezza: But that's in the if block:

https://github.com/parse-community/parse-dashboard/blob/46b1346f20e01d28e772e5ef8e1706c8ce63e917/src/dashboard/Data/Browser/Browser.react.js#L1265-L1273

What happens if the condition `!rows['*']` is false, what limit is then applied?

dblythy: Oh right, it uses query.eachBatch, which does not accept a limit (so unlimited)

mtrezza: So it will export all objects of a class? In what case for example is the condition false?

dblythy: If `rows["*"]` is export all, so the inverse is selected rows only

mtrezza: Can you help understand the logic of that if block. It seems to me it means if only selected rows should be exported (`!rows[“*”]`), then set the limit in the block. Inversely it implies that if all rows of a class should be exported, then do not set any limit.

If that’s how it works, then
- why do we need to set a limit at all if the limit can already be inferred from the selected objectIds? In other words, if object A and B are selected for export, then we’ll never export more than 2 objects, so why do we need a limit?
- if all rows of a class should be exported (no rows selected), then there is no limit set. If a class has a high number of records, then that can have unintended consequences, i.e bring down the DB or cause high data traffic costs

dblythy: - the limit code is existing from the previous behaviour of export selected, I’m assuming it’s to override any server default limit. 
- it’s not possible to limit `eachBatch`, it automatically iterates across the whole class.

mtrezza: > I’m assuming it’s to override any server default limit

makes sense

> it’s not possible to limit eachBatch, it automatically iterates across the whole class

I think such a query requires some kind of safety confirmation. The 10k was a safety limit. Maybe we should add a confirmation dialog if >10k rows will be exported? If there are rows already loaded in the data browser, we could also calculate the average row data size and multiply that by the total row count.

So the message could be like:

> Do you really want to export all rows?
> Estimated row count: 12,345
> Estimated data size: 0.5 GB

dblythy: How about a "confirm export all" field and a warning that the function can cause database slowness?

dblythy:
Screen Shot 2023-01-23 at 10 19 15 am

src/dashboard/Data/Browser/Browser.react.js
mtrezza: I would keep it in a single dialog, instead of showing an additional warning dialog after the previous dialog. Like:

> ⚠️ Exporting all rows may significantly impact database performance.

mtrezza (changes requested):
Nice rework of the dialog!

I looked at the delete rows dialog, to use the same wording here:

> **Confirm this action**
> Enter "export all" to continue.

Unlike a delete action, it may not be clear why this is a critical operation. So in addition we could add a text that says:

> ⚠️ Exporting all rows may significantly impact resources.

Maybe you could add that below the "Estimated data size", or if it looks acceptable add it to the confirmation segment:

> **Confirm this action**
> ⚠️ Exporting all rows may significantly impact resources. Enter "export all" to continue.

dblythy:
Screenshot 2023-01-25 at 10 51 31 am

How's this look?

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -5,36 +5,95 @@
...
   render() {
     let selectionLength = Object.keys(this.props.selection).length;
+    const fileSize = new TextEncoder().encode(JSON.stringify(this.props.data, null, this.state.exportType === '.json' && this.state.indentation ? 2 : null)).length / this.props.data.length
     return (
       <Modal
         type={Modal.Types.INFO}
         icon='warn-outline'
         title={this.props.selection['*'] ? 'Export all rows?' : (selectionLength === 1 ? 'Export 1 selected row?' : `Export ${selectionLength} selected rows?`)}
-        subtitle={this.props.selection['*'] ? 'Note: Exporting is limited to the first 10,000 rows.' : ''}
+        subtitle={this.props.selection['*'] ? 'Large datasets are exported as multiple files of up to 1 GB each.' : ''}
```
mtrezza:
```suggestion
```

I think this info is true in any case; I'd move this down to the dialog text.

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -5,36 +5,95 @@
...
         disabled={!this.valid()}
         confirmText='Export'
         cancelText='Cancel'
         onCancel={this.props.onCancel}
-        onConfirm={this.props.onConfirm}>
-        {}
+        onConfirm={() => this.props.onConfirm(this.state.exportType, this.state.indentation)}>
+        {this.props.selection['*'] && <div className={styles.row} >
+          <Label text="Do you really want to export all rows?" description={<span className={styles.label}>Estimated row count: {this.props.count}<br/>Estimated export size: {this.formatBytes(fileSize * this.props.count)}<br/>⚠️ Exporting all rows may significantly impact resources.</span>}/>
```
mtrezza:
```suggestion
          <Label text="Do you really want to export all rows?" description={<span className={styles.label}>Estimated row count: {this.props.count}<br/>Estimated export size: {this.formatBytes(fileSize * this.props.count)}<br/><br/>⚠️ Exporting all rows may severely impact server or database resources.<br/>Large datasets are exported as multiple files of up to 1 GB each.</span>}/>
```

mtrezza (changes requested):

src/dashboard/Data/Browser/ExportSelectedRowsDialog.react.js
```diff
@@ -5,36 +5,94 @@
...
+            </Dropdown>
+          } />
+        {this.state.exportType === '.json' && <Field
+          label={<Label text='Indentation' />}
+          input={<Toggle value={this.state.indentation} type={Toggle.Types.YES_NO} onChange={(indentation) => {this.setState({indentation})}} />} />
+          }
+        {this.props.selection['*'] && <Field
+          label={
+            <Label
+              text='Confirm Export all'
```
mtrezza:
```suggestion
              text='Confirm this action'
```

mtrezza (approved):
Looks good!

dblythy:
Thanks for your help!

parseplatformorg:
🎉 This change has been released in version [5.1.0-alpha.5](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0-alpha.5)

parseplatformorg:
🎉 This change has been released in version [5.1.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0)

Issue #2374, opened 2023-01-27
Add PR deployment preview


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

Reviewing a dashboard PR requires more effort than PRs from other repos because 
- Parse Dashboard is a visual product with UI
- We don't have UI testing as part of the CI

The dashboard PR has to be cloned locally in a development environment to see the effect of most PRs.

### Feature / Enhancement Description

Add a preview deployment to easily review and preview PRs online without having to clone the PR locally.

See https://docs.uffizzi.com/references/uffizzi-ci/#2-install-authorize-the-uffizzi-github-app

### Example Use Case

n/a

### Alternatives / Workarounds

Clone locally.

### 3rd Party References
n/a

### Uffizzi Feedback

- There seems to be an URL issue, see https://github.com/parse-community/parse-dashboard/pull/2370#issuecomment-1406928232:
  > https://https://app.uffizzi.com/projects/5259/deployments

- It would be nice to have all files in a `.uffizzi` directory. Currently we have:
  ```sh
  ./docker-compose.uffizzi.yml
  ./nginx-uffizzi/...
  ```

- When Uffizzi adds a comment it should probably delete any existing Uffizzi comments, otherwise the thread gets full of comments, and these comments are quite long. 

- While deploying a new commit, the preview URL shows an nginx error page; it's unclear whether the deployment / PR has an issue; it would be better to show a "deployment is progress" page
![image](https://user-images.githubusercontent.com/5673677/215320606-5d78d8e9-e121-4bae-bf40-f44432770a33.jpeg)

mtrezza:
@gadkins, @ShrutiC-git

I've opened this issue to the get Uffizzi integration working. #2364 has been merged.

I've tried existing PR https://github.com/parse-community/parse-dashboard/pull/2370, closed and re-opened it. ~~Uffizzi shows a deployment for that PR, but it didn't post a comment in the PR thread with the deployment URL. Does that only work for new PRs?~~ Uffizzi posted a comment with the URL, so it looks good.

[...]

PR #2375, opened 2023-01-27
fix: Dashboard may display blank page when selecting an app after login

dblythy:

### Issue Description

Related: https://github.com/parse-community/parse-dashboard/issues/2374

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/Parse-Dashboard/app.js
+++ b/Parse-Dashboard/app.js
@@ -173,7 +173,7 @@ module.exports = function(config, options) {
     }
 
     app.get('/login', csrf(), function(req, res) {
-      const redirectURL = req.url.includes('?redirect=') && req.url.split('?redirect=')[1];
+      const redirectURL = req.url.includes('?redirect=') && req.url.split('?redirect=')[1].length > 1 && req.url.split('?redirect=')[1];
       if (!users || (req.user && req.user.isAuthenticated)) {
         return res.redirect(`${mountPath}${redirectURL || 'apps'}`);
       }
--- a/Parse-Dashboard/app.js
+++ b/Parse-Dashboard/app.js
@@ -207,7 +207,11 @@ module.exports = function(config, options) {
     // For every other request, go to index.html. Let client-side handle the rest.
     app.get('/*', function(req, res) {
       if (users && (!req.user || !req.user.isAuthenticated)) {
-        return res.redirect(`${mountPath}login?redirect=${req.url.replace('/login', '')}`);
+        const redirect = req.url.replace('/login', '');
+        if (redirect.length > 1) {
+          return res.redirect(`${mountPath}login?redirect=${redirect}`);
+        }
+        return res.redirect(`${mountPath}login`);
       }
       if (users && req.user && req.user.matchingUsername ) {
         res.append('username', req.user.matchingUsername);
--- a/src/dashboard/DashboardView.react.js
+++ b/src/dashboard/DashboardView.react.js
@@ -32,11 +32,8 @@ export default class DashboardView extends React.Component {
   }
 
   onRouteChanged() {
-    const appId = this.context.applicationId;
     const path = this.props.location?.pathname ?? window.location.pathname;
-    console.log({path, appId});
-    const route = path.split(appId)[1].split('/')[1];
-    console.log({route});
+    const route = path.split('apps')[1].split('/')[2];
     if (route !== this.state.route) {
       this.setState({ route });
     }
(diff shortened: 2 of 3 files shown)


dblythy:
I think this should fix https://github.com/parse-community/parse-dashboard/issues/2374#issuecomment-1407173616

dblythy:
@mtrezza I think this should fix the issue, how do we get the preview to show?

dblythy:
Does the pr preview auto-update when new commits are pushed? I'm not sure it does, as the dashboard has retained login between commits

mtrezza:
Yes it should, but it takes a while.

dblythy:
Do you have any tips for debugging? Seems to work fine locally but getting this error on the preview

Screen Shot 2023-01-28 at 11 50 15 am

mtrezza:
I think it has something to do with the URL in the browser that is parsed there? What you could do it output the value to the browser JS console that should be split so you see what value is causing this. My shot in the dark here is that it has something to do with the URL `https://pr-2375-deployment-12561-parse-dashboard.app.uffizzi.com/`, but that's just a guess.

mtrezza:
I think you are right, you may have to close and re-open the PR to redeploy the preview, then it takes about 15 mins.

mtrezza:
Here's the issue:

```js
const appId = 'parse';
const path = '/apps/MyParseApp/browser';
const route = path.split(appId)[1].split('/')[1]; // throws because path.split(appId)[1] == undefined
```

Is this a misconfiguration of the dashboard?

Btw I'm not sure the `route` logic works in all cases, for example if `appId` is `apps` like:
`apps/apps/browser/_Installation`

ShrutiC-git:
hey @mtrezza @dblythy ~

The previews should be updated (without having to close and reopen the PR) as you push commits to that open PR. It could be the case that @dblythy you were seeing a cached version of the preview..?

Once commits are pushed to the PR, this triggers build and deploy again — so the preview gets updated, or you'll have a new preview in case the old one was deleted/expired. 

Was the Uffizzi CI console unable to help here? Ideally, as the containers get rebuilt, the Uffizzi CI will indicate that the containers are building and when the preview was last updated. Were you all not seeing that?

dblythy:
I had to close and re-open, I tried clearing cache and reloading and didn't work.

This is ready for review

ShrutiC-git:
thanks for letting me know, @dblythy. 
The desired behavior is the preview gets updated with new commits. It could be a configuration issue. I'll take a look into what's going on.

gadkins:
@dblythy @mtrezza  

Let me clarify Uffizzi behavior:

- You do not need to close and reopen PRs every time you push a new commit
- Uffizzi is receiving webhooks on every commit, so it will rebuild your application each time.
- Each time it rebuilds a new commit, it will preserve the state from the previous build if volumes are mounted in the compose. This tells Uffizzi to preserve this data as long as this PR is open.
- If you close a PR, Uffizzi will destroy the preview and all its state. If you reopen the PR, you will get an fresh environment with empty volumes and a new comment. 
- On new commits, Uffizzi will update the existing comment on the PR. It will not post a new comment. The multiple comments you are seeing are caused by opening and closing the PR several times.
- You can see which commit is deployed in the [Uffizzi Dashboard](https://app.uffizzi.com/sign_in), along with the timestamp:

Screen Shot 2023-01-27 at 11 35 52 PM

mtrezza:
Let's try to keep the discussion about Uffizzi on https://github.com/parse-community/parse-dashboard/issues/2374 otherwise we have bits all over unrelated PRs

mtrezza:
I've opened https://github.com/parse-community/parse-dashboard/issues/2374 to discuss the Uffizzi integration. Let's try to keep the discussion there, to keep this topic consolidated and make it easier for others to follow. I will respond there.

mtrezza (approved):
Looks good! Thanks for the quick fix.

parseplatformorg:
🎉 This change has been released in version [5.1.0-alpha.7](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [5.1.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [5.1.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.1.0)

Issue #2403, opened 2023-03-10
Add filter patterns


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

Combinations of filters that are set in the data browser are lost when navigating away or reloading the dashboard.

### Feature / Enhancement Description

Add feature to set preset filter patterns via filter dialog and preset filter patterns via dashboard configuration.

### Example Use Case

A filter pattern that finds objects in the `_User` class where `createdAt >= 10 days ago` && `updatedAt <= 1 day ago`

The filer dialog could look like this, with existing buttons less verbose, so there is enough space for a 4th button:

<img width="535" alt="image" src="https://user-images.githubusercontent.com/5673677/224434026-8f469ee4-abea-4db0-bcf1-2088446976f4.png">

The dashboard config could look like this:
```
const dashboard = ParseDashboard({
  filters: [
      {
         title: 'Recently deleted users', // The filer pattern title in the filter menu
         classes: ['_User'] // The classes for which the filter will be available; should accept regex to allow for all with `.*`
         filter: [ // The filter pattern as array (is filter order preserved)?
           {
             key: 'createdAt',
             condition: 'gte', // Reuse the syntax that is already used internally
             value: '2022-03-01T00:00:00.000Z'
           },
           {
             key: 'updatedAt',
             condition: 'lte', // Reuse the syntax that is already used internally
             value: '2022-03-10T00:00:00.000Z'
           }
         ]
      }
  ],
});
```

### Alternatives / Workarounds

Enter filter pattern manually (status quo).

dblythy:
Is this similar to #2041?

mtrezza:
Uh, a year has passed and I've added it again; at least this one has a screenshot and a suggestion for the config syntax.

#2041 is similar but not the same; it's just more in the idea of a database view. Essentially the same functionality, but maybe a different layout suggestion. I've added a screenshot in the issue there, to show the different UI. Maybe both feature can coexist and share much of the config parsing logic?

mtrezza:
[...]

PR #2404, opened 2023-03-11
feat: Add filter views to save frequently used filters in data browser

### LINKED ISSUE #2041: Custom views with predefined filters

### Current Limitation
There is no way to save a filter setting in dashboard. Every time a class is opened, the filter has to be newly applied.

Allow to create custom views (new section in dashboard) that consist of a class with a pre-defined filter setting.

The views of a class could be accessible in an expandable sub menu of the class in the data browser:
image

The benefit would be that the number of objects is immediately visible, without even opening the view. And the view can be accessed with 1 click, compared to a menu in the data browser of the class with a dropdown, which would require 3 clicks to navigate to.

### Example Use Case
Given a class `Product` that contains all products, it is possible to can create a view named `Chairs >100$` that shows the `Product` class with pre-defined filters for `product type: chair` and `price: >100`.

Set filters manually ever ytime a class page is opened in data browser.

dblythy:

### Issue Description

Adds ability to save filters

Closes: #2041

### Approach

Saves predefined filters in localstorage with similar logic to `ColumnPreferences`, only under a new class `ClassPreferences`. This data might be used for future class specific features

### TODOs before merging

- [x] Add tests


dblythy:
Some of the changes are due to local version of the `prettier` - do we have a code linting / styler?

mtrezza:
You mean whether we have a lint script to auto-correct ?

dblythy:
Yep 😊

mtrezza:
It seems to be missing here, feel free to just throw it in with this PR, maybe same names as with Parse Server:

```
"lint": "flow && eslint --cache ./",
"lint-fix": "eslint --fix --cache ./",
```

dblythy:
https://user-images.githubusercontent.com/4974769/224459723-bd3da46c-83ab-4a81-8926-362db1769b0c.mov

Demo of the PR, i'll try do some work around making sure the sidebar shows `active` on the filtered tab

mtrezza:
This is interesting with Uffizzi, I think to preview this we'd have to deploy a config file and later remove it before merging the PR. That woudn't be ideal because we may forget to remove and release it.

src/lib/ClassPreferences.js
```diff
@@ -0,0 +1,40 @@
+const VERSION = 'v1'; // In case we ever need to invalidate these
```
mtrezza: I would give this an int value (without the 'v'), in case we want to check for ranges in the future.

dblythy:
I think Ufuzzi will be fine for this because I've done all the filter saving in local storage (similar to how column preferences are managed)

src/lib/ClassPreferences.js
```diff
@@ -0,0 +1,40 @@
...
+let cache = {};
```
mtrezza, line 2: I'm unsure how practical this type of ephemeral storage is for filter views. Maybe we could allow to export these filter definitions to copy/paste them into the dashboard config file to keep the filters persistently? This way the dashboard can be used as a convenient view designer, and the user can make the filter permanent through config deployment.

src/components/BrowserFilter/BrowserFilter.react.js
```diff
@@ -137,10 +161,18 @@ export default class BrowserFilter extends React.Component {
...
-                  value="Apply these filters"
-                  width="245px"
+                  value={this.state.confirmName ? 'Confirm' : 'Save filters'}
+                  width="120px"
+                  onClick={() => this.state.confirmName ? this.save() : this.setState({confirmName: true})}
+                />
+                <Button
+                  color="white"
+                  primary={true}
+                  value="Apply filters"
```
mtrezza: Can you remove the `filters` from all the buttons in the dialog? It's the filters dialog, so it's implicit that `Apply` means `Apply the filters`.

mtrezza (changes requested):

src/components/BrowserFilter/BrowserFilter.react.js
```diff
@@ -137,10 +161,18 @@ export default class BrowserFilter extends React.Component {
                 <Button
                   color="white"
                   primary={true}
-                  value="Apply these filters"
-                  width="245px"
+                  value={this.state.confirmName ? 'Confirm' : 'Save filters'}
```
mtrezza: Could you make the `Save` button not highlighted by default? There should only be 1 highlighted (default) button in a UI dialog.

mtrezza (changes requested):

src/components/BrowserFilter/BrowserFilter.react.js
```diff
@@ -118,7 +138,11 @@ export default class BrowserFilter extends React.Component {
                 renderRow={props => (
                   <FilterRow {...props} active={this.props.filters.size > 0} parentContentId={POPOVER_CONTENT_ID} />
                 )}
-              />
+                />
+                {this.state.confirmName && <Field
+                  label={<Label text='What should we this filter?' />}
```
mtrezza: Let's give these "views" a distinct name, this will help to linguistically distinguish between the filter dialog and the new filter view on the side. If "View" is too general, maybe "Filter view"?

Also, I find many dialogs and buttons to be too verbose in Parse Dashboard; maybe let's try to make them more concise, for example:

```suggestion
                  label={<Label text='Filter view name' />}
```

mtrezza (changes requested):
image

For the filter dialog, could you:
- Move the "Save" button the very left? It's a button that is used the least (once to save, then not anymore).

----

image

For the save dialog, could you:
- Remove buttons "Clear all", "Add filter", "Apply filter"
- Rename "Confim" to "Save"
- Add a "Cancel" or "Back" button on the left

mtrezza:
This PR seems to mix https://github.com/parse-community/parse-dashboard/issues/2041 and https://github.com/parse-community/parse-dashboard/issues/2403. https://github.com/parse-community/parse-dashboard/issues/2403#issuecomment-1464802813 may be important to consider here. We may need to find a way to differentiate between "saving a filter pattern" and "saving a filter view" in the UI.

Maybe a solution is to add a "more" button "..." to the very left that opens a dropdown where one can choose:
- `Download filter definition` // to copy/paste into config file
- `Upload filter definition` // same as loading a filter from config file, but allows to upload only the definition
- `Save filter view locally` // to save locally, i.e. ephemeral storage as it does now

Instead of the `Download filter definition` and `Upload filter definition` a button to toggle between an editable raw JSON definition of the current filter setting could be an alternative. So one can toggle from the UI view to the JSON view and just copy the definition, or paste it, or just edit it. Not sure how difficult that would be to implement, but it would make the file up/download logic unnecessary.

mtrezza:
I find this to be a good UI solution, much nicer than the arrow on the left

image

dblythy:
Perhaps we can add a settings page that allows exporting of certain configurations to be defined later on Parse Dashboard? An example is the `ColumnPreferences` feature, it's quite difficult to extract that at the moment

mtrezza:
I think that's an interesting idea. Could be a good workout until we find a concept for storing dashboard settings in the cloud.

src/lib/ClassPreferences.js
dblythy: Hoping after #2406, we can have a way to export `ClassPreferences` to be defined in the Dashboard config

mtrezza (changes requested):
There's a text alignment issue - the longer the filter name, the further left the leading edge of the text:

image

mtrezza (changes requested):
image


This PR also reveals a semantic issue - a filter condition is called "filter" as well. To fix this let's just simplify the buttons:

- "Clear al" to "Clear"
- "Add filter" to "Add"
- "Save filters" to "Save"
- "Apply filter" to "Apply"

It's a simple dialog and probably easy to infer what they mean.

mtrezza (changes requested):
image

> What should we **call** this filter?

mtrezza:
Before we get more conflicts, @dblythy could you take a look and try to get this ready for merge?

dblythy:
Screenshot 2023-06-09 at 11 51 54 am

Screenshot 2023-06-09 at 12 37 13 pm

I've done some work making the highlight and selection of the filter cleaner:

Screenshot 2023-06-09 at 12 35 48 pm

dblythy:
Uffuzi hasn't updated to the latest commit, let's see if reopening helps...

mtrezza (approved):
Looks good, and works in preview env; what a fine feature!

parseplatformorg:
🎉 This change has been released in version [5.2.0-alpha.12](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-alpha.12)

parseplatformorg:
🎉 This change has been released in version [5.2.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [5.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0)

PR #2436, opened 2023-05-25
feat: Add links to users and roles in ACL dialog and handle invalid entries

### LINKED ISSUE #2435: Browse to users and roles from PermissionsDialog

- [ X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ X] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

The PermissionsDialog current only shows the objectId and/or name of users and roles.
To quickly see user or role details a manual lookup in the respective classes is needed.
Selection and copy/paste of the objectId role name does not seem to be possible.
And even if it would, it would still be manual overhead for the simple task of traversing roles.

Make user and roles in the PermissionsDialog clickable and show the object in the browser.
Ideally this would happen in a 2nd browser window (tab) so the PermissionsDialog remains open for the selected class/object.

Even better would be a dedicated UI to traverse the ACL tree (role -> users/subroles) within the PermissionsDialog or some new dialog. This is however outside the scope of this feature request.

### Example Use Case

1) create a few roles with multiple users and subroles
2) add multiple users and subroles to each subrole
3) select any class/object and add the top level roles to the ACL
4) while inspecting the ACL, click on any user or role to browse that item in a new window

none (manual copy/paste doesn't work as text selection is disabled)
it's effectively not possible to quickly traverse an ACL

### 3rd Party References

Not investigated as this is relative basic/simple request.

FransGH:

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [X] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description

The PermissionsDialog currently offers no way to inspect users and roles. 

Closes: #2435 and #2437

### Approach

This makes roles and user in the most simple way clickable and open in a separate browser window (tab).
It avoids complex in-app navigation issues by not attempting to browse within the same window and offer an option to return back to the CLP dialog.

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/components/PermissionsDialog/PermissionsDialog.react.js
+++ b/src/components/PermissionsDialog/PermissionsDialog.react.js
@@ -1022,7 +1022,7 @@ export default class PermissionsDialog extends React.Component {
             </span>
           </p>
           <p className={styles.hint}>
-            username: <span className={styles.selectable} style={{color:type.user.name ? undefined : '#f00'}}>{type.user.name ?? 'USER NOT FOUND'}</span>
+            username: <span className={styles.selectable} style={{color:type.user.name ? undefined : '#f00'}}>{type.user.name ?? 'user not found'}</span>
           </p>
         </span>
       );
@@ -1036,7 +1036,7 @@ export default class PermissionsDialog extends React.Component {
             </span>
           </p>
           <p className={styles.hint}>
-            id: <span className={styles.selectable} style={{color:type.role.id ? undefined : '#f00'}}>{type.role.id ?? 'ROLE NOT FOUND'}</span>
+            id: <span className={styles.selectable} style={{color:type.role.id ? undefined : '#f00'}}>{type.role.id ?? 'role not found'}</span>
           </p>
         </span>
       );
--- a/src/components/PermissionsDialog/PermissionsDialog.react.js
+++ b/src/components/PermissionsDialog/PermissionsDialog.react.js
@@ -971,7 +971,21 @@ export default class PermissionsDialog extends React.Component {
     return output;
   }
 
+  urlForKey(key) {
+    let isRole = key.startsWith('role:')
+    let className = isRole ? '_Role' : '_User';
+    let field = isRole ? 'name' : 'objectId';
+    let value = isRole ? key.replace('role:', '') : key
+    let filters = JSON.stringify([{
+      field,
+      constraint: 'eq',
+      compareTo: value
+    }]);
+    return window.location.href.split('browser/')[0] + `browser/${className}?filters=${encodeURIComponent(filters)}`;
+  }
+
   renderRow(key, columns, types) {
+
     const pill = text => (
       <span className={styles.pillType}>
         <Pill value={text} />
@@ -982,14 +996,16 @@ export default class PermissionsDialog extends React.Component {
     const type = (types && types.get(key)) || {};
 
     let pointer = this.state.pointerPerms.has(key);
-    let label = <span>{key}</span>;
+    let label = <span><a target="_blank" href={this.urlForKey(key)} >{key}</a></span>;
 
     if (type.user) {
       label = (
         <span>
           <p>
             <span>
-              <span className={styles.selectable}>{type.user.id}</span>
+              <span className={styles.selectable}>
+                <a target="_blank" href={this.urlForKey(key)} >{type.user.id}</a>
+              </span>
               {pill('User')}
             </span>
           </p>
@@ -1005,7 +1021,7 @@ export default class PermissionsDialog extends React.Component {
           <p>
             <span>
               <span className={styles.prefix}>{'role:'}</span>
-              {type.role.name}
+                <a target="_blank" href={this.urlForKey(key)} >{type.role.name}</a>
             </span>
           </p>
           <p className={styles.hint}>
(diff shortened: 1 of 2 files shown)


FransGH:
The way the url is composed is a bit hackie and will clearly fail if the current location is not ../browser/..
Didn't dive that deep into the code to get a clean base url but happy with suggestions!

mtrezza:
Could you add a before / after screenshot so its easier to see where the changes are?

FransGH:
Screenshot from v3.2.1 vs latest alpha. Don't ask me what happened to the role-objectIds between these versions, I didn't take them out. The only visual change as a result of the PR is that the names are now blue (= default link color) as opposed to grey and the cursor changes to a pointer while hovering (standard link behaviour). 
PastedGraphic-2
PastedGraphic-1

mtrezza:
I've just tried it out, I think it's nicely done, for me the links are even better visible than in your screenshot:
image

mtrezza:
> The way the url is composed is a bit hackie and will clearly fail if the current location is not ../browser/..

@parse-community/dashboard any suggestion?

FransGH:
There appear to be two different layouts for the permission list. One shows the object id's the other doesn't. The one with the object Id's, like in your screenshot, looks better. I see both flavours on both the older and latest dashboard depending on the object's ACL.

This is now fixed, see #2437

FransGH:
image

mtrezza:
Great, could you explain, what is the "and handle invalid entries"?

FransGH:
> Great, could you explain, what is the "and handle invalid entries"?
Basically fixed a bug that caused the render function to not work as intended when just one of the role or user objects was not found. With 'handling' I mean that they now explicitly appear as 'not found' on the UI.

mtrezza:
Got it. That could have been a separate PR, but we can merge this as 1 since you it's already mixed.

In what cases do you think the URL approach could cause issues?

FransGH:
I haven't experienced any issues with the url approach so far. It really solved a major pain I was having while migrating from user to role based ACLs. Still like to improve how the url is composed so hoping for some feedback.

mtrezza:
Well if there is no feedback then let's merge as is. If it's not working someone will hopefully open an issue and we can look at it then.

Maybe you want to take a look at the code and see how that is solved at other places?

If you take a look at this for example:

https://github.com/parse-community/parse-dashboard/blob/52ca7fc064cca60c7c7474da7417b39d0e596dbd/src/dashboard/Data/Browser/Browser.react.js#L858

There's a `generatePath` import that may be helpful. It seems to be used to compose the URL from the current context.

FransGH:
Yep, that worked, thx!

src/components/PermissionsDialog/PermissionsDialog.react.js
```diff
@@ -1005,11 +1032,11 @@ export default class PermissionsDialog extends React.Component {
           <p>
             <span>
               <span className={styles.prefix}>{'role:'}</span>
-              {type.role.name}
+                <a target="_blank" href={this.urlForKey(key)} >{type.role.name}</a>
             </span>
           </p>
           <p className={styles.hint}>
-            id: <span className={styles.selectable}>{type.role.id}</span>
+            id: <span className={styles.selectable} style={{color:type.role.id ? undefined : '#f00'}}>{type.role.id ?? 'ROLE NOT FOUND'}</span>
```
mtrezza:
```suggestion
            id: <span className={styles.selectable} style={{color:type.role.id ? undefined : '#f00'}}>{type.role.id ?? 'role not found'}</span>
```

src/components/PermissionsDialog/PermissionsDialog.react.js
```diff
@@ -982,20 +1008,21 @@ export default class PermissionsDialog extends React.Component {
...
+              <span className={styles.selectable}>
+                <a target="_blank" href={this.urlForKey(key)} >{type.user.id}</a>
+              </span>
               {pill('User')}
             </span>
           </p>
           <p className={styles.hint}>
-            {'username: '}
-            <span className={styles.selectable}>{type.user.name}</span>
+            username: <span className={styles.selectable} style={{color:type.user.name ? undefined : '#f00'}}>{type.user.name ?? 'USER NOT FOUND'}</span>
```
mtrezza:
```suggestion
            username: <span className={styles.selectable} style={{color:type.user.name ? undefined : '#f00'}}>{type.user.name ?? 'user not found'}</span>
```

mtrezza (approved):
Looks good!

src/components/PermissionsDialog/PermissionsDialog.react.js
mtrezza: Since this is already highlighted with read font color, I've removed the ALL CAPS

src/components/PermissionsDialog/PermissionsDialog.react.js
mtrezza: Since this is already highlighted with read font color, I've removed the ALL CAPS

parseplatformorg:
🎉 This change has been released in version [5.2.0-alpha.2](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-alpha.2)

parseplatformorg:
🎉 This change has been released in version [5.2.0-beta.2](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-beta.2)

parseplatformorg:
🎉 This change has been released in version [5.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0)

PR #2463, opened 2023-06-17
feat: Add typing with auto-complete to select a filter field in the data browser

### LINKED ISSUE #1891: Add auto-complete typing to column dropdowns

### Current Limitation
Currently, selecting a column from a list (dropdown) can only be done by scrolling through the list.

Allow to type the column name with auto-complete, to make it easier to select the column. For example, even just typing the letter `u` makes it much faster to find the column `user`, than scrolling down and having to visually scan all column names.

As a visual aid, the entered filter string should also be highlighted (bold) in the filtered column name results.

### Example Use Case

Entering `obj` should filter the list by column names containing (not just starting with) that string:

image

n/a

patelmilanun:
- add strict prop to autocomplete to set value from suggestion only
- add inner popover logic to detect inner popover click
- add containerStyle to style input container
- add suggestionsItemStyle to style suggestion list items

### Issue Description
We needed a way to have autocomplete for our field selection box when inside filter. This is required to get a quick search over long list of coulmns.

Closes: #1891 

### Approach
This PR adds autocomplete to the filter field selector. This PR also adds some styling props to customize the look of it. It also adds new prop called strict. This is required to always have a value that is in the suggestion list i.e, one of the column. I have made value coming from the parent a default value. If you pass this prop as false or don't pass it at all it will work as the old Autocomplete. So no breaking changes.

### TODOs before merging
NA

### DIFF
--- a/src/components/BrowserFilter/FilterRow.react.js
+++ b/src/components/BrowserFilter/FilterRow.react.js
@@ -95,6 +95,20 @@ let FilterRow = ({
       }
     }, [])
 
+    // This will add https://github.com/parse-community/parse-dashboard/pull/2463#issuecomment-1595918018
+    const buildSuggestions = (input) => {
+      const lowerCaseInput = input.toLowerCase();
+      return fields.filter((field) => {
+        let inputIndex = 0;
+        for (let i = 0; i < field.length; i++) {
+          if (field[i].toLowerCase() === lowerCaseInput[inputIndex]) {
+            inputIndex++;
+          }
+        }
+        return inputIndex === lowerCaseInput.length;
+      });
+    };
+
     return (
       <div className={styles.row}>
         <Autocomplete
@@ -129,7 +143,7 @@ let FilterRow = ({
           value={currentField}
           suggestions={fields}
           onChange={onChangeField}
-          buildSuggestions={(input) => fields.filter((s) => s.startsWith(input))}
+          buildSuggestions={buildSuggestions}
           buildLabel={() => ''}
         />
         <ChromeDropdown
--- a/src/components/BrowserFilter/FilterRow.react.js
+++ b/src/components/BrowserFilter/FilterRow.react.js
@@ -6,6 +6,7 @@
  * the root directory of this source tree.
  */
 import ChromeDropdown  from 'components/ChromeDropdown/ChromeDropdown.react';
+import Autocomplete    from 'components/Autocomplete/Autocomplete.react';
 import { Constraints } from 'lib/Filters';
 import DateTimeEntry   from 'components/DateTimeEntry/DateTimeEntry.react';
 import Icon            from 'components/Icon/Icon.react';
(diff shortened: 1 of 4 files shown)


patelmilanun:
@mtrezza can u please review this, I have added some new functionality to the core Autocomplete component. Even though I have checked in places where the original one is used and it's working fine, I'm still not that confident I don't know why. Maybe it's my first time modifying core component. Please check it once.

mtrezza:
From what I've tested it works fine, but it's strictly suggesting fields that *start with* the entered text. It would be better to:

- a) suggest any field that contains the entered text anywhere in the field name, e.g. `cat` matches `xcat` and `catx`, but not `cxat`.
- b) suggest any field text that contains the entered letters in the given order (but not necessarily following each other), e.g. `cat` matches `xcat `cxat` but not `act`.

(b) is more versatile and is more graceful with typos, so if possible I'd suggest (b); if that's too complex then I'd go with (a). Not sure how you do the matching, but (b) and (a) are easy in regex.

patelmilanun:
Hmm, I was thinking of the same. This is as-it-is component of Autocomplete and doesn't have the functionalities you mentioned. So I'll try to add it.

patelmilanun:
Sorry my bad, there is a prop that do this. I'm pushing new changes now please take a look.

src/components/BrowserFilter/FilterRow.react.js
```diff
@@ -92,15 +93,59 @@ let FilterRow = ({
       if (input !== null && editMode) {
         input.focus();
       }
-    }, [])    
+    }, [])
+
+    // This will add https://github.com/parse-community/parse-dashboard/pull/2463#issuecomment-1595918018
+    const buildSuggestions = (input) => {
```
mtrezza, line 98: I think this whole method can be simplified to:

```
const regex = new RegExp(input.split('').join('.*?'), 'i');
return fields.filter(f => f.test(regex));
```

That would be approach [(b)](https://github.com/parse-community/parse-dashboard/pull/2463#issuecomment-1595918018). Comment above method can be removed.

mtrezza (changes requested):

src/components/BrowserFilter/FilterRow.react.js
patelmilanun: I had the regex version of the code but not decided to push as it is a little bit confusing to the other contributing person. So decided to keep the logic very simple.

What should I do now? Should I change it with your given code?

mtrezza: Confusing in what way?

patelmilanun: Not confusing but a complex one.

mtrezza: The regex in https://github.com/parse-community/parse-dashboard/pull/2463#discussion_r1233232407 isn't really complex. We can use a regex where it makes sense (performance, maintainability, etc). I'd say the dashboard code should largely be maintainable with mid-level programming knowledge, regex being part of that.

mtrezza: @patelmilanun Do you want to use the suggested regex (if it works)? Then we can go ahead and merge this.

patelmilanun: Yes, I'll apply it and push it. I was just mentioning that I have generated the regex with the help of AI which was complex just like https://github.com/parse-community/parse-dashboard/pull/2464#discussion_r1233236494 this one, the one in the screenshot. And that is why decided to not go with it.

Since you suggested a better one I can do that. It's just that I'm not that good with regex and thus implemented the simplest solution.

patelmilanun: There is slight error in ur suggestion, instead of `f => f.test(regex)` we do `f => regex.test(f)`

mtrezza: I personally didn't come across an AI that generated good quality regex yet. The one I suggested should do the trick. Could you try it out?

mtrezza: Tried and works fine

mtrezza (approved):
Looks good! Nice job, this is a great feature

parseplatformorg:
🎉 This change has been released in version [5.2.0-alpha.17](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-alpha.17)

parseplatformorg:
🎉 This change has been released in version [5.3.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [5.3.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0)

PR #2464, opened 2023-06-18
fix: Pasting location coordinates into field of type `GeoPoint` does not work in data browser

### LINKED ISSUE #2462: Allow to paste coordinates tuple in data browser

### Current Limitation

While it's possible to select a geo point field in the data browser and copy its content tuple (long, lat), it's not possible to paste the tuple. This makes it difficult to copy a location from one field and paste it into another field as long and lat need to be pasted individually in two separate text fields.

https://github.com/parse-community/parse-dashboard/assets/5673677/1d451abc-63c8-4b8e-9538-6fa844ea6c3d

Allow to paste the copies geo point in a single step into a geo point field.

### Example Use Case

1. In data browser select a geo point field.
2. Copy the field value Crtl+C; adds  `(12.345, 12.345)` to the pasteboard.
3. Select another geo point field, double click and use Crtl+V to paste `(12.345, 12.345)`.

Copy and paste long and lat separate to a new geo field.

patelmilanun:
- add formats like below to be paste-able in geo point field
    - (x, y)
    - (x,y)
    - x, y
    - x,y
  
### Issue Description
There was a paste functionality for GeoPoint that was accepting the x,y format already. But as when we are copying the field we are getting the format like (x, y) which makes the existing functionality outdated.

Closes: #2462 

### Approach
I basically added other formats including (x, y) that will get accepted when pasting.

### TODOs before merging
NA

### DIFF
--- a/src/components/GeoPointEditor/GeoPointEditor.react.js
+++ b/src/components/GeoPointEditor/GeoPointEditor.react.js
@@ -106,17 +106,16 @@ export default class GeoPointEditor extends React.Component {
       let value = e.target.value;
 
       if (!validateNumeric(value)) {
-        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y
-        const regex = /\((-?\d+\.?\d*),\s*(-?\d+\.?\d*)\)|(-?\d+\.?\d*),\s*(-?\d+\.?\d*)/;
-        const match = value.match(regex);
+        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y and many more
+        const regex =
+          /[\[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;
+        const match = regex.exec(value);
 
         if (!match) {
           return null;
         }
 
-        let values = match[0].replace(/[()]/g, '').split(',');
-
-        values = values.map(val => val.trim());
+        const values = [match.groups.x, match.groups.y];
 
         if (values[0].length > 0 && validateNumeric(values[0])) {
 
--- a/src/components/GeoPointEditor/GeoPointEditor.react.js
+++ b/src/components/GeoPointEditor/GeoPointEditor.react.js
@@ -106,7 +106,6 @@ export default class GeoPointEditor extends React.Component {
       let value = e.target.value;
 
       if (!validateNumeric(value)) {
-        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y and many more
         const regex =
           /[[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;
         const match = regex.exec(value);
--- a/src/components/GeoPointEditor/GeoPointEditor.react.js
+++ b/src/components/GeoPointEditor/GeoPointEditor.react.js
@@ -106,26 +106,32 @@ export default class GeoPointEditor extends React.Component {
       let value = e.target.value;
 
       if (!validateNumeric(value)) {
-        var values = value.split(',');
+        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y
+        const regex = /\((-?\d+\.?\d*),\s*(-?\d+\.?\d*)\)|(-?\d+\.?\d*),\s*(-?\d+\.?\d*)/;
+        const match = value.match(regex);
 
-        if (values.length == 2) {
-          values = values.map(val => val.trim());
+        if (!match) {
+          return null;
+        }
 
-          if (values[0].length > 0 && validateNumeric(values[0])) {
+        let values = match[0].replace(/[\(\)]/g, '').split(',');
 
-            if (values[1].length <= 0 || !validateNumeric(values[1])) {
-              this.setState({ latitude: values[0] });
-              this.longitudeRef.current.focus();
-              this.longitudeRef.current.setSelectionRange(0, String(this.state.longitude).length);
-              return;
-            }
+        values = values.map(val => val.trim());
 
-            if (validateNumeric(values[1])) {
-              this.setState({ latitude: values[0] });
-              this.setState({ longitude: values[1] });
-              this.longitudeRef.current.focus();
-              return;
-            }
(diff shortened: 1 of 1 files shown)


mtrezza:
> There was a paste functionality for GeoPoint that was accepting the x,y format already. But as when we are copying the field we are getting the format like (x, y) which makes the existing functionality outdated.

Who would have thought. So this is actually a bug fix I suppose.

patelmilanun:
Yes, we can say that. It now accepts total of 4 format.

src/components/GeoPointEditor/GeoPointEditor.react.js
```diff
@@ -106,26 +106,32 @@ export default class GeoPointEditor extends React.Component {
...
-            }
-
-            if (validateNumeric(values[1])) {
-              this.setState({ latitude: values[0] });
-              this.setState({ longitude: values[1] });
-              this.longitudeRef.current.focus();
-              return;
-            }
+        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y
+        const regex = /\((-?\d+\.?\d*),\s*(-?\d+\.?\d*)\)|(-?\d+\.?\d*),\s*(-?\d+\.?\d*)/;
```
mtrezza: The regex seems too complex while being quite strict. This is easier to read and extend in the future, while being more tolerant:

`[\[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?`

It accepts:

```
1.0,1.0
1.0, 1.0
(1.0,1.0)
("1.0","2.0")
['1.0','1.0']
```

I'm not sure about different locales though, how should we support that? If a user's locale uses a comma as decimal marker, then the input could be `("1,0", "1,0"`). But maybe we can ignore that for now sine the dashboard is not i18n'ed anyway.

mtrezza (changes requested):

src/components/GeoPointEditor/GeoPointEditor.react.js
patelmilanun: ok i'll go with your suggestion

mtrezza: Great, just a note, the regex contains named capture groups, which should make it easier to extract the `x` and `y` values.

```
const regex = /[\[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;
const input = '[3.14, 2.718]';
const match = regex.exec(input);
const x = match.groups.x;
const y = match.groups.y;
```

patelmilanun: it does not accept ("1.0","2.0") , ['1.0','1.0'] basically square brackets and quote with parentheses.

mtrezza: let me check...

mtrezza: The regex works, here's the test:

```
const values = [
  '1.2,3.4',
  '1.2, 3.4',
  '(1.2,3.4)',
  '("1.2","3.4")',
  `['1.2','3.4']`,
];

const regex = /[\[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;

for (const value of values) {
  const match = regex.exec(value);
  const x = match.groups.x;
  const y = match.groups.y;
  const correct = parseFloat(x) === 1.2 && parseFloat(y) === 3.4;
  console.log('correct: ' + correct);
}
```

So it must be another issue. I guess the pasted value is different.

patelmilanun: sorry i was not considering the fact that it is a named capture groups. Now its working. Thanks.

src/components/GeoPointEditor/GeoPointEditor.react.js
```diff
@@ -106,26 +106,31 @@ export default class GeoPointEditor extends React.Component {
...
-              return;
-            }
-
-            if (validateNumeric(values[1])) {
-              this.setState({ latitude: values[0] });
-              this.setState({ longitude: values[1] });
-              this.longitudeRef.current.focus();
-              return;
-            }
+        // This regex will match this form of input: (x, y) or (x,y) or x, y or x,y and many more
```
mtrezza:
```suggestion
```

src/components/GeoPointEditor/GeoPointEditor.react.js
```diff
@@ -106,26 +106,30 @@ export default class GeoPointEditor extends React.Component {
...
-            }
-
-            if (validateNumeric(values[1])) {
-              this.setState({ latitude: values[0] });
-              this.setState({ longitude: values[1] });
-              this.longitudeRef.current.focus();
-              return;
-            }
+        const regex =
+          /[[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;
```
mtrezza, line 110: Are you sure this regex works without the escape character?

mtrezza:
```suggestion
        const regex = /[[("' ]?(?<x>[0-9.]+)["' ]?,["' ]?(?<y>[0-9.]+)["' )\]]?/;
```

mtrezza (approved):
Looks good, tested in preview and works

parseplatformorg:
🎉 This change has been released in version [5.2.0-alpha.18](https://github.com/parse-community/parse-dashboard/releases/tag/5.2.0-alpha.18)

parseplatformorg:
🎉 This change has been released in version [5.3.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.1](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [5.3.0](https://github.com/parse-community/parse-dashboard/releases/tag/5.3.0)

Issue #2478, opened 2023-06-24
Improve filer dialog auto-complete usability


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

https://github.com/parse-community/parse-dashboard/pull/2463 introduced auto-complete in the filter dialog. However, it's usability should be improved. When opening the filter dialog, a key is pre-selected and needs to be manually deleted for the whole key list to become visible. This can be more time consuming than before the auto-complete feature was added.

### Feature / Enhancement Description

Behavior when opening empty filter dialog (without any filter set), and when clicking on the "Add" button to add a new filter condition:
- Auto-focus on key text field 
- Do not pre-select any field, i.e. the key text field should be empty and the whole key list should be visible

mtrezza:
cc @patelmilanun

patelmilanun:
I intentionally added the default key for autocomplete. If we do not do it than we need to put some validation. The component itself doesn't have the validation so it need to be added externally. 

Do you still want to go ahead with this? I mean if u click on the field the whole text will get selected anyway. It's just 1 keystroke more than the older version.

mtrezza:
I'd expect by default seeing an empty text field with the whole list showing.

The pre-selection of the key is really arbitrary and chances are that in most cases its *not* the text field someone is looking for. Could you disable the 'apply' and 'save' buttons unless every field has a valid key selected? If that's too difficult, I would just clear the text field as soon as someone clicks into it.

I think the issue is that the dashboard is currently not very keyboard friendly. There's a lot of mouse action and that is always slower than keyword shortcuts. I would imagine at some point we get to the following scenario:

1. Hit key `F` -> filter dialog opens; text field has focus automatically.
2. Enter `em` -> fields `email` and `email2` shows in list.
3. Hit arrow down key -> field `email` is selected.
4. Hit `Enter` key -> search filter is applied (like clicking the "Apply" button)

In order to get to that, it would be good if when you open the filter dialog, the text field gets focused and is empty, with the whole field list showing.

mtrezza:
@Bhavyajain21 this may be interesting for you

PR #2482, opened 2023-06-27
fix: Keyboard accessibility is broken in filter dialog in data browser

### LINKED ISSUE #2469: The filters dropdown is currently inaccessible via arrow keys.

### Issue Description

PR: #2463 
When using an autocomplete field, it is anticipated that we will have the ability to choose from a dropdown array using the keyboard's arrow keys. 

### Steps to reproduce

1) Open filter
2) Type 'id'
3) Try to select 'objecrId' from dropdown by down arrow key.

### Actual Outcome

Not accessible by arrow keys.

### Expected Outcome

it should be accessible by arrow keys.

AshishBarvaliya:

### Issue Description

Closes: #2469 

### Approach

The keyboard accessibility code was present, but unfortunately, it wasn't functioning due to a bug.

### DIFF
--- a/src/components/Autocomplete/Autocomplete.react.js
+++ b/src/components/Autocomplete/Autocomplete.react.js
@@ -246,26 +246,16 @@ export default class Autocomplete extends Component {
   onKeyDown(e) {
     const { activeSuggestion, filteredSuggestions } = this.state;
 
-    // Enter
-    const { userInput } = this.state;
-
-      if (e.keyCode === 13) {
-            if (userInput && userInput.length > 0) {
-        this.props.onSubmit(userInput);
-      }
-    } else if (e.keyCode === 9) {
-      // Tab
-      // do not type it
-      e.preventDefault();
-
-      e.stopPropagation();
-      // move focus to input
-      this.inputRef.current.focus();
+    if (e.keyCode === 13 || e.key === 'Enter') {
       this.setState({
         active: true,
         activeSuggestion: 0,
         showSuggestions: false,
         userInput: filteredSuggestions[activeSuggestion]
+      });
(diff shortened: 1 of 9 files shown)


AshishBarvaliya:
I have implemented the basic selection functionality, such as moving up, down, and using the `enter` key for selection. However, I will make the necessary updates as per your suggested flow.

> 1. Hit key `F` -> filter dialog opens.
> 2. Enter `em` -> fields `email` and `email2` shows in list.
> 3. Hit arrow down key -> field `email` is selected.
> 4. Hit `Enter` key -> search filter is applied (like clicking the "Apply" button)

source: #2478

mtrezza:
> The keyboard accessibility code was present, but unfortunately, it wasn't functioning due to a bug.

Hah, who would have thought! Thanks for looking into this.

mtrezza:
Could you also add that when hitting the `Tab` key the focus jumps from the filter text input field to the condition field? See step 4 below:

1. Hit key F -> filter dialog opens.
2. Enter em -> fields email and email2 shows in list.
3. Hit arrow down key -> field email is selected.
4. Hit Tab key -> condition field is in focus and drop down opens ("key exists", "key does not exist", etc.)

For me it currently sets the focus to the browser address bar when hitting Tab.

And could you change the selection style from the current dark blue text color to only changing the background color of the selected field? I would use one of the blue colors that we already have.

image

mtrezza:
@AshishBarvaliya Could you resolve the conflicts so we can get this ready for merge?

mtrezza:
@AshishBarvaliya Do you think we could get this merged? It seems to be ready, just the conflicts.

dblythy:
Closed by #2771

PR #2528, opened 2024-02-20
fix: App metrics for user and installation counts show dash

### LINKED ISSUE #2502: "At a glance" total user metric always displays "-"
### New Issue Checklist

### Issue Description

The metric at the "apps" dashboard always shows a "-" on both total users and total installations:
image

I have a class called "_User" per the default with one user in right now on the testing DB. I have tried to search for anyone else having this issue but could not find any. Also tried going through the source code but did not find a solution. Have I missed a config setting or is it a bug?

### Steps to reproduce

This is my config. Database has the default "_User" class. On the readme the dashboard prints total users: 0
```js
const config = {
  apps: [
    {
      serverURL: env.publicServerURL,
      appId: env.appId,
      masterKey: env.masterKey,
      readOnlyMasterKey: env.readOnlyMasterKey,
      appName: env.appName,
      supportedPushLocales: ["en"],
      production: env.DB == "prod",
      iconName: "SuitClubs.svg",
    },
  ],
  users: [
    {
      user: "master",
      pass: env.dashWritePass,
    },
    {
      user: "moderator",
      readonly: true,
      pass: env.dashReadOnlyPass,
    },
  ],
  iconsFolder: "icons",
};

```

### Actual Outcome

`total users: -`

### Expected Outcome

`total users: 1`
### Environment

Dashboard
- Parse Dashboard version: `5.2.0`
- Browser (Safari, Chrome, Firefox, Edge, etc.): `Arc and Chrome`
- Browser version: `latest, but also last 1-2 years while i've used parse on different configurations`

Server
- Parse Server version: `6.2.2`
- Operating system: `macos Ventura 13.5.1`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local and fly.io and aws app runner and aws ec2`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `MongoDB Atlas prob. latest`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas`

### Logs
Did not find any relevant logs.

AshishBarvaliya:

### Issue Description

Closes: #2502 

### Approach

The positioning of the update stats code was incorrect, so I restructured the code to ensure that the stats are updated before the function returns.

### DIFF
--- a/src/dashboard/Apps/AppsIndex.react.js
+++ b/src/dashboard/Apps/AppsIndex.react.js
@@ -115,10 +115,10 @@ class AppsIndex extends React.Component {
   }
 
   componentWillMount() {
+    document.body.addEventListener('keydown', this.focusField);
     AppsManager.getAllAppsIndexStats().then(() => {
       this.forceUpdate();
     });
-    document.body.addEventListener('keydown', this.focusField);
   }
 
   componentWillUnmount() {
--- a/src/dashboard/Apps/AppsIndex.react.js
+++ b/src/dashboard/Apps/AppsIndex.react.js
@@ -115,15 +115,15 @@ class AppsIndex extends React.Component {
   }
 
   componentWillMount() {
+    AppsManager.getAllAppsIndexStats().then(() => {
+      this.forceUpdate();
+    });
+    document.body.addEventListener('keydown', this.focusField);
     if (AppsManager.apps().length === 1) {
       const [app] = AppsManager.apps();
       this.props.navigate(`/apps/${app.slug}/browser`);
       return;
     }
-    document.body.addEventListener('keydown', this.focusField);
-    AppsManager.getAllAppsIndexStats().then(() => {
-      this.forceUpdate();
-    });
   }
 
   componentWillUnmount() {
--- a/src/dashboard/Apps/AppsIndex.react.js
+++ b/src/dashboard/Apps/AppsIndex.react.js
@@ -119,11 +119,6 @@ class AppsIndex extends React.Component {
       this.forceUpdate();
     });
     document.body.addEventListener('keydown', this.focusField);
-    if (AppsManager.apps().length === 1) {
-      const [app] = AppsManager.apps();
-      this.props.navigate(`/apps/${app.slug}/browser`);
-      return;
-    }
   }
 
   componentWillUnmount() {


patelmilanun:
It will still show wrong number because of https://github.com/parse-community/parse-server/issues/8502, but the fix is for this issue so this issue is fixed with this PR.

mtrezza:
So what should we do with this PR?

AshishBarvaliya:
Since it is fixing the UI problem, we should merge it as of now.

patelmilanun (approved):
I think as the issue is about metric showing "-" which is fixed with this PR so we should be good to go for this PR. 

As the correct metric will show for the mongodb and only cause problem in postgres.

mtrezza (dismissed):

mtrezza:
What's the issue that this PR fixes? In the current dashboard version it seems to show dashes only for 1 or 2 seconds and then shows the numbers. So the dashes just seem to be a placeholder until the app metrics are loaded?

AshishBarvaliya:
I've checked that I installed the NPM package and set up an app with Express.js. But I'm seeing a '-' here that shouldn't be.

Screenshot 2024-02-23 at 1 39 58 PM

mtrezza:
Looking at the code, if there is only 1 app, the dashboard should navigate directly to the data browser. If there are multiple apps, then the code change shouldn't have any effect. In your case, if you have only 1 app, then maybe the bug is that it doesn't navigate to the data browser? How do you get to that app overview page, do you see it after login or when you are in the data browser and click on the "Your Apps" menu?

AshishBarvaliya:
I have tested with this [uffizzi link](https://app.uffizzi.com/github.com/parse-community/parse-dashboard/pull/2528). After logged in this is the main page I am getting. Is this behaviour expected? 

Screenshot 2024-02-24 at 8 54 28 AM

mtrezza:
I don't think so. From the code I'd say that if there is only 1 app it should go directly to the data browser. It doesn't go there, so there may be an issue with the redirect code. That explains why the stats don't update.

The fix in this PR doesn't seem to make much sense, because it updates the stats and then immediately auto-redirects, if there is only 1 app.

I think the auto-redirect is actually a strange behavior. Why not always show the apps overview page for consistency and entirely remove the auto-redirect code. What do you think?

AshishBarvaliya:
Since redirections never seem to work for me, I've gotten used to relying on the menu. I agree with you about removing the redirection code to ensure consistency.

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [5.4.0-alpha.6](https://github.com/parse-community/parse-dashboard/releases/tag/5.4.0-alpha.6)

parseplatformorg:
🎉 This change has been released in version [6.0.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/6.0.0-beta.1)

parseplatformorg:

PR #2529, opened 2024-02-23
feat: Add descriptive statistics for number cells in data browser

### LINKED ISSUE #2506: Add descriptive statistics to data browser

### Current Limitation

Currently it's not possible to see descriptive statistics in the data browser. Instead, to calculate the sum of two fields they have to calculated manually, or the data has to be exported to processed.

Add a feature that:
- Allows to select cells in the data browser across rows and columns.
- Calculates and displays common descriptive statistics about the selected cells, like "Sum", "Mean", "p99".

### Example Use Case

This should work similar to a spreadsheet program where cells can be selected and these statistics are displayed. 

image

AshishBarvaliya:

### Issue Description

Closes: #2506 

### Approach

1. Holding down the Shift key while clicking will conveniently highlight the entire area.
2. Currently, I'm displaying statistics in the toolbar using a default theme. Feel free to share your preferences.
3. By clicking on the stats button, users can opt for different types of statistics.
4. If any selected column isn't categorized as "Number," it won't be displayed or included in calculations.
5. Null or undefined values will be disregarded, ensuring they don't impact the count or any statistics.

ss1:
Screenshot 2024-02-23 at 1 33 09 PM

ss2:
Screenshot 2024-02-23 at 1 33 17 PM

### DIFF
--- a/src/components/Toolbar/Toolbar.scss
+++ b/src/components/Toolbar/Toolbar.scss
@@ -85,3 +85,41 @@ body:global(.expanded) {
   right: 14px;
   top: 4px;
 }
+
+.stats {
+  position: absolute;
+  right: 20px;
+  bottom: 10px;
+  background: $blue;
+  border-radius: 3px;
+  padding: 2px 4px;
+  font-size: 16px;
+  color: white;
+  box-shadow: none;
+  border: none;
+}
+
+.stats_popover_container {
+  display: flex;
+  flex-direction: column;
+  background: $blue;
+  gap: 4px;
+  padding: 2px 0px;
+  font-size: 16px;
+  color: white;
+}
+
+.stats_popover_item {
+  cursor: pointer;
+  padding: 0px 8px;
+
+  &:hover {
+    color: $blue;
+    background-color: white;
+  }
+}
+
+.active {
+  color: $blue;
+  background-color: white;
+}
--- a/src/components/Toolbar/Toolbar.scss
+++ b/src/components/Toolbar/Toolbar.scss
@@ -92,7 +92,7 @@ body:global(.expanded) {
   bottom: 10px;
   background: $blue;
   border-radius: 3px;
-  padding: 2px 4px;
+  padding: 2px 6px;
   font-size: 14px;
   color: white;
   box-shadow: none;
@@ -103,7 +103,9 @@ body:global(.expanded) {
   display: flex;
   flex-direction: column;
   background: $blue;
-  gap: 4px;
+  border-radius: 3px;
+  margin-top: 1px;
+  gap: 2px;
   padding: 2px 0px;
   font-size: 14px;
   color: white;
@@ -111,7 +113,7 @@ body:global(.expanded) {
 
 .stats_popover_item {
   cursor: pointer;
-  padding: 0px 8px;
+  padding: 0px 6px;
 
   &:hover {
     color: $blue;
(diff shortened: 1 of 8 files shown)


mtrezza:
Wow! This is a truly impressive PR!

Could the selected cells have a highlighted cell background color, so that it's easier to identify the selected cells visually? Similar to how you'd see it in a spreadsheet UI.

AshishBarvaliya:
I've added a lighter shade of blue to ensure that the text and background have sufficient contrast.

Screenshot 2024-02-25 at 6 52 45 PM


Let me know if any alternative styles are needed for the stats menu or if you'd like to add any other types of stats.

mtrezza (commented):
> If any selected column isn't categorized as "Number," it won't be displayed or included in calculations.

I found that if there is a string column within the selection, the stats field won't be displayed at all. Is that intended, or should it just be ignored in the calculation? For example, this selection doesn't display anything:

image

I think that actually makes sense, because maybe we cannot always correctly convert a string to a number value. So maybe we should not try to convert it at all and not display anything. Because in the example above there is a number from a number column and a number from a string column. So it could be unclear whether the string number is included or not, if we still displayed the stats field. There are also characters that look like numbers but are not actually different unicodes, so converting some of them and others not may be too confusing. What do you think?

Could you please take another look at the failing jobs?

AshishBarvaliya:
> I found that if there is a string column within the selection, the stats field won't be displayed at all. Is that intended, or should it just be ignored in the calculation? For example, this selection doesn't display anything:

Yes, this intentional behaviour is designed to prevent confusion. Currently, only 'Number' types are allowed. We can consider adding exceptions for strings if necessary or requested, but for now, it's not required.

I've also addressed the issue with the jobs that were failing.

mtrezza (dismissed):
Looks good! Waiting for tryout in preview env.

src/components/Toolbar/Toolbar.scss
```diff
@@ -85,3 +85,41 @@ body:global(.expanded) {
...
+  border: none;
+}
+
+.stats_popover_container {
+  display: flex;
+  flex-direction: column;
+  background: $blue;
+  gap: 4px;
+  padding: 2px 0px;
+  font-size: 16px;
```
mtrezza:
```suggestion
  font-size: 14px;
  color: white;
  box-shadow: none;
  border: none;
}

.stats_popover_container {
  display: flex;
  flex-direction: column;
  background: $blue;
  gap: 4px;
  padding: 2px 0px;
  font-size: 14px;
```

mtrezza (dismissed):

src/components/Toolbar/Toolbar.scss
```diff
@@ -85,3 +85,41 @@ body:global(.expanded) {
...
 }
+
+.stats {
+  position: absolute;
+  right: 20px;
+  bottom: 10px;
+  background: $blue;
+  border-radius: 3px;
+  padding: 2px 4px;
+  font-size: 14px;
```
mtrezza, line 96: I've changed the font size from 16px to 14px to align with the other font sizes in the toolbar. Not sure but I may have messed up the layout of the drop down, could you please take a look? Then this is ready to merge - great job!

AshishBarvaliya: Updated the styles. Now it would look like this.

 
Screenshot 2024-02-26 at 2 08 02 PM

parseplatformorg:
🎉 This change has been released in version [5.4.0-alpha.7](https://github.com/parse-community/parse-dashboard/releases/tag/5.4.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [6.0.0-beta.1](https://github.com/parse-community/parse-dashboard/releases/tag/6.0.0-beta.1)

parseplatformorg:

Issue #2571, opened 2024-06-02
Add page size and current page index to data browser


badboy-tian:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ ] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation

can not set pagesize, so sometimes load slow
### Feature / Enhancement Description

add pagesize and pageindex in the bottom of the page
### Alternatives / Workarounds

no
### 3rd Party References

<img width="1139" alt="image" src="https://github.com/parse-community/parse-dashboard/assets/2835129/80231b96-3d62-45f4-9d30-1a2cfe4a5412">

<img width="1312" alt="image" src="https://github.com/parse-community/parse-dashboard/assets/2835129/b6efe99b-18cc-445a-be45-e2416f9866a7">

mtrezza:
Good idea! What is the 3rd party ref? Looks like a simple clean design.

badboy-tian:
@mtrezza This is a Chinese version of parse-server, maybe. But it does not support self-built servers, and the minimum consumption is 30 yuan per day.  https://www.leancloud.cn/
![image](https://github.com/parse-community/parse-dashboard/assets/2835129/b39e8c36-db37-4726-9693-7f206fd4d1d3)

mtrezza:
I see, yes, we are aware of this service.

Studycode001:
Can I add this feature plzz

alokdangre:
Can I add this feature please?
@badboy-tian

badboy-tian:
yes, you can

Studycode001 ***@***.***>于2024年7月12日 周五13:47写道：

JiteshSinghShekhawat:
is this issue still open ? @badboy-tian @mtrezza

badboy-tian:
@JiteshSinghShekhawat  yes...

JiteshSinghShekhawat:
Thanks for confirming! Could you please assign this issue to me so I can start working on it?

badboy-tian:
@JiteshSinghShekhawat i do not know how to assign this issue to you.

badboy-tian:
@mtrezza perhaps need help. assign this issue to @JiteshSinghShekhawat

[...]

PR #2624, opened 2024-10-30
fix: Pagination in data browser not working

### LINKED ISSUE #1551: Data browser shows only first 200 records

muhleder:

### Issue Description

Related:  #1551 

### Approach
First issue is that the scrolling element in BrowserTable must have changed, so the scroll listener was never being fired.

Main issue really though was that the fetchNextPage code is much more complex than it needed to be. I've just used the existing state.lastMax variable to record how many records have been retrieved, and use the query.skip() function to do the pagination (as recommended in the query api documentation).

I've also added a state.fetchingNextPage variable to prevent the same fetchNextPage request being fired in quick succession. I noticed this happening during testing and it actually managed to crash my server on the bigger tables.

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/src/dashboard/Data/Browser/Browser.react.js
+++ b/src/dashboard/Data/Browser/Browser.react.js
@@ -84,6 +84,7 @@ class Browser extends DashboardView {
       lastMax: -1,
       newObject: null,
       editCloneRows: null,
+      fetchingNextPage: false,
 
       lastError: null,
       lastNote: null,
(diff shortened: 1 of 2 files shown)


mtrezza:
The history of this issue is almost as infinite as the scrolling itself. We believe that the only efficient and bug-free solution would be using https://github.com/parse-community/parse-server/issues/7637. Everything else has failed so far. Surely pagination is easy in an environment where data does not change, but as soon as a new record is added between page loads, a simple `skip` won't do the trick anymore. Maybe the "complex pagination query" has been "over-engineered" to a point where a simple solution ends up working seemingly better. How many edge cases we'd be again opening up with going back to square 1 -- no idea. In any case, since pagination currently doesn't work at all, there's also not much to loose.

A few considerations:

1. The  "complex pagination query" that you've replaced is found at 2 places:

    /src/dashboard/Data/Browser/ObjectPickerDialog.react.js
    /src/dashboard/Data/Browser/Browser.react.js
    
    That begs the questions why duplicate code and should this be de-duplicated, or does the query need to modified only in 1 of the two places?
    
2. We'd need to make sure not to introduce any new index requirements for the pagination to work.

3. It'd be worth reading through https://github.com/parse-community/parse-dashboard/pull/1706 before we reset the pagination logic, just to not repeat old mistakes.

5. Which edge cases do we know of and want to handle (and how?) or explicitly ignore at this point? E.g.
   - New record added in-between page loads.
   - Multiple records with same `createdAt` (if sorted by `createdAt`).
   - Multiple records with same field name `n` (if sorted by `n`).

Note I've made a small edit in the issue description; this PR seems like a temporary workaround (better buggy pagination than none?), but a true fix would require cursor based pagination, so the PR as it currently is would not close https://github.com/parse-community/parse-dashboard/issues/1551.

mtrezza:
I think https://github.com/parse-community/parse-dashboard/issues/2571 would be an alternative to allow pagination without getting into many of the issues mentioned in https://github.com/parse-community/parse-dashboard/pull/2624#issuecomment-2446977044, since only 1 page will be shown at a time.

PR #2748, opened 2025-04-13
refactor: Optimize latest package version check

mtrezza:
No real reason for this refactor other than I tried to upgrade [package-json](https://github.com/sindresorhus/package-json) to 8+ which is now pure ESM, but that failed even with `await import`, likely because the module is using `import` internally. So I at least keep the refactor.

### DIFF
--- a/Parse-Dashboard/app.js
+++ b/Parse-Dashboard/app.js
@@ -1,27 +1,34 @@
 'use strict';
+
 const express = require('express');
 const path = require('path');
-const packageJson = require('package-json');
 const csrf = require('csurf');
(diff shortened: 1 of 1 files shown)


parseplatformorg:
🎉 This change has been released in version [7.1.0-alpha.7](https://github.com/parse-community/parse-dashboard/releases/tag/7.1.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [7.1.0](https://github.com/parse-community/parse-dashboard/releases/tag/7.1.0)

PR #3360, opened 2026-05-18
fix: Chart not displayed when formula references an undefined field

mtrezza:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-dashboard/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-dashboard/blob/alpha/LICENSE).

## Issue

In the Data Browser, when creating a chart (e.g. bar chart) and adding a calculated value with the `formula` operator, the chart did not render any data if the formula referenced a field whose value was `undefined` (or missing) on the data rows.

This was inconsistent with the `sum` operator, which already treats a missing/undefined field as `0` and still renders the chart.

## Approach

When `expr-eval-fork` evaluates a formula and encounters an identifier that is not in the variables map, it throws. The error was being caught and `evaluateFormula` returned `null` for every row, leaving the chart with no data.

`evaluateFormula` now fills in any variables referenced by the parsed formula but missing from the variables map with `0` before evaluation. This makes `formula` behave consistently with `sum`: a referenced field that is undefined on a row is treated as `0`, and the chart renders.

Note that this only affects runtime evaluation. Formula validation in the chart configuration dialog still rejects unknown identifiers, so genuine typos are caught at configuration time.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Formulas now gracefully handle missing variables by defaulting them to 0 instead of producing errors or invalid results.

* **Tests**
  * Updated formula evaluation tests to verify undefined variable handling.
  * Added chart rendering tests to ensure visualizations display correctly when referenced fields are undefined.

[Review Change Stack](https://app.coderabbit.ai/change-stack/parse-community/parse-dashboard/pull/3360?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/FormulaEvaluator.js
+++ b/src/lib/FormulaEvaluator.js
@@ -140,7 +140,19 @@ export function evaluateFormula(formula, variables) {
   try {
     const processedFormula = preprocessFormula(formula);
     const expr = parser.parse(processedFormula);
-    let result = expr.evaluate(variables);
+
+    // Default any variables referenced by the formula but missing from the
+    // provided variables to 0. This keeps charts rendering when a referenced
+    // field has no value on a row (consistent with the "sum" operator, which
+    // already treats missing fields as 0).
+    const safeVariables = { ...(variables || {}) };
+    for (const varName of expr.variables()) {
+      if (!(varName in safeVariables)) {
+        safeVariables[varName] = 0;
+      }
+    }
+
+    let result = expr.evaluate(safeVariables);
 
     // Convert boolean results to numbers (for comparison operators)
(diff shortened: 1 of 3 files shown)


parseplatformorg:
🎉 This change has been released in version [9.1.2-alpha.1](https://github.com/parse-community/parse-dashboard/releases/tag/9.1.2-alpha.1)

parseplatformorg:
🎉 This change has been released in version [9.2.0](https://github.com/parse-community/parse-dashboard/releases/tag/9.2.0)

PR #3379, opened 2026-07-04
fix: Preserve line breaks in info panel text item

### LINKED ISSUE #2651: Line breaks are lost when displaying text in info panel
### New Issue Checklist

### Issue Description
When showing a multi line text in info panel, the line breaks are lost, and text is rendered as a single line.

Image

### Steps to reproduce
Setup [info panel](https://github.com/parse-community/parse-dashboard?tab=readme-ov-file#info-panel) on a collection, and return following from the cloud function:
```js
{
  panel: {
    segments: [
      {
        title: "User Info",
        items: [
          {
            type: "text",
            text: `
line 1

line 2`,
          },
        ],
      },
    ],
  },
}
```

### Actual Outcome
Besides the line breaks, text is wrapped into single line.

### Expected Outcome
I would expect line breaks to be preserved.

### Environment

Dashboard
- Parse Dashboard version: `6.0.0-alpha.28`
- Browser (Safari, Chrome, Firefox, Edge, etc.): `Chrome`
- Browser version: `131.0.6778.205 (Official Build) (arm64)`

Server
- Parse Server version: `any`
- Operating system: `any`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `any`
- Database version: `any`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `any`

dblythy:
Closes #2651

Info panel `text` items lost line breaks: newlines (`\n`) in the value collapsed into spaces because the text was rendered in a plain `<p>` (default `white-space: normal`). Setting `white-space: pre-line` preserves author-entered line breaks while still collapsing redundant spaces. The text remains an escaped React child (no `dangerouslySetInnerHTML`), so there is no XSS risk.

Before

before

After

after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text display so line breaks in paragraph content are preserved as expected.


mtrezza (changes requested):
No images, CR hit limit, branch not up-to-date. req again when all done

dblythy:
@mtrezza images restored

mtrezza:
What's the source of text that now shows line breaks? What's causing these breaks?

dblythy:
This here is the response of a cloud function, if you returned `hello\nworld` for example

mtrezza:
Condensing text in a side panel is a feature I'd argue, not a bug. There is limited space in a side panel. It currently condenses line breaks, tabs, spaces. Respecting just line breaks seems arbitrary to me. And what if someone requests to respect tabs and spaces tomorrow? I believe by default it should collapse (status quo) because information density is key. Formatting should be deterministic in code - it's HTML after all. I'd rather lean towards allowing HTML tags, which gives a dev more freedom, including `<ul>`, etc. See https://github.com/parse-community/parse-dashboard/issues/2651#issuecomment-2629573289.

Btw, doesn't the text element already support a `style` property? A segment does, so a dev could probably set `white-space: pre-line` already? Does this issue even require a PR?

dblythy:
Respecting newlines isn't really arbitrary - `\n` is a hard line break in Unicode (UAX #14), not a CSS thing, whereas a space isn't. iOS, Android and web all treat `\n` as a break already, so honouring it just follows the text standard. `pre-line` is only how the web opts back in after CSS collapses everything.

That said, you're right that a dev can already pass `style` through as a workaround today:

```js
{ type: 'text', text: '...', style: { whiteSpace: 'pre-line' } }
```

i think it'd stump most dashboard devs to have to reach for CSS just to keep a line break they explicitly typed.

On raw HTML, IMV i'd steer clear. The text often interpolates user data (`Bio: ${user.get('bio')}`), so it's an XSS vector straight into the dashboard, which holds the masterKey. If we want formatting, sanitised markdown feels safer but a bigger scope.

Don't feel strongly though - happy for you to close this and the issue if you're good with the workaround.

mtrezza:
> Respecting newlines isn't really arbitrary - \n is a hard line break in Unicode

I don't understand the argument - tabs are also a unicode char. It is arbitrary in the sense that the *default* is to condense text. Why do you want to single out line breaks but keep condensing tabs and spaces? 

> i think it'd stump most dashboard devs to have to reach for CSS just to keep a line break they explicitly typed.

I would be very surprised - that should never be an expectation in HTML. Not only is it not the default behavior, condensing text is also the historic normal, long before CSS. `white-space` is a fairly young property. The expectation of a dev should be condensing, unless explicitly setting the CSS property.

PR #2991, opened 2025-10-04
feat: Add `matches regex` filter to data browser replacing limited `string contains string` filter

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Issue Description

The existing "string contains string" filter is really a regex filter with fixed regex modifier `i` (case insensitive). The filter name is difficult to understand and the fixed modifier prohibits the versatility of regex patterns.

### Approach

- Removed the existing "string contains string" filter.
- Added the "matches regex" filter with a options dialog to set modifiers.
- Fixed an issue where an entered number value would be incorrectly interpreted as a date value.

Note that the `u` (unicode) modifier is only supported from Parse Server [8.3.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/8.3.0-alpha.2).

Issue #2052, opened 2022-02-25
Consistent filter descriptions


cyb3rko:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).

### Current Limitation
There's no limitation, it's just regarding consistency.

![image](https://user-images.githubusercontent.com/46033730/155753106-ecbe521d-9d57-4385-9daa-ec66d19811a0.png)

### Feature / Enhancement Description
I feel like the filter "string contains string" should be renamed to "contains" to fit the other filters like "equals" or "starts with" because these do not have "string" in their descriptions either.

### Example Use Case
X

### Alternatives / Workarounds
X

### 3rd Party References
X

davimacedo:
Thanks for opening the issue. I think it makes sense. Would be willed to open a PR?

cyb3rko:
Same answer as mentioned in #2051 :)

mtrezza:
Since this is only a renaming of a text, you could just replace it and submit the PR. It has probably less to do with web development than with simply opening a PR 🙂

cyb3rko:
Yeah, this one is more simple than the other one.
I just didn't have the time to do it yet, sorry.

cyb3rko:
I just searched for it and haven't found it.
Where is it located?

davimacedo:
I believe in the two first file results of this search: https://github.com/parse-community/parse-dashboard/search?q=string+contains+string

cyb3rko:
Take a look, is that it?
I think changing the description in this file should be enough, right?

parseplatformorg:
🎉 This change has been released in version [4.0.0-alpha.20](https://github.com/parse-community/parse-dashboard/releases/tag/4.0.0-alpha.20)

cyb3rko:
@mtrezza 
Unfortunately we have to reopen this issue then :neutral_face: according to #2131.

mtrezza:
Not necessary, see https://github.com/parse-community/parse-dashboard/issues/2131#issuecomment-1141480793

mtrezza:
Right, re-opened , let's give it another try  🙂

PR #3392, opened 2026-07-05
feat: Add contains-substring filter for String fields

### LINKED ISSUE #2153: Undocumented regexp matching is confusing
### New Issue Checklist

### Issue Description
"string contains string" actually uses Parse.Query's `.matches(str,'i')` specifier, which is not documented and can cause problems because various punctuation has special meaning in regexps. However, as regexps are also very useful, it would be best to keep this feature while renaming it "matches" or "matches regexp" or "matches pattern" or "regexp" or "pattern", while also have a separate choice called "contains" or "contains substring", which would match a regexp produced by : 

    function escapeRegex(s:string):string {
        return s.replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&')
    }

### Steps to reproduce
Filter a string field with a "string contains string" clause (in any version of Dashboard in which it still works, e.g. 4.0.0) ; search using a `.` in place of any specific character, which will find your items anyway ; then try searching for `(` in a string column that has that character : searching for `\(` finds those items even though no `\` is in that column, and searching for literally `(` finds 0 results.

### Actual Outcome
"string contains string" searches for regexp.

### Expected Outcome
"string contains string" would search for a substring, another option would search for provided regexp.

### Environment

Dashboard
- Parse Dashboard version: 4.0.0
- Browser (Safari, Chrome, Firefox, Edge, etc.): Firefox 100.0.1

Server
- Parse Server version: 5.2.0
- Operating system: Heroku-20
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): Heroku

Database
- System (MongoDB or Postgres): MongoDB
- Database version: Unknown
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): ObjectRocket at Heroku

dblythy:
Closes #2153

### Problem

For a String column, the only substring-style filter is **matches regex**, which passes the input to `Parse.Query#matches` verbatim as a regular expression. Regex metacharacters therefore have special meaning: `.` matches any character, and metacharacters like `(` are invalid on their own, so a plain literal search either returns the wrong rows or fails with a server error. There is no way to do a simple literal substring search.

### Fix

Add a separate **contains substring** operator for String fields, backed by the SDK's `Parse.Query#contains`, which wraps the input in `\Q...\E` so the whole value is matched literally (the same quoting `startsWith` / `endsWith` already use). The existing **matches regex** operator is unchanged, so raw regex remains available.

The "rename the confusing operator" half of the issue was already handled in #2991 (the old mislabelled entry was replaced by the explicit **matches regex**), so this PR only adds the new safe operator. This does not touch the Array-field labels discussed in #2052.

Unit test added in `src/lib/tests/queryFromFilters.test.js` covering literal substring matching, metacharacter escaping, and a contrast asserting `matches` stays a raw regex.

### The new operator

The String field condition list now offers **contains substring** below **matches regex**:

new option

### Literal `.` (data: `a(b`, `c.d`, `hello`, `world`)

**matches regex** treats `.` as a wildcard and returns every row:

matches regex dot

**contains substring** matches `.` literally and returns only `c.d`:

contains dot

### Literal `a(b`

**matches regex** with `a(b` is an invalid regular expression, so the query fails (500) and nothing is shown:

matches regex paren

**contains substring** with `a(b` matches the parenthesis literally and returns the correct row:

contains paren

## Summary by CodeRabbit

* **New Features**
  * Added support for a new string filter option, allowing substring matching in queries.
  * Expanded the available string constraint list to include the new option.

* **Bug Fixes**
  * Query filtering now correctly applies substring matching when this new constraint is used.
  * Added test coverage to ensure substring and regex-based filters behave as expected, including literal handling of special characters.

### DIFF
--- a/src/lib/Filters.js
+++ b/src/lib/Filters.js
@@ -64,6 +64,12 @@ export const Constraints = {
     composable: true,
     comparable: true,
   },
+  contains: {
+    name: 'contains substring',
+    field: 'String',
+    composable: true,
+    comparable: true,
+  },
   before: {
     name: 'is before',
     field: 'Date',
@@ -185,7 +191,18 @@ export const FieldConstraints = {
   Pointer: ['exists', 'dne', 'eq', 'neq', 'starts', 'containedIn', 'unique'],
   Boolean: ['exists', 'dne', 'eq', 'neq', 'containedIn', 'unique'],
   Number: ['exists', 'dne', 'eq', 'neq', 'lt', 'lte', 'gt', 'gte', 'containedIn', 'unique'],
-  String: ['exists', 'dne', 'eq', 'neq', 'starts', 'ends', 'matches', 'containedIn', 'unique'],
+  String: [
+    'exists',
+    'dne',
+    'eq',
+    'neq',
+    'starts',
+    'ends',
+    'matches',
+    'contains',
+    'containedIn',
+    'unique',
+  ],
   Date: [
     'exists',
     'dne',
--- a/src/lib/queryFromFilters.js
+++ b/src/lib/queryFromFilters.js
@@ -232,6 +232,9 @@ export function addConstraintFromValues(query, field, constraint, compareTo, mod
     case 'matches':
       query.matches(field, String(compareTo), modifiers);
       break;
+    case 'contains':
+      query.contains(field, String(compareTo));
+      break;
     case 'keyExists':
       query.exists(field + '.' + compareTo);
       break;
--- a/src/lib/tests/queryFromFilters.test.js
+++ b/src/lib/tests/queryFromFilters.test.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2016-present, Parse, LLC
+ * All rights reserved.
+ *
+ * This source code is licensed under the license found in the LICENSE file in
+ * the root directory of this source tree.
+ */
+jest.dontMock('../queryFromFilters');
+
+const { addConstraintFromValues } = require('../queryFromFilters');
+const Parse = require('parse');
+
+function whereFor(field, constraint, compareTo, modifiers) {
+  const query = new Parse.Query('Test');
+  addConstraintFromValues(query, field, constraint, compareTo, modifiers);
+  return query.toJSON().where;
+}
+
+describe('addConstraintFromValues contains constraint', () => {
+  it('matches a plain substring literally', () => {
+    expect(whereFor('name', 'contains', 'foo')).toEqual({
+      name: { $regex: '\\Qfoo\\E' },
+    });
+  });
+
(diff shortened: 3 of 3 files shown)


mtrezza:
I don't understand what this PR is trying to achieve? What's the purpose of adding a "contains" filter, which was removed in https://github.com/parse-community/parse-dashboard/pull/2991 because it was just a subset of "matches regex"?

dblythy:
Oh, I didn't realise this previously existed. I thought the ticket was about adding the `.contains` method that Parse Server support (even in the PR you linked it was a regexp)

FWIW `startsWith`, `endsWith` can also just be done with a regexp.

Happy to close though if you prefer

mtrezza:
@coderabbitai how are startsWith, endsWith implemented? Do they have the same issue as #2991 with a fixed "i" regex option?

mtrezza:
@coderabbitai The issue in #2991 was also that no options can be customized, so emojis (/u option) for example are not recognized.

mtrezza:
@coderabbitai We need a unified solution; the same issue exists with startsWith, endsWith, no matching flexibility, which is exactly why https://github.com/parse-community/parse-dashboard/pull/2991 was merged. These kind of filter conditions with hard-coded, arbitrary, unintuitive matching limitations are impractical. How can we allow the user to quickly access a startsWith/endsWith/contains functionality and optionally use unescaped strings, given the already implemented features?

mtrezza:
@coderabbitai Should the strings entered in starts, ends, contains be automatically regex-escaped? Is that the current behavior?

mtrezza:
@coderabbitai What if we add an "escape" option to the "matches regex" options list. Then we change the startsWith/endsWith/contains to be a shortcut to a version of the regex filter, with "escape" checked and disabled, so the user can't change it. So for all 3 filters we'd use the regex condition since it supports the modifiers and they are all disguised regex conditions anyway. This also solves the issue that startsWith/endsWith/contains don't currently allow modifiers.

mtrezza:
@coderabbitai A dashboard UI checkbox isn't about "what the server receives", it's about "what the user gets". For the dashboard user, it's irrelevant whether an option is applied dashboard side or server side. The UI should just not confuse developers, for example, by pretending that the escaping option was a regex flag. We can visually separate the regex options from other options in the filter menu to make this distinction. Escaping can be useful in the "matches regex" dialogue as well. It's simply a supplemental feature. For consistency, it would make sense to use the same filter form for the "matches regex" as well as the "starts with", etc. options, since regex options apply to the "starts with" filter, and escaping also applies to the "matches regex" filter.

mtrezza:
@coderabbitai I agree with your suggestions.

